PR #97, opened 2013-08-22
Adding PlayORM as a 3rd party framework

easility:
As PlayORM has started supported MongoDB, I am adding it into the list of 3rd party libraries.


mpaik:

PR #107, opened 2013-10-28
xd is reserved for block devices on  Windows instance where sd is for Li...

scotttam:
Revising AWS documentation to have the proper block device names for Linux.


mpaik:

PR #173, opened 2014-05-22
Update serialize-documents-with-the-csharp-driver.txt

ianhorton:
Updating the method name so that it matches the one used in the rest of the example.


mpaik:
Merged

PR #178, opened 2014-06-09
Added information on mongo shell's whereabouts.

rickeywang:
I had trouble finding mongo binary, only to discover that it was never installed.

Fixed capitalization on the way too.


mpaik:

PR #192, opened 2014-08-12
Correct document field

ahmehri:
The result of the `$group` operation is the document with `_id` and `average` keys. Therefore, the `$sort` operation should be realized on the `average` and not the `amount` field.


mpaik:
Merged

PR #203, opened 2014-09-13
working disk mount script

raslin:
from here:
https://developers.google.com/compute/docs/disks#use_multi_instances

in the formatting disks section,
the old script not working.

thanks.


mpaik:
Outdated; GCE platform redacted.

PR #224, opened 2014-11-12
update with redhat installer

guiditoito:
not everyone uses debian :)


mpaik:
Merged with edits - thanks!

PR #229, opened 2014-11-20
Update google-compute-engine.txt

ashitaka1:
"Individual persistent disks" is the opposite of what this setup does. Corrected to "One persistent disk"


mpaik:

PR #238, opened 2015-01-14
Added file for Getting Started with the 3.0 Java Driver

jyemin:
This is a copy of the getting started with the Java driver page, updated to use the new CRUD API.  I looked it over, but it may violate our style guide in myriad ways.  Feel free to fix it up as you see fit.


tychoish:
merged

PR #245, opened 2015-02-23
Add missing Java class for example below

gianpaj:
Note. There are some extra classes that are needed for examples underneath, but not all the classes are added. For example `BasicDBObject`


kay-kim:
merged!

PR #262, opened 2015-03-17
DOCS-3318

mpaik:


source/includes/fact-authenticate-with-java-gssapi.rst
```diff
@@ -0,0 +1,48 @@
...
+underlying GSSAPI Java libraries can acquire a Kerberos ticket:
+
+.. code-block:: properties
+
+   java.security.krb5.realm=MYREALM.ME
+   java.security.krb5.kdc=mykdc.myrealm.me
+
+.. note::
+
+   Many authentication problems with GSSAPI/Kerberos occur due to
```
jyemin: So is this the only bit that has really changed?

mpaik: Yes, aside from some minor copyediting. The note block is the salient bit
for the ticket.

On Tue, Mar 17, 2015 at 10:39 AM, Jeff Yemin <notifications@github.com>
wrote:

> In source/includes/fact-authenticate-with-java-gssapi.rst
> <https://github.com/mongodb/docs-ecosystem/pull/262#discussion_r26577912>:
>
> > +With Kerberos you specify neither the password nor the database name,
> > +as authentication is performed via the Kerberos key distribution
> > +center (KDC).
> > +
> > +To actually run a program that authenticates with Kerberos, you will
> > +typically need to specify several system properties so that the
> > +underlying GSSAPI Java libraries can acquire a Kerberos ticket:
> > +
> > +.. code-block:: properties
> > +
> > +   java.security.krb5.realm=MYREALM.ME
> > +   java.security.krb5.kdc=mykdc.myrealm.me
> > +
> > +.. note::
> > +
> > +   Many authentication problems with GSSAPI/Kerberos occur due to
>
> So is this the only bit that has really changed?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/mongodb/docs-ecosystem/pull/262/files#r26577912>.
>



-- 
{
  name     : *"*Michael Paik*", *
  title    : "Director of Documentation",
  company  : "MongoDB, Inc.",
  web      : ["www.mongodb.org", "www.mongodb.com"]
}

source/includes/fact-authenticate-with-java-gssapi.rst
```diff
@@ -0,0 +1,48 @@
...
+   mismatches between server realm names. In general, you should
+   specify exactly the same hostname as used as the realm name in your
+   Kerberos configuration when creating GSSAPI credentials. If you
+   need to use an IP address instead, ensure that you force
```
craiggwilson: It's not just an IP address. It's any hostname that resolves but is NOT the fqdn of the hostname. So, using the shortname "myserver" instead of "myserver.mydomain" is also a candidate for using canonicalization. However (and this is why it's not the default) the kerberos RFC's state that canonicalization can cause problems (and has security repercussions) -> see http://www.ietf.org/rfc/rfc4120.txt section 1.3.

mpaik: Here we run into the common issue of explaining complications with software that we don't build. I'll revise to try to straddle the line.

mpaik:
Merged

PR #270, opened 2015-03-25
Remove eval from syntax table

bjori:
It isn't one of the most important functions, nor is it recommend to use at all


mpaik:
Merged

PR #275, opened 2015-04-06
DOCS-3365: Remove an old unmaintained admin UI link, and a for-pay one

i80and:
RockMongo link is currently timing out, we have a report that it's very old, and the GitHub hasn't been updated in about a year.

Mongo Owl seems to be a proprietary for-pay service, and I'm skeptical that we want to be linking to it.


i80and:
Removed the Mongo Owl bit

PR #277, opened 2015-04-10
Update ruby-driver-tutorial.txt

nickmilon:
Documentation bugs on delete operations


kay-kim:
Hi -- thank you so very much for your fix.  It seems we have already merged in a commit to fix this.  As such, we will close this pull request, but definitely appreciate your taking the time and effort to make the fix.

Regards,

Kay Kim

PR #280, opened 2015-04-20
RUBY-874: Fix deletion docs for Ruby driver

durran:
Docs were wrong on the names of the delete methods in the Ruby driver.


mpaik:
Merged

PR #283, opened 2015-04-27
RUBY-903 Client#use and Client#with return new client

estolfo:


kay-kim:
merged and published.

PR #284, opened 2015-05-05
Mongo Ruby Driver versions > 1.10 do not support ensureIndex

mtowers:
Mongo 2.6 introduced `createIndexes`, whereas prior versions used `ensureIndex`.  With the advent of the ruby driver 1.10, support for `ensureIndex` was dropped.

### DIFF
--- a/source/drivers/ruby.txt
+++ b/source/drivers/ruby.txt
@@ -78,22 +78,22 @@ MongoDB Compatibility
      - Minor incompatibility
    * -
      - ``1.10``
-     - Supported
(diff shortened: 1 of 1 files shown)


kay-kim:
Thank you for the pull request and apologies for the lateness of this response.  We were in the midst of redoing the compatibility tables to now show just the recommended versions.  The redesign on the tables obviates the difficulty of defining what is considered minor, etc.  As such, will close this pull request without merging, but much appreciate your time and efforts.

Regards,

Kay Kim

PR #297, opened 2015-08-18
uppercase Jazz to match genre string in example

elitan:


kay-kim:
Merged.  Thank you.  The updated content will be published in a day or two.

PR #298, opened 2015-08-18
Include Camo project in node-js.txt

scottwrobinson:
I created a class-based ES6 ODM for Mongo called [Camo](https://github.com/scottwrobinson/camo). I think it's a good alternative for ES6 users and for JavaScript novices thanks to its use of ES6 classes. [Here](http://stackabuse.com/introducing-camo-a-class-based-es6-odm-for-mongo-like-databases/) is a short article I wrote on its advantages and how to use it.


kay-kim:
Merged. Will publish the updated content within a day or two. Thanks.

kay-kim:
oops - meant to close and comment (and not just comment)

PR #299, opened 2015-08-22
Be a little bit more descriptive about Morphix's features

ParaPenguin:
I noticed Morphia doesn't have a working cache, and Morphium's hasn't always worked for me. So if people are looking for a library which has a cache, this should help them fine it.


kay-kim:
Thanks.  Merged.  Will republish with your changes in a day or two.

PR #302, opened 2015-09-03
Fix Mongo Logger level configuration

raphaelabreu1:
According the spec_helper, the configuration is:

Mongo::Logger.logger.level = Logger::DEBUG


kay-kim:
Thank you so much for this fix and sincere apologies for the delay in response. It turns out that it was done as part of JIRA ticket DOCS-6164.  As such, I will go ahead and close this pull request.  But once again, we much appreciate your taking the time to find and fix this.

Regards,

Kay Kim

PR #307, opened 2015-09-17
Add note about configuring orm as Mongoid in Rails

estolfo:


kay-kim:
merged.  will republish later this afternoon.

PR #309, opened 2015-09-25
Update Ruby driver bulk write documentation

estolfo:


kay-kim:
merged (and published)

PR #311, opened 2015-10-01
[Mongoid tutorial] Add note about Spring caching config files for Rails applications

estolfo:
Thanks to @jhn for investigating.


kay-kim:
merged!

PR #312, opened 2015-10-05
Fix Criteria#add_to_set parameter

victorkohl:
Passing two parameters to Criteria#add_to_set will result in a `ArgumentError: wrong number of arguments (2 for 1)` error.


kay-kim:
Thank you.  This has been merged.

PR #314, opened 2015-10-09
Fixing #bit and #push_all parameters

victorkohl:
Criteria#bit and Criteria#push_all expects a hash instead of multiple parameters


kay-kim:
Merged.  Thank you!

PR #315, opened 2015-10-09
Update PHP links and compatibility tables

jmikola:
This adds the next-generation PHP driver ([mongodb-labs/mongo-php-driver-prototype](https://github.com/mongodb-labs/mongo-php-driver-prototype) and [mongodb-labs/mongo-php-hhvm-prototype](https://github.com/mongodb-labs/mongo-php-hhvm-prototype)) to the compatibility tables.

At present, I'm not linking to the next extension at the top of the page; however, we will follow up with another PR as we approach as stable release in the next month.

Some other small fixes are included as well.

/cc @bjori @derickr

### DIFF
--- a/source/drivers/driver-compatibility-reference.txt
+++ b/source/drivers/driver-compatibility-reference.txt
@@ -481,22 +481,32 @@ MongoDB Compatibility
      - MongoDB 2.6
      - MongoDB 3.0
 
-   * - 1.6
+   * - mongodb-1.1
      - |checkmark|
      - |checkmark|
      - |checkmark|
 
-   * - 1.5
+   * - mongodb-1.0
+     - |checkmark|
+     - |checkmark|
+     - |checkmark|
+
+   * - mongo-1.6
+     - |checkmark|
+     - |checkmark|
+     - |checkmark|
+
+   * - mongo-1.5
      - |checkmark|
      - |checkmark|
      - 
 
-   * - 1.4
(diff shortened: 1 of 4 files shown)


source/drivers/driver-compatibility-reference.txt
```diff
@@ -513,37 +523,63 @@ Language Compatibility
...
-   :class: compatibility
+   :class: compatibility-large
 
    * - PHP Driver
      - PHP 5.3
      - PHP 5.4
      - PHP 5.5
      - PHP 5.6
+     - PHP 7.0
+     - HHVM
```
derickr: We're going to have to start to be more specific here too. Right now, it should say "HHVM 3.9".

source/drivers/php.txt
```diff
@@ -121,26 +131,48 @@ Language Compatibility
...
-   :class: compatibility
+   :class: compatibility-large
 
    * - PHP Driver
      - PHP 5.3
      - PHP 5.4
      - PHP 5.5
      - PHP 5.6
+     - PHP 7.0
+     - HHVM
```
derickr: Same here, should be "HHVM 3.9"

source/drivers/driver-compatibility-reference.txt
jmikola: :+1:

kay-kim:
merged!

PR #323, opened 2015-11-06
Links to new PHP driver

adaptivedavid:
Update links to point to new PHP driver.


derickr:
This is superseded by the bigger update from #324 . Thanks though!

PR #339, opened 2016-01-26
MONGOID-3882 Note in docs that model needs dynamic attributes when using counter_cache

estolfo:


schmalliso:
Merged!

schmalliso:
@estolfo Should the version number at the top of the page be updated to `5.1.0`?

estolfo:
Yes, let me open another pull request for that, as it needs to change in the code block as well.
Thanks!

PR #348, opened 2016-02-26
fix path to add MongoDB yum repo

carlbernardo:
While trying to set up MongoDB with my AWS EC2 instance, I found this error. On the current documentation, it states to add a MongoDB yum repo with the following commands:

screen shot 2016-02-25 at 8 07 57 pm

However, that path is not '/etc/yum.repos.d/mongodb.repo. It's actually the following:

screenshot 2016-02-24 21 24 20

I fixed the path on the docs so users don't run into errors of this path not existing while trying to run the commands above.


kay-kim:
Thank you very much.  This has been merged.

PR #349, opened 2016-02-26
MONGOID-4240 Update documentation for Criteria#pluck behavior

estolfo:


mrpoundsign:
This may also need to be in the upgrade notes, as this was a behavior change. I'm not entirely sure of the behavior in Mongoid 4 as we went from Mongoid 3 to Mongoid 5 but it may help people migrating.

estolfo:
Thanks for pointing that out. I've tested on Mongoid 4.0.2 and the behavior matches that of 5.x and of master, so I'm not sure it needs to go in the upgrade notes.

PR #350, opened 2016-03-01
Fixes incorrect method name

dmgarland:
https://github.com/mongodb/mongo-ruby-driver/blob/master/lib/mongo/grid/fs_bucket.rb#L228


kay-kim:
Merged!  Thank you.

PR #354, opened 2016-03-15
Update PHP documentation resources links

jmikola:
https://jira.mongodb.org/browse/DOCS-7427

Installation docs were consolidated upstream. Also, library API reference is contained within the top-level docs, so we can change that title to more accurately reflect the destination page.


jmikola:
Just realized that this created my PR branch on the mongodb repository instead of my fork. Feel free to delete it after merging.

schmalliso:
Merged! Thanks so much!

PR #359, opened 2016-04-26
DOCS-7745 Fix Mongoid tutorial typos

estolfo:


kay-kim:
merged!  Thank you.

source/tutorial/mongoid-documents.txt
```diff
@@ -379,7 +379,7 @@ for use in criteria. For example we may want to write a query like so:
   point = Point.new(12, 24)
   Venue.where(location: point)
 
-Not that when accessing custom fields from the document, you will get a new instance
+Note that when accessing custom fields from the document, you will get a new instance
```
kay-kim: oops -- just noticed this, is it supposed to say "generating a new object from the object from the raw ..."

estolfo: yeah, it should be 

"This is because Mongoid is generating a new object from the raw attributes on each access."

kay-kim:
thank you.  the two objects threw me : )

PR #364, opened 2016-05-17
Have updated the typos

cholpona:
initial models.py worked for me only when I put Comment before Post. And there is typo templateS/admin/base.html (S is missing)


kay-kim:
Thank you!  This has been merged.

Regards,

Kay Kim

PR #368, opened 2016-05-25
Adding capistrano-mongo-sync gem

alexfarrill:
We open sourced a ruby gem to sync your local and remote mongo db's, wanted to share it


kay-kim:
merged.  thank you.

PR #380, opened 2016-08-15
Replace tutorial for mongoengine tublelog with getting started with P…

llvtt:
…yMODM.


kay-kim:
merged!  Also added a redirect so that if someone lands on the old tutorial page, it'll go to the pymodm page.

llvtt:
Nice! Thank you.

PR #382, opened 2016-08-26
C Driver docs have moved to mongoc.org

ajdavis:
C Driver docs have moved to mongoc.org


kay-kim:
merged! and published.

PR #384, opened 2016-09-28
Update C++ driver compatibility matrices

xdg:


kay-kim:
merged!  I know, all 10,000 years later.

PR #385, opened 2016-10-26
Update PHP library and extension doc links

jmikola:
In https://github.com/mongodb/mongo-php-library/pull/275 (for [PHPLIB-230](https://jira.mongodb.org/browse/PHPLIB-230)), I've removed the PHP library's `gh-pages` branch and added redirects to the manual documentation. This was one more place that we were linking to the old docs.

Note: There is no direct replacement for the generated API documentation, but the API reference in the current manual is superior anyway. It made sense to link to the `mongodb` extension documentation in its place.

### DIFF
--- a/source/drivers/php.txt
+++ b/source/drivers/php.txt
@@ -12,12 +12,13 @@ PHP MongoDB Driver
    :depth: 1
    :class: twocols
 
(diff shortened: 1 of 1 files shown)


kay-kim:
merged

PR #399, opened 2017-03-08
Revisions for PHP driver page

jmikola:
See full commit messages for detailed changes.

### DIFF
--- a/source/drivers/php.txt
+++ b/source/drivers/php.txt
@@ -23,17 +23,22 @@ For the official MongoDB PHP Driver reference, see:
 Drivers
 -------
 
-The currently maintained driver is the **mongodb** extension available from
-PECL. This driver can be used stand-alone, although it is very bare-bones. You
-should consider using the driver with the complimentary *PHP library*, which
-implements a more full-featured API on top of the bare-bones driver.
+The currently maintained driver is the `mongodb
+<https://pecl.php.net/package/mongodb>`_ extension available from PECL. This
+driver can be used stand-alone, although it is very bare-bones. You should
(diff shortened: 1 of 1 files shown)


source/drivers/php.txt
```diff
@@ -188,20 +183,6 @@ Language Compatibility
...
-  - `Part 1: Getting Started <http://blog.mongodb.org/post/24960636131/mongodb-for-the-php-mind-part-1>`_
-
-  - `Part 2: Queries and Indexes <http://blog.mongodb.org/post/26903435041/mongodb-for-the-php-mind-part-2>`_
-
-  - `Part 3: Data Modeling <http://blog.mongodb.org/post/35133050249/mongodb-for-the-php-mind-part-3>`_
-
-.. toctree::
-   :titlesonly:
-
-   /drivers/php-libraries
```
jmikola: Before merging, please confirm if removing this link is kosher given that we already have a `:ref:` to the page above. I'm not sure why this redundant link existed in the first place, but I'd like to make sure the build process was not depending on the `toctree` block here.

i80and: `toctree` isn't a link (although it creates one by default): it defines the document structure. If a page isn't listed in any `toctree`, it becomes an orphan and throws a warning during the build.

If you just want to suppress showing the link, then you would use the `:hidden:` option.

jmikola: Since the PHP libraries page already shows up in the left-side navigation panel (on https://docs.mongodb.com/ecosystem/drivers/php/), does that make this redundant?

i80and: It shows up there because of this `toctree`

PR #401, opened 2017-03-17
Update amazon-ec2.txt - Update udev rules for EC2  

nishantb10gen:
As reported by user and confirmed in my tests: 
On EC2 Centos7, this is the correct entry for the readahead. It is missing the "change" part:

ACTION=="add|change", KERNEL=="xvda", ATTR{bdi/read_ahead_kb}="16"


kay-kim:
merged!  Thank you!

PR #404, opened 2017-05-02
DOCS-10084: Pymongo version compatibility is wrong

steveren:


source/includes/driver-compatibility-caveat.rst
```diff
@@ -0,0 +1,6 @@
+.. note::
+
+   The compatibility matrices show which versions of the driver support
+   all of the features of a given major release. Older driver versions
```
kay-kim: Unfortunately, I believe this may not be true that it supports "all" of the features... Hence we just write which drivers are recommended for x version.

steveren:
I changed the wording of the note to be a bit more general about why some driver versions are recommended for use with particular server versions.

source/includes/driver-compatibility-caveat.rst
```diff
@@ -0,0 +1,6 @@
+.. note::
+
+   The compatibility matrices show which versions of the driver are
+   recommended for use with a given major release. Older driver versions
+   may be able to connect to a version of MongoDB which is not officially
+   supported, but not all features will be available.
```
kay-kim: This second sentence needs work.  Your sentence in effect states:

Older driver version can connects to unsupported versions of MongoDB but not all features will be available.

Is that what you meant?

kay-kim: Oh, forgot.  No need to single-source things that are only used in one file.

steveren: Yes, that's correct. It says older versions *may* be able to connect to unsupported versions, which is true and was brought up in DOCS-9400 as well as in my conversations with Shane Harvey about what exactly we mean by compatible and incompatible. He pointed out that old versions of pymongo can still connect to MongoDB version 3.4 and perform some kinds of operations, and that it's frequently possible to upgrade a MongoDB server without upgrading the driver (or changing application code) without any ill effects. He mentioned that they test PyMongo 2.9 against MongoDB 2.4, 2.6, 3.0, 3.2, 3.4, and 3.5, even though it's only officially supported for 3.0 and lower.

All that being said, if you think this paragraph is going to confuse more people than it helps, I'm happy to spike it altogether and leave this with just the minimum viable ticket, which is correcting the python.txt page so that its compatibility matrix agrees with the main driver compatibility page.

steveren: I single-sourced it because Shane and I discussed the possibility of including the caveat paragraph on each of the individual driver pages and I wanted to leave the possibility open, but if we're only going to use it on the one page (the driver compatibility reference) I can inline it.

kay-kim: If old versions of pymongo can still connect to MongoDB 3.4 -- well, MongoDB 3.4 is a supported version.   The phrasing seems to suggest that MongoDB does not support MongoDB 3.4.

kay-kim: If we're going to include it on each page, then by all means single-source.

steveren: AH YES ok I see what you mean, sorry for being dense. I'll make a new version of that sentence momentarily.

steveren: If we arrive at a version of the paragraph we're happy with, I can include it on each of the driver pages and update this PR.

source/includes/driver-compatibility-caveat.rst
```diff
@@ -0,0 +1,7 @@
+.. note::
+
+   The compatibility matrices show which versions of the driver are
+   recommended for use with a given major release. Older driver versions
+   may be able to connect to a version of MongoDB which is not
+   recommended for that driver version, but not all
```
kay-kim: change ", but" to "; however, ..."

steveren:
Last requested change is done.

kay-kim:
merged!

PR #406, opened 2017-05-24
Updated compatibility reference for Scala drivers

rozza:
DOCS-10296


rozza:
Note: I was unable to get `make html` work, so haven't visually inspected the changes.

kay-kim:
merged!  Will fix any restructured issues on our end and publish.

Thanks much.

PR #414, opened 2017-10-13
DOCS-10895: Remove outdated mgo & Erlang driver standalone pages

stennie:
- remove mgo & Erlang pages from navigation
 - add mgo (GlobalSign fork) which is currently more actively maintained than the original mgo


kay-kim:
merged!

PR #424, opened 2017-11-29
Updated C#/.NET compatibility tables. (DOCS-11023)

rob-mongodb:
Staged at https://docs-mongodbcom-staging.corp.mongodb.com/ecosystem/robertjustice/DOCS-11023/drivers/csharp.html.

### DIFF
--- a/source/drivers/csharp.txt
+++ b/source/drivers/csharp.txt
@@ -72,12 +72,19 @@ MongoDB Compatibility
      - MongoDB 3.2
      - MongoDB 3.4
 
+   * - Version 2.4
+     - |checkmark|
+     - |checkmark|
+     - |checkmark|
+     - |checkmark|
+     - |checkmark|
+
    * - Version 2.3
      - |checkmark|
      - |checkmark|
      - |checkmark|
      - |checkmark|
-     - |checkmark|
+     - 
 
    * - Version 2.2
      - |checkmark|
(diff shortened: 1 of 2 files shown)


rob-mongodb:
@schmalliso , good catch! Staging is updated and ready for another look.

source/drivers/csharp.txt
```diff
@@ -127,6 +134,12 @@ MongoDB Compatibility
      - .NET 4.5
      - .NET Core
 
+   * - Version 2.4
```
schmalliso: This isn't appearing on staging, currently -- not sure why. Assuming it does build correctly, LGTM!

schmalliso:
One comment - LGTM assuming it all builds correctly!

rob-mongodb:
@jyemin , this small update is ready for a technical review.

jyemin:
After discussing with @rstam, we decided it would be good to explicitly mention all the versions of .NET Core, as we do with .NET Framework. So the table would look like:

```
Driver Version ...        .NET Core 1.0 | .NET Core 1.1 | .NET Core 2.0
2.4                                x    |         x     |        x 
2.3                                x    |         x     |        x 
```

rob-mongodb:
@jyemin , staging is updated and ready for another look.

source/drivers/csharp.txt
```diff
@@ -125,37 +132,57 @@ MongoDB Compatibility
      - .NET 3.5
      - .NET 4.0
      - .NET 4.5
-     - .NET Core
+     - .NET Core 1.1
```
jyemin: Re-order so that 1.0 is before 1.1

rob-mongodb:
@kay-kim , this is ready for merge into 3.6 and master.

PR #427, opened 2018-01-06
DOCS-11171: update Node.js driver compatibility table to include Mong…

steveren:
…oDB 3.6

---
This change is [Reviewable](https://reviewable.io/reviews/mongodb/docs-ecosystem/427)

### DIFF
--- a/source/drivers/driver-compatibility-reference.txt
+++ b/source/drivers/driver-compatibility-reference.txt
@@ -516,8 +516,10 @@ MongoDB Compatibility
      - MongoDB 3.0
      - MongoDB 3.2
      - MongoDB 3.4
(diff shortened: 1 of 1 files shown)


source/drivers/driver-compatibility-reference.txt
```diff
@@ -516,30 +516,35 @@ MongoDB Compatibility
      - MongoDB 3.0
      - MongoDB 3.2
      - MongoDB 3.4
+     - MongoDB 3.6
```
rkumar-mongo: Should we include a row for the 3.0 driver? Or can we generally assume it has the same compatibilities as 2.2.12?

kay-kim:
duplicate

PR #441, opened 2018-04-30
Use consistent labels for PHP server/language compat tables

jmikola:
We recently had a question from a support associate about PHPLIB language compatibility. This should address that.

---
This change is [Reviewable](https://reviewable.io/reviews/mongodb/docs-ecosystem/441)

### DIFF
--- a/source/includes/extracts-driver-compatibility-matrix.yaml
+++ b/source/includes/extracts-driver-compatibility-matrix.yaml
@@ -196,6 +196,8 @@ post: |
    In the table below, `mongodb <https://pecl.php.net/package/mongodb>`_ and
    `mongo <https://pecl.php.net/package/mongo>`_ refer to the new and legacy
    {{driver}}, respectively.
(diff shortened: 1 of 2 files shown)


source/includes/extracts-driver-compatibility-matrix.yaml
```diff
@@ -196,6 +196,8 @@ post: |
    In the table below, `mongodb <https://pecl.php.net/package/mongodb>`_ and
    `mongo <https://pecl.php.net/package/mongo>`_ refer to the new and legacy
    {{driver}}, respectively.
+   `PHPLIB <https://github.com/mongodb/mongo-php-library>`_ refers to the
+   userland library.
```
jmikola: This makes the `post` for `php-driver-compatibility-matrix-language` identical to the one for `php-driver-compatibility-matrix-mongodb`. If you have a better way to reduce the duplication and have them share one snippet, that'd be great.

PR #462, opened 2018-11-05
DOCSP-3333: Go Driver Docs

skerschb:
Placeholder for Go Driver Docs.

### DIFF
--- a/source/drivers/go.txt
+++ b/source/drivers/go.txt
@@ -1,0 +1,61 @@
+========================
+MongoDB Go Driver (beta)
+========================
+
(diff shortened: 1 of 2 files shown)


source/drivers/go.txt
```diff
@@ -0,0 +1,71 @@
+<<<<<<< HEAD
```
ravanscafi: I guess that you merged some conflicts.

skerschb: this should not be merged at the moment. I will revoke the PR.

PR #474, opened 2018-12-14
Remove PHP 5.4 and add 7.3 in compat table

jmikola:
https://jira.mongodb.org/browse/PHPC-1175

PHP 7.3 was recently released. Driver version 1.5.0+ is compatible. I've also removed PHP 5.4 since none of the listed driver versions support it, and 5.4 itself is very much end-of-life.

PR #475, opened 2018-12-14
Add PHP 7.3 to language compat table

jmikola:
This is related to https://github.com/mongodb/docs-ecosystem/pull/474.

What whatever reason, we have different compatibility tables on the PHP-specific and cross-language pages. Here, I've left PHP 5.4 in place since it's still relevant to old driver versions present in this table.


schmalliso:
Also merged! You are very admirably on top of things, good sir!

PR #476, opened 2018-12-14
Update post blocks for PHP server and language compat tables

jmikola:
The legacy mongo extension is not present in either of these tables and can be removed. The userland library isn't present in the language compat table and can be removed.

### DIFF
--- a/source/includes/extracts-driver-compatibility-matrix.yaml
+++ b/source/includes/extracts-driver-compatibility-matrix.yaml
@@ -180,7 +180,7 @@ replacement:
   driver: MongoDB PHP driver
 post: |
    In the table below, `mongodb <https://pecl.php.net/package/mongodb>`_ and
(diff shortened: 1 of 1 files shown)


source/includes/extracts-driver-compatibility-matrix.yaml
```diff
@@ -180,10 +180,8 @@ replacement:
   driver: MongoDB PHP driver
 post: |
    In the table below, `mongodb <https://pecl.php.net/package/mongodb>`_ and
-   `mongo <https://pecl.php.net/package/mongo>`_ refer to the new and legacy
-   {{driver}}, respectively.
-   `PHPLIB <https://github.com/mongodb/mongo-php-library>`_ refers to the
-   userland library.
+   and `PHPLIB <https://github.com/mongodb/mongo-php-library>`_ refer to the
```
derickr: There are two "and"s in a row now.

schmalliso:
Merged!

PR #482, opened 2019-01-21
The current version of ruby-bson is 4.x, redirect by default there

p-mongo:


kay-kim:
It seems as though Sue has already made the change -- https://github.com/mongodb/docs-ecosystem/commit/9b63f743b0f3d490cb1aa5923ea505dd4e52d0bc
So, will close.

PR #492, opened 2019-05-03
Update community-supported-drivers.txt

zookzook:
I have updated the list of drivers for Elixir.

### DIFF
--- a/source/drivers/community-supported-drivers.txt
+++ b/source/drivers/community-supported-drivers.txt
@@ -71,8 +71,8 @@ Community Supported Drivers Reference
   - `Elixir Driver <https://github.com/ericmj/mongodb>`_ - MongoDB
     driver in Elixir, a functional language built on top of the Erlang VM.
 
-  - `Elixir Driver <https://github.com/zookzook/elixir-mongodb-driver>`_ - An alternative MongoDB
-    driver for Elixir based on the work of `Elixir Driver <https://github.com/ericmj/mongodb>`_
+  - `Elixir Driver <https://github.com/zookzook/elixir-mongodb-driver>`__ - An alternative MongoDB
+    driver for Elixir based on the work of `Elixir Driver <https://github.com/ericmj/mongodb>`__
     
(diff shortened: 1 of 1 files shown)


zookzook:
I have signed the agreement.

source/drivers/community-supported-drivers.txt
```diff
@@ -70,6 +70,9 @@ Community Supported Drivers Reference
 
   - `Elixir Driver <https://github.com/ericmj/mongodb>`_ - MongoDB
     driver in Elixir, a functional language built on top of the Erlang VM.
+
+  - `Elixir Driver <https://github.com/zookzook/elixir-mongodb-driver>`_ - An alternative MongoDB
```
terakilobyte: One last small set of changes. Can you add an additional `_` to the end of your links? This is for our end to avoid warnings.

```
`Elixir Driver <https://github.com/zookzook/elixir-mongodb-driver>`__
```

And an additional `_` to the second link in your PR.

Thanks much!

zookzook: To be honest, I do not understand why I should put another `_`. All links in the `community-supported-drivers.txt` file have always only one `_`. Can you enlighten me?

Is this right?

```
  - `Elixir Driver <https://github.com/zookzook/elixir-mongodb-driver>`__ - An alternative MongoDB
    driver for Elixir based on the work of `Elixir Driver <https://github.com/ericmj/mongodb>`__
````

terakilobyte: What you have is correct with the double underscores.

This creates a "one-off" URL rather than an explicit target. Really what this means is that our docs build system doesn't try to duplicate ids, which it was doing for `#elixir-driver`. By using the `__` after the links we don't create an in page reference to the second one. This won't impact discoverability in any way, and doesn't lessen how awesome Elixir or the driver is.

zookzook:
Ok, I added the second '_' to each link. Thanks for the explanation and for your patience.

zookzook:
I did push the requested changes. Is there any more to do?

PR #510, opened 2019-06-25
Fix Python console code for Inventory use case

gianpaj:
See an example from the "MongoDB Applied Design Patterns" book:
https://books.google.com.ua/books?id=t2dBlNfKk6gC&lpg=PA63&ots=8xxoVBJX5n&dq=collectionsharded%20ok&pg=PA63#v=onepage&q=collectionsharded%20ok&f=false

### DIFF
--- a/source/use-cases/inventory-management.txt
+++ b/source/use-cases/inventory-management.txt
@@ -460,12 +460,12 @@ cart and inventory collections:
 
 .. code-block:: pycon
 
-   >>> db.command('shardCollection', 'inventory'
-   ...            'key': { '_id': 1 } )
-   { "collectionsharded" : "inventory", "ok" : 1 }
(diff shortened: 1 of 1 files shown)


ccho-mongodb:
The Inventory Management use case has been removed from this repo.

gianpaj:
I see, thanks. Could you point me to where it lives now?

I can't find it searching on Google or Github.

These use cases and examples are very useful for future reference.

Thanks

ccho-mongodb:
They've been transferred to another team and may only be used if they find use in republishing them. Those documents were not being maintained and did not fit into our new navigation structure. For recent documentation that is actively maintained, consider using the documentation here:
https://pymongo.readthedocs.io/en/stable/tutorial.html

PR #519, opened 2019-07-16
Update compatibility matrices for 4.2

terakilobyte:
Please review. This updates the compatibility tables according to DOCSP-6230
Rendered Views:
- [x] [C](https://docs-mongodbcom-staging.corp.mongodb.com/ecosystem/docsworker/DOCSP-6230/drivers/c.html#compatibility)
- [x]  [C++](https://docs-mongodbcom-staging.corp.mongodb.com/ecosystem/docsworker/DOCSP-6230/drivers/cxx.html#compatibility)
- [x] [C#](https://docs-mongodbcom-staging.corp.mongodb.com/ecosystem/docsworker/DOCSP-6230/drivers/csharp.html#compatibility)
- [ ] [Go](https://docs-mongodbcom-staging.corp.mongodb.com/ecosystem/docsworker/DOCSP-6230/drivers/go.html#compatibility)
- [x] [Java](https://docs-mongodbcom-staging.corp.mongodb.com/ecosystem/docsworker/DOCSP-6230/drivers/java.html#compatibility)
- [x]  [Node](https://docs-mongodbcom-staging.corp.mongodb.com/ecosystem/docsworker/DOCSP-6230/drivers/node.html#compatibility)
- [x]  [Perl](https://docs-mongodbcom-staging.corp.mongodb.com/ecosystem/docsworker/DOCSP-6230/drivers/perl.html#compatibility)
- [x]  [PHP](https://docs-mongodbcom-staging.corp.mongodb.com/ecosystem/docsworker/DOCSP-6230/drivers/php.html#compatibility)
- [x] [PyMongo](https://docs-mongodbcom-staging.corp.mongodb.com/ecosystem/docsworker/DOCSP-6230/drivers/pymongo.html#compatibility)
- [x] [Scala](https://docs-mongodbcom-staging.corp.mongodb.com/ecosystem/docsworker/DOCSP-6230/drivers/scala.html#compatibility)

The docset for Ruby is built from that repo. I've submitted a [PR](https://github.com/mongodb/mongo-ruby-driver/pull/1444) against the 2.10 branch.


source/includes/language-compatibility-table-php.rst
```diff
@@ -13,6 +13,15 @@
...
 
+   * - mongodb-1.6
+     - |checkmark|
+     - |checkmark|
+     - |checkmark|
+     - |checkmark|
+     - |checkmark|
+     - |checkmark|
+
+
```
jmikola: Extra whitespace.

source/includes/mongodb-compatibility-table-c.rst
```diff
@@ -4,14 +4,26 @@
...
    * - C Driver Version
+     - MongoDB 4.2
      - MongoDB 4.0
      - MongoDB 3.6
      - MongoDB 3.4
      - MongoDB 3.2
      - MongoDB 3.0
      - MongoDB 2.6
 
+   * - Version 1.15
```
jmikola: Was skipping 1.12 through 1.14 (inclusive) intentionally?

In the past, Version 1.8 and Version 1.7 were both included even though they support the same server versions. Would it make more sense to denote a range of versions in the first column (e.g. "Version 1.11 - 1.14") and consolidate rows with identical checkmarks?

source/includes/mongodb-compatibility-table-c.rst
```diff
@@ -20,71 +32,80 @@
...
+     -
      -
      -
      - |checkmark|
      - |checkmark|
      - |checkmark|
 
    * - Version 1.1.0
-     - 
+     -
```
jmikola: While you're updating this, I think you can simply drop the tailing ".0" from 1.1.0 and 1.0.0.

source/includes/mongodb-compatibility-table-cxx.rst
```diff
@@ -12,7 +13,18 @@
      - MongoDB 2.6
      - MongoDB 2.4
 
+   * - mongocxx 3.5.x
```
jmikola: I'd suggest dropping the ".x" from each version in this table. Also, other tables tend to report "Version X.Y" instead of repeating the driver name ("mongocxx" in this case). PHP is a notable exception to that because we have both a low- and high-level driver and it helps to differentiate.

source/includes/mongodb-compatibility-table-go.rst
```diff
@@ -4,14 +4,25 @@
...
    * - Go Driver
+     - MongoDB 4.2
      - MongoDB 4.0
      - MongoDB 3.6
      - MongoDB 3.4
      - MongoDB 3.2
      - MongoDB 3.0
      - MongoDB 2.6
 
+   * - 1.1.0
```
jmikola: I suggest dropping trailing ".0" here as well, and any other places where a placeholder for the patch version is used (be it ".0" or ".x"). Patch versions should be irrelevant as far as server compatibility is concerned.

source/includes/language-compatibility-table-scala.rst
```diff
@@ -8,6 +8,10 @@
      - Scala 2.11
      - Scala 2.12
 
+   * - 2.7
```
rozza: Scala driver 2.7 also adds Scala 2.13 support - can that be added?

source/includes/mongodb-compatibility-table-c.rst
terakilobyte: @jmikola There's no good reason, it just looks like that info was missing. I've updated to include missing versions.

I'm not specifically against doing a version range but it won't be consistent across the rest.

source/includes/language-compatibility-table-php.rst
```diff
@@ -6,12 +6,12 @@
...
-     - PHP 5.6
-     - PHP 7.0
-     - PHP 7.1
-     - PHP 7.2
      - PHP 7.3
+     - PHP 7.2
+     - PHP 7.1
+     - PHP 7.0
+     - PHP 5.6
+     - PHP 5.5
```
jmikola: Extension 1.6 will not have support for PHP 5.5 (see: [PHPC-1377](https://jira.mongodb.org/browse/PHPC-1377)).

I realize we don't report language compatibility for PHPLIB (although I suggest doing so in [DOCSP-6259](https://jira.mongodb.org/browse/DOCSP-6259)), but library version 1.5 also drops support for PHP 5.5 ([PHPLIB-429](https://jira.mongodb.org/browse/PHPLIB-429)).

terakilobyte: I've made significant updates to the PHP pages to incorporate the other ticket in JIRA

source/drivers/php.txt
```diff
@@ -17,44 +17,56 @@ MongoDB PHP Driver
...
+- `Library Architecture <https://www.php.net/manual/en/mongodb.architecture.php>`__
 
-- `API Reference <https://php.net/manual/en/book.mongodb.php>`_
+- API Reference
 
-- `Changelog <https://github.com/mongodb/mongo-php-driver/releases>`__
+  - `Library <https://www.php.net/manual/en/set.mongodb.php>`__
 
-- `Source Code <https://github.com/mongodb/mongo-php-driver>`__
+  - `Extension <https://php.net/manual/en/book.mongodb.php>`__
```
jmikola: This links directly to the API reference classes directly under the `MongoDB\Driver` namespace, but it omits other classes in the extension, such as:

 * https://www.php.net/manual/en/book.bson.php
 * https://www.php.net/manual/en/mongodb.monitoring.php
 * https://www.php.net/manual/en/mongodb.exceptions.php

Unlike the library, we don't have a common deep link for all API reference, so I think it'd be best to just use the main documentation page here instead: https://www.php.net/manual/en/set.mongodb.php. Although the page lists articles first, all classes in the extension are linked directly from that page, too.

And similar to my earlier suggestion, you may want to use https://php.net/mongodb instead, since it's a shorter URL and points to the same entry point for the extension docs.

source/drivers/php.txt
```diff
@@ -17,44 +17,56 @@ MongoDB PHP Driver
...
+  - `Extension <https://github.com/mongodb/mongo-php-driver>`__
 
 
 Installation
 ------------
 
-You can install the driver using 
-`PECL <https://www.php.net/manual/en/mongodb.installation.pecl.php>`_:
+The preferred method of installing the PHP library is with
+`Composer <https://getcomposer.org/>`_ by running the following from your project root:
```
jmikola: I appreciate the sentiment of instructing users to install the library up front; however, if they actually try to run these steps in the order presented this will fail since the library requires the extension to already be installed.

It may make sense to revert to showing the extension installation step first.

source/drivers/php.txt
```diff
@@ -99,7 +111,11 @@ How to get help
 - Join our `Google Group <https://groups.google.com/forum/#!forum/mongodb-user>`__.
 - Ask on `Stack Overflow <https://stackoverflow.com/questions/tagged/mongodb+php>`__.
 - Visit our `Support Channels <https://docs.mongodb.com/manual/support/>`__.
-- See our `JIRA <https://jira.mongodb.org/projects/PHP/summary>`__ to raise issues or request features.
+- See our JIRA to raise issues or request features.
+
+  - `Extension <https://jira.mongodb.org/projects/PHPC/summary>`__
+
+  - `Library <https://jira.mongodb.org/projects/PHPLIB/summary>`__
```
jmikola: Thanks for linking to these pages! I just realized that the descriptions in JIRA still report these as experimental projects and actually point users to use the unmaintained legacy driver! Will fix that immediately.

source/drivers/php.txt
```diff
@@ -17,44 +17,56 @@ MongoDB PHP Driver
 Introduction
 ------------
 
-The currently maintained driver is the `mongodb
-<https://pecl.php.net/package/mongodb>`_ extension available from PECL.
+The PHP driver consists of two components, the MongoDB `extension <https://github.com/mongodb/mongo-php-driver>`_
+and `library <https://docs.mongodb.com/php-library/current/>`_.
```
jmikola: An alternative thought came to me while looking at the PHP table changes below. Rather than link to the documentation for each project here, would it make sense to link to the GitHub repositories themselves (as you do in the tables)? If so, we could then change "API Reference" in the bullet list below to "Documentation". That also may make more sense given that the extension doesn't have a deep link _just_ to API reference, and we need to link to the doc homepage anyway. For the library, I don't think we're losing much by linking to the homepage instead of the reference link directly, since the latter is easily accessible from the main menu.

source/includes/extracts-driver-compatibility-matrix.yaml
```diff
@@ -179,9 +179,9 @@ inherit:
 replacement:
   driver: MongoDB PHP driver
 post: |
-   In the table below, `mongodb <https://pecl.php.net/package/mongodb>`_ and
-   `PHPLIB <https://github.com/mongodb/mongo-php-library>`_ refer to the
-   {{driver}} and userland library, respectively.
+   In the table below, `ext <https://pecl.php.net/package/mongodb>`_ and
```
jmikola: Noting that the PECL website was quite slow, perhaps you want to use https://github.com/mongodb/mongo-php-driver here instead? That would be consistent with linking to the library's GitHub repo as well.

FYI: PECL is really the package manager entry. The equilvalent library link is https://packagist.org/packages/mongodb/mongodb; however, I don't think we need to point to that, since it mainly presents data extracted from the GitHub repository.

source/includes/extracts-driver-compatibility-matrix.yaml
```diff
@@ -191,8 +191,9 @@ replacement:
   driver: MongoDB PHP driver
   language: PHP
 post: |
-   In the table below, `mongodb <https://pecl.php.net/package/mongodb>`_
-   refers to the {{driver}}.
+   In the table below, `ext <https://pecl.php.net/package/mongodb>`_ and
```
jmikola: Likewise, may want to use https://github.com/mongodb/mongo-php-driver here.

source/drivers/php.txt
terakilobyte: I'm guessing this doesn't relate to the source code link, but rather the Extension API ref. I've updated that link as well as the initial extension link.

source/drivers/php.txt
terakilobyte: I like linking to their github repos. I'm hesitant to change it from API to documentation because it doesn't follow the format of the other pages. We'd like to move to a standard template for driver pages, while leaving enough room for driver specifics.

jmikola: In this case, should "library" link to https://github.com/mongodb/mongo-php-library then, so both links are consistent?

terakilobyte: actually I think this is fine looking at it

source/drivers/php.txt
jmikola: Correct, I'm referring to:

```
Extension <https://php.net/manual/en/book.mongodb.php>`__
```

....from the new side of the diff.

source/includes/extracts-driver-compatibility-matrix.yaml
terakilobyte: Updated this to point to github

source/drivers/php.txt
terakilobyte: Makes sense, I'll get this change in.

PR #521, opened 2019-08-12
local master key section for CSFLE Use Case

ccho-mongodb:
## Summary
Add description and code snippets for the Client Side Field Level Encryption feature, local master key section.

## JIRA
https://jira.mongodb.org/browse/DOCSP-6486

## Staged
https://docs-mongodbcom-staging.corp.mongodb.com/ecosystem/docsworker/DOCSP-6486/use-cases/local-master-key.html


source/use-cases/local-master-key.txt
```diff
@@ -0,0 +1,89 @@
...
+  }
+
+  // Specify a local master key in the KMS provider settings
+  Map<String, Map<String, Object>> kmsProviders = new HashMap<String, Map<String, Object>>() {{
+    put("local", new HashMap<String, Object>() {{
+        put("key", localMasterKey);
+    }});
+  }};
+
+The ``kmsProviders`` Map defined above will be used later when specifying the KMS in the encryption settings for the client. At this stage, you should now have a **master key** in a file and code to load it into a Map that will be appended to the configuration of your client.
```
cbush: "be used" -> Avoid passive voice. You also repeat this information in the last part: "a Map that will be appended to the configuration of your client". Is there a way to merge these?

Consider removing "At this stage" since we are already at the end of the document and the phrase doesn't seem to be conveying much information. "Should" could also go since they just followed the instructions.

source/use-cases/local-master-key.txt
```diff
@@ -0,0 +1,89 @@
+A. Create a Local Master Key
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Client-Side Field Level Encryption for MongoDB uses the envelope encryption mechanism in which there are two types of keys: 1) **data encryption keys** which are used to encrypt/decrypt the sensitive data fields and 2) a **master key** that is used to encrypt the **data encryption keys**.
+
+The **master key** is stored separately in a Key Management System (KMS) in order to prevent the **data encryption keys** from being used directly to access the sensitive data.
```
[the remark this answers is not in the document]
MongoCaleb: +1 on Bush's comment

source/use-cases/local-master-key.txt
```diff
@@ -0,0 +1,89 @@
...
+       name: "Windows 8 and older Powershell"
+       content: |
+
+         .. code-block:: shell
+
+           powershell -command "$r=[byte[]]::new(64);$g=[System.Security.Cryptography.RandomNumberGenerator]::Create();$g.GetBytes($r);[Convert]::ToBase64String($r)" | Out-File -FilePath .\master-key.txt
+
+Next, in our MongoDB client application, add the following logic to read the **master key** from the file we created and construct the KMS settings:
+
+.. code-block:: java
```
MongoCaleb: You need to add the tabs here, too, and have code snippets for each language, no?

source/use-cases/local-master-key.txt
```diff
@@ -0,0 +1,89 @@
+A. Create a Local Master Key
```
MongoCaleb: If this is intended to be the complete page, get rid of the "A" header, since it's the only "step".

source/use-cases/local-master-key.txt
```diff
@@ -0,0 +1,89 @@
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Client-Side Field Level Encryption for MongoDB uses the envelope encryption mechanism in which there are two types of keys: 1) **data encryption keys** which are used to encrypt/decrypt the sensitive data fields and 2) a **master key** that is used to encrypt the **data encryption keys**.
```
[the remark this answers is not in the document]
MongoCaleb: +1

source/use-cases/local-master-key.txt
ccho-mongodb: The current direction I think this page (open for debate) will take is that the entire guide (of which this is a single section) will be composed for a single language, and that a selector will allow the user to switch between the available languages. If the individual sections had language selectors, this could be confusing to the user when they get to the next code snippet and it is not in the same language that they had selected prior.

I think your comment surfaces a good point that providing multiple options could be confusing and instead we should steer the user towards doing everything one way. The prior tabbed block just shows users how to create a file which can be accomplished on command line or through the language's file management sdk.

Given the above explanation, what are your thoughts on whether to keep the prior code block? @MongoCaleb

source/use-cases/local-master-key.txt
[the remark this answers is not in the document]
ccho-mongodb: Do you think it would be best to stop using bold prior to or after Line 6?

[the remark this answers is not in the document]
cbush: I would stop using it after the first time each phrase is introduced. So after "master key" on line 4.

source/use-cases/local-master-key.txt
ccho-mongodb: This will be part of a Use Case guide. Nick created the document template here: https://github.com/mongodb/docs-ecosystem/commit/4f47fece63aeab1145d7ae5bb8ea6e73ea32d2b6

I suppose I could branch from Nick's branch but am tentative in doing so, given the above template hasn't gone through the PR review process yet. What would you suggest?

source/use-cases/local-master-key.txt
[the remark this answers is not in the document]
ccho-mongodb: Good idea, I will reformat as a list.
I'm waiting for Ravind's documentation to see whether there will be any explanation of envelope encryption that I can link to, but in the meantime, should I link to a Wikipedia article or try to explain it in this Use Case guide?

source/use-cases/local-master-key.txt
ccho-mongodb: Thanks, I was attempting to summarize the concepts that they should have picked up by reading through that section (similar to post-conditions). The last sentence talks about both the master key generation as well as setting values in a Map object. Would it be better to:

1) Omit and instead use the summary section at the end of this Use Case document as templated by Nick to : https://github.com/mongodb/docs-ecosystem/commit/4f47fece63aeab1145d7ae5bb8ea6e73ea32d2b6 or otherwise start a new paragraph

2) Create a separate paragraph and reword it to more clearly establish what artifacts the reader has now created

3) Omit the recap info completely

And a slightly off-topic question:
Do we use "we" or "you"? Related discussion here:
https://ux.stackexchange.com/questions/57305/we-vs-you-in-documentation

source/use-cases/local-master-key.txt
```diff
@@ -0,0 +1,89 @@
...
+
+The **master key** is stored separately in a Key Management System (KMS) in order to prevent the **data encryption keys** from being used directly to access the sensitive data.
+
+The **data encryption keys** can safely be stored in a MongoDB collection alongside the encrypted sensitive data fields since they have been encrypted by the **master key**.
+
+In Client-Side Field Level Encryption, a client that has access to both **master key** and **data encryption keys** is necessary to perform encryption/decryption.
+
+In the code example below, we will set up the **master key** in a local KMS store as a file. This setup should only be used for local testing purposes and never for production. A secure service such as AWS KMS should always be used in production environments to ensure the **master key** is not compromised.
```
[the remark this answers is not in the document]
ccho-mongodb: Good idea. Do you think this should be emphasized in a "Note" box to make it more conspicuous or would it be sufficient just to reword?

source/use-cases/local-master-key.txt
MongoCaleb: The tabs are designed so that, when you choose a language in the first example on the page, any other tabs on the page also switch to that language. That being said, it sounds like there may be a bigger picture to consider, and I'll defer to Ravind and/or Nick on that.

source/use-cases/local-master-key.txt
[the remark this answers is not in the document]
ccho-mongodb: Will do, thanks.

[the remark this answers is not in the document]
ccho-mongodb: Thanks, will update.

source/use-cases/local-master-key.txt
nlarew: Tabs are an area that we should do user research on, but for now let's think about these questions:
- Do users expect tabbed code blocks or do they think on a full-page level?
- Is the page a single piece of content (in this case probably yes) or will users follow links to specific sub-sections of this page (e.g. to view a specific example/step)

Given that this guide is more of a "top to bottom" flow, we could probably just use pills at the top and have non-tabbed code blocks. That said, I have an inkling that users will get confused so maybe we should also tab each code block individually to be safe.

For now, I don't think we need to have an answer to this - we can decide for the page holistically once the other pieces are compiled.

source/use-cases/local-master-key.txt
nlarew: nit: we will set up -> we set up

Generally we want to avoid future imperfect tense in favor of present tense.

source/use-cases/local-master-key.txt
nlarew: In general you > we.

The rough rule of thumb that I use:
- **we** is for discussing things that the user is provided by the docs
   - e.g. a code block that we walk through - "In this code, we first do x then y."
- **you** is for imperative commands and advisories
   - e.g. a specific direction - "(You) configure the baz to..." (the "you" is often implied in imperative mood)
   - e.g. an optional check - "You should ensure that..."
   - e.g. a note/warning - "You must enable someSetting before foo"

source/use-cases/local-master-key.txt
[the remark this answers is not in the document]
ccho-mongodb: Sure, will add some info since Ravind is not creating any docs on this.

source/use-cases/local-master-key.txt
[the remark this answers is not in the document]
ccho-mongodb: Thanks, will update.

source/use-cases/local-master-key.txt
ccho-mongodb: Sounds like the addition of showing users how to create a local key file by just using command line tools would overcomplicate the UX. I'm satisfied with steering the user towards running a separate application to write the local key file, but would definitely want to discuss adding language-independent steps to language-specific guides at some point in the future.

source/use-cases/local-master-key.txt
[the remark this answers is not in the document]
ccho-mongodb: I will link to the AWS article and clarify the singular master key.

OT: Is `dinogrind` an example of an `encryption sausage`?

source/use-cases/local-master-key.txt
```diff
@@ -0,0 +1,89 @@
...
+
+
+Generate the local **master key** and save it to a file. You should only need to run this once:
```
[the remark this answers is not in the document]
ccho-mongodb: What is a "stepfile" and are there any examples? Do you just mean that it should have headings and/or ToC? Or just enumerated lists?

source/use-cases/local-master-key.txt
ccho-mongodb: Thanks!

source/use-cases/local-master-key.txt
```diff
@@ -1,89 +1,23 @@
...
+Client-Side Field Level Encryption for MongoDB uses an encryption strategy called envelope encryption in which keys used to encrypt/decrypt data (called **data encryption keys**) are encrypted with another key (called the **master key**). 
 
-The **master key** is stored separately in a Key Management System (KMS) in order to prevent the **data encryption keys** from being used directly to access the sensitive data.
+This strategy bolsters the security, minimizes encryption/decryption overhead, and allows the data encryption keys to be stored in the same replica set as the encrypted data. If a basic encryption strategy was used instead of envelope encryption, the computational and network load of encryption and decryption would need to be handled by the system that manages your keys. For more information on envelope encryption and key management concepts, read `this AWS article <https://docs.aws.amazon.com/kms/latest/developerguide/concepts.html#enveloping>`_.
```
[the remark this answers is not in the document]
nlarew: nit: we should use the name of an article when we link to it, if reasonable

```read this AWS article``` ->
```rst
refer to the `envelope encryption <https://docs.aws.amazon.com/kms...>`_ overview in the AWS docs.
```

source/use-cases/local-master-key.txt
```diff
@@ -1,89 +1,23 @@
...
 
-The **data encryption keys** can safely be stored in a MongoDB collection alongside the encrypted sensitive data fields since they have been encrypted by the **master key**.
+The master key is stored separately in a Key Management System (KMS) in order to prevent the data encryption keys from being used directly to access the sensitive data.
 
-In Client-Side Field Level Encryption, a client that has access to both **master key** and **data encryption keys** is necessary to perform encryption/decryption.
+The data encryption keys can safely be stored in a MongoDB collection alongside the encrypted sensitive data fields since they have been encrypted by the master key.
```
nlarew: These two sentences both seem out of place - we just finished up a paragraph about envelope encryption and referred users to an outside guide, then these carry on talking about envelope encryption concepts.

Can we make these more contextually relevant to MongoDB FLE specifically rather than envelope encryption?
- Should these move to just after the opening sentence (i.e. above the previous paragraph)?
- Should we combine these with the previous paragraph? "alongside the encrypted data" seems like an area that envelope encryption gets a big win vs basic.

Also (nit) they both use passive voice but should preferably be active.

source/use-cases/local-master-key.txt
```diff
@@ -1,89 +1,23 @@
...
 
-In the code example below, we will set up the **master key** in a local KMS store as a file. This setup should only be used for local testing purposes and never for production. A secure service such as AWS KMS should always be used in production environments to ensure the **master key** is not compromised.
+In Client-Side Field Level Encryption, a client that can access both master key and data encryption keys is necessary to perform encryption/decryption.
```
nlarew: nit: I think we could invert this sentence for a more active voice and to move the focus to the requirement it's calling out.

```rst
To perform client-side field level encryption and decryption, a client must have access to both the
master key and the data encryption keys.
```

source/use-cases/local-master-key.txt
```diff
@@ -1,89 +1,23 @@
...
 
+In the steps below, we set up a local master key saved to a file and load it into the MongoDB client settings in your development environment.
```
nlarew: "set up a local master key" and "save to a file" should be distinct

- we generate a local master key
- save it to a file
- and configure a MongoDB client to use the key

source/includes/steps-create-local-master-key.yaml
```diff
@@ -0,0 +1,68 @@
+title: Create the local master key
+ref: create-local-master-key
+level: 4
+content: |
+  Generate the local master key and save it to a file. You should only need to run this once:
```
nlarew: Let's be a little more imperative. Some users need the extra hand holding.

```rst
The following script generates a local master key and saves it to a file called ``master-key.txt``
in the same directory as the script. Copy the code and run it to generate the master key for this
application.
```

Also, do we need to specify that it should only run once? I would say we don't unless there is a significant downside to running it twice (e.g. crazy errors) or we see that users commonly do run it multiple times when they don't need to.

nlarew: Can we start this step with a brief overview of the anatomy of a master key? It looks like it has 96 bytes but a little more info would be nice (e.g. "it's 96 randomly generated bytes that have no relation to eachother")

```rst
Envelope encryption uses 96 byte master keys where each byte can be any random value. To perform
CSFLE, you must generate a master key.
```

source/includes/steps-create-local-master-key.yaml
```diff
@@ -0,0 +1,68 @@
...
+                 new SecureRandom().nextBytes(localMasterKey);
+
+                 String path = "master-key.txt";
+                 try (FileOutputStream stream = new FileOutputStream(path)) {
+                   stream.write(localMasterKey);
+                 } catch (IOException e)  {
+                   e.printStackTrace();
+                 }
+               }
+             }
```
nlarew: nit: should we add code comments? (I'm open to not including them if you think they're unnecessary or clutter the content.)

```java
import java.io.FileOutputStream;
import java.io.IOException;
import java.security.SecureRandom;

public class CreateMasterKeyFile {
  public static void main(final String[] args) {

    // Generate the master encryption key
    final byte[] localMasterKey = new byte[96];
    new SecureRandom().nextBytes(localMasterKey);

    // Save the key to a file on the local filesystem
    try (FileOutputStream stream = new FileOutputStream(path)) {
      stream.write(localMasterKey);
    } catch (IOException e)  {
      e.printStackTrace();
    }
  }
}
```

source/includes/steps-create-local-master-key.yaml
```diff
@@ -0,0 +1,68 @@
...
+
+---
+title: Load the local master key into the client settings
```
nlarew: I'm not sure that this is really "loading into the client settings" - we don't have any MongoClient code or JSON Schema in this. I like your wording in the step content: **Construct the KMS Settings Object**

source/includes/steps-create-local-master-key.yaml
```diff
@@ -0,0 +1,68 @@
...
+                 }
+               }
+             }
+
+---
+title: Load the local master key into the client settings 
+ref: load-local-master-key
+level: 4
+content: |
+  Next, in our MongoDB client application, add the following logic to read the master key from the file we created and construct the KMS settings:
```
nlarew: At this point I don't think we've introduced a client application. How do we plan to incorporate this object into the app? Once we know that, we may be able to be opinionated in this step without needing to specify exactly where to run the code.

Maybe this could also output to a file and we could read that file and serialize the KMS map out of it later when we need it?

nlarew: Alternatively, we could break this step out and include it after we've introduced the app

ccho-mongodb: To address all three of your comments in this section:

I thought it would be best to keep all the logic of creating and including the local master key in this section rather than returning to reference it later (perhaps in "[Section C. Configure the MongoDB Client](https://github.com/mongodb/docs-ecosystem/co
mmit/4f47fece63aeab1145d7ae5bb8ea6e73ea32d2b6#diff-c5c354e57498148e8ef37f38712ce945R67)") since it is possible to forget content from a couple sections prior, but perhaps that does not match what you were thinking.

Let me know if you have a preference for one of the following:

1) This is not a prerequisite for creating JSON Schema, so it could be swapped with that section. The introduction on keys could be moved into the Introduction section.
2) Relocate this step to Configure the MongoDB Client. The introduction on keys can stay in Create a Local Master Key.

source/includes/steps-create-local-master-key.yaml
ccho-mongodb: I have seen comments across the Java driver documentation of (subjectively) varying quality and need, so I thought it was acceptable and potentially helpful:
E.g. http://mongodb.github.io/mongo-java-driver/3.11/driver/tutorials/client-side-encryption/

Let me know what matches our usual convention (or how you'd like to establish it) as I do not have a strong opinion either way. Do/did we have a code documentation style guide that might help identify conventions like this?

source/use-cases/local-master-key.txt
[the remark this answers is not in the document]
ccho-mongodb: Should I defer the points that are not clear to the AWS article?
I was attempting to address your previous ask to include a sentence on why envelope encryption was better than basic by providing a summary and referring them to a source that provided additional detail.

source/use-cases/local-master-key.txt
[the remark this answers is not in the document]
ccho-mongodb: Questions answered in-person.

source/includes/steps-create-local-master-key.yaml
ccho-mongodb: Sounds good, will add!

source/use-cases/local-master-key.txt
ccho-mongodb: I agree. I think it needs a lead-in and also should probably only include info about the master key since that's all this step is concerned with.

source/use-cases/local-master-key.txt
ccho-mongodb: Will update, thanks.

source/use-cases/local-master-key.txt
ccho-mongodb: Good suggestion. I will need to defer on this until we figure out whether to include the configuration step.

source/use-cases/local-master-key.txt
ccho-mongodb: Removed the command line tool tabs

source/use-cases/local-master-key.txt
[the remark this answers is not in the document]
ccho-mongodb: Deferred explanation and comparison to AWS article

source/includes/steps-create-local-master-key.yaml
ccho-mongodb: Resolved in-person, moving settings to client setup section

source/includes/steps-create-local-master-key.yaml
ccho-mongodb: Resolved in person, will be creating a new section on setting up encryption client to create data keys.

source/use-cases/local-master-key.txt
```diff
@@ -1,23 +1,48 @@
...
 
-This strategy bolsters the security, minimizes encryption/decryption overhead, and allows the data encryption keys to be stored in the same replica set as the encrypted data. If a basic encryption strategy was used instead of envelope encryption, the computational and network load of encryption and decryption would need to be handled by the system that manages your keys. For more information on envelope encryption and key management concepts, read `this AWS article <https://docs.aws.amazon.com/kms/latest/developerguide/concepts.html#enveloping>`_.
+In CSFLE, the master key, used by the MongoDB driver to create and encrypt data keys, is securely stored in a Key Management Service (KMS). The data encryption keys, used to encrypt document fields, is stored in a user-specified collection alongside the encrypted data.
```
nlarew: It's preferred to use a KMS but it doesn't necessarily use one - our first is all about local master keys.

e.g.
```rst
MongoDB drivers use the master key, which is stored either locally or in
a KMS, to create and encrypt data keys. They then store the encrypted
data keys in a user-specified collection in the same database as the
encrypted data.
```

nlarew: nit: data encryption keys ... _is_ stored ->... _are_ stored

source/use-cases/local-master-key.txt
```diff
@@ -1,23 +1,48 @@
...
 
-This strategy bolsters the security, minimizes encryption/decryption overhead, and allows the data encryption keys to be stored in the same replica set as the encrypted data. If a basic encryption strategy was used instead of envelope encryption, the computational and network load of encryption and decryption would need to be handled by the system that manages your keys. For more information on envelope encryption and key management concepts, read `this AWS article <https://docs.aws.amazon.com/kms/latest/developerguide/concepts.html#enveloping>`_.
+In CSFLE, the master key, used by the MongoDB driver to create and encrypt data keys, is securely stored in a Key Management Service (KMS). The data encryption keys, used to encrypt document fields, is stored in a user-specified collection alongside the encrypted data. 
 
-The master key is stored separately in a Key Management System (KMS) in order to prevent the data encryption keys from being used directly to access the sensitive data.
+To encrypt and decrypt document fields, you must use a client using a MongoDB driver compatible with CSFLE, and it must be able to access both the master key and data encryption keys.
```
nlarew: nit: you must use ~a client using~ a MongoDB driver ~compatible~ _that supports_ ~with~ CSFLE

source/use-cases/local-master-key.txt
```diff
@@ -1,23 +1,48 @@
...
 
-The data encryption keys can safely be stored in a MongoDB collection alongside the encrypted sensitive data fields since they have been encrypted by the master key.
-
-In Client-Side Field Level Encryption, a client that can access both master key and data encryption keys is necessary to perform encryption/decryption.
-
-In the steps below, we set up a local master key saved to a file and load it into the MongoDB client settings in your development environment.
+In the code below, we generate the local master key -- a 96-byte, randomly generated value -- and save it to a file.
```
nlarew: nit: value -> string

source/use-cases/local-master-key.txt
```diff
@@ -0,0 +1,48 @@
+A. Create a Local Master Key
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+MongoDB Client-Side Field Level Encryption (CSFLE) uses an encryption strategy called envelope encryption in which keys used to encrypt/decrypt data (called **data encryption keys**) are encrypted with another key (called the **master key**). For more information on the features of envelope encryption and key management concepts, see `AWS Key Management Service Concepts <https://docs.aws.amazon.com/kms/latest/developerguide/concepts.html#enveloping`_.
```
nlarew: This link is missing a closing `>` - it should be:

```
`AWS Key Management Service Concepts <https://docs.aws.amazon.com/kms/latest/developerguide/concepts.html#enveloping>`_
```

source/use-cases/local-master-key.txt
nlarew: This seems like something that makes more sense under **Requirements** where we can more easily specify that this means 4.2+ drivers

source/use-cases/local-master-key.txt
ccho-mongodb: Oops, good catch. Should have seen this in the generated html.

source/use-cases/local-master-key.txt
ccho-mongodb: Sorry, that _are_ my bad. Do we have any licenses to Grammar.ly? After revising a few times, proofreading becomes tougher.

source/use-cases/local-master-key.txt
ccho-mongodb: I will remove it for now. I think it could go in the *Requirements* section, but the master and data encryption keys are not explained in any of the preceding sections (intro, problem, or solution). This requirement -- client must be able to access both master key and data encryption keys -- probably fits best in *Convert to a Remote Master Key* because that makes the requirement more difficult to satisfy because without network access to the KMS, the feature will not work.

source/use-cases/local-master-key.txt
ccho-mongodb: 👍

source/use-cases/local-master-key.txt
```diff
@@ -0,0 +1,46 @@
...
+In CSFLE, the master key, used by the MongoDB driver to create and encrypt data keys, is securely stored in a Key Management Service (KMS). The data encryption keys, used to encrypt document fields, are stored in a user-specified collection alongside the encrypted data. 
+
+In the code below, we generate the local master key -- a 96-byte, randomly generated string -- and save it to a file.
+
+.. admonition:: Local Master Keys Are Not Secure
+   :class: important
+   
+   To ensure that the master key cannot be compromised, do not use a local master key in a production environment. Instead, use a secure `Key Management System <https://en.wikipedia.org/wiki/Key_management#Key_management_system>`_ such as `AWS KMS <https://aws.amazon.com/kms/>`_.
+   
+   We demonstrate how to transition from a locally-hosted master key to a remote AWS KMS master key in a later step of this guide.
```
ccho-mongodb: Updated this sentence (original: "We demonstrate how to transition from a local key to AWS KMS in a later step of this guide.") to provide parallelism.

source/use-cases/local-master-key.txt
```diff
@@ -0,0 +1,46 @@
+A. Create a Local Master Key
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
```
terakilobyte: I think it would be good to provide some emphasis on "envelope encryption" since it's a term we're introducing.

source/use-cases/local-master-key.txt
```diff
@@ -0,0 +1,46 @@
...
+MongoDB Client-Side Field Level Encryption (CSFLE) uses an encryption strategy called envelope encryption in which keys used to encrypt/decrypt data (called **data encryption keys**) are encrypted with another key (called the **master key**). For more information on the features of envelope encryption and key management concepts, see `AWS Key Management Service Concepts <https://docs.aws.amazon.com/kms/latest/developerguide/concepts.html#enveloping>`_.
+
+In CSFLE, the master key, used by the MongoDB driver to create and encrypt data keys, is securely stored in a Key Management Service (KMS). The data encryption keys, used to encrypt document fields, are stored in a user-specified collection alongside the encrypted data.
```
terakilobyte: I think we can simplify how this reads by breaking it into two sentences.

In CSFLE, a MongoDB driver uses the master key to create and encrypt data data encryption keys. The master key is securely stored in a Key Management Service (KMS)...

Thoughts?

source/use-cases/local-master-key.txt
ccho-mongodb: Sounds good to me, will update.

source/use-cases/local-master-key.txt
ccho-mongodb: I am fine with whatever phrasing changes. From the pattern you introduced, should it be broken into four sentences to provide symmetry?
1) what the master key does
2) where the master key is stored
3) what the data encryption keys do
4) where the data encryption keys are stored

Or maybe some other combination of three sentences?
1) what the master key does
2) where the master key is stored
3) What the data encryption keys do and where they are stored

ccho-mongodb: Resolved in person

source/use-cases/local-master-key.txt
```diff
@@ -1,23 +1,48 @@
...
+The master key, used by the MongoDB driver to create and encrypt data keys, should be stored remotely in a Key Management Service (KMS). The data encryption keys, generated and used by the MongoDB driver to encrypt and decrypt document fields, are stored in a key vault collection in the same database as the encrypted data.
 
-The master key is stored separately in a Key Management System (KMS) in order to prevent the data encryption keys from being used directly to access the sensitive data.
+In the following example, we forgo using a KMS for the master key and instead generate a local one for testing purposes. This local master key is a 96-byte, randomly generated string saved to a file.
```
nlarew: nit: passive voice -> active voice

"... randomly generated string that we save to a file."

nlarew: nit: typically it's best to keep "the following" as close as possible to the actual thing you're referring to. In this instance, it's not totally clear whether "the following example" is the code block or the rest of the page. If we put "the following" into the sentence just before the code block and end that sentence with a colon it may be clearer.

Maybe we could split this into two sentences:
- to make it easier to develop & test we'll initially use a local master key
- the following code randomly generates a 96-byte master key and saves it to a local file

We already have the content of the second bullet point in the tabbed section, so maybe just break out the first bullet point here

source/use-cases/local-master-key.txt
```diff
@@ -1,23 +1,48 @@
...
 
-The data encryption keys can safely be stored in a MongoDB collection alongside the encrypted sensitive data fields since they have been encrypted by the master key.
-
-In Client-Side Field Level Encryption, a client that can access both master key and data encryption keys is necessary to perform encryption/decryption.
-
-In the steps below, we set up a local master key saved to a file and load it into the MongoDB client settings in your development environment.
+If you have access to AWS KMS and wish to skip this testing step, refer to :ref:`Convert to a Remote Master Key` instead.
```
nlarew: this seems related to the "Local Master Keys Are Not Secure" admonition - maybe we can combine them?

source/use-cases/local-master-key.txt
nlarew: Nice I like the change!

source/use-cases/local-master-key.txt
```diff
@@ -1,23 +1,48 @@
...
+
+.. tabs::
+
+  tabs:
+
+    - id: java-master-key-generator
+      name: "Java"
+      content: |
+
+        The following script generates a 96-byte local master key and saves it to a file called ``master-key.txt`` in the directory from which the script is executed.
```
nlarew: Currently we provide this same info twice in quick succession - let's pick one spot

source/use-cases/local-master-key.txt
ccho-mongodb: Will update to convey the point about making it easier to develop and test by using a local master key.

source/use-cases/local-master-key.txt
ccho-mongodb: Removed sentence in favor of admonition sentence.

source/use-cases/local-master-key.txt
ccho-mongodb: Removed the detail from the prior mention.

source/use-cases/local-master-key.txt
```diff
@@ -1,18 +1,16 @@
...
-In the following example, we forgo using a KMS for the master key and instead generate a local one for testing purposes. This local master key is a 96-byte, randomly generated string saved to a file.
-
-If you have access to AWS KMS and wish to skip this testing step, refer to :ref:`Convert to a Remote Master Key` instead.
+In this step, we generate a local master key to expedite setup of our development environment.
 
 .. admonition:: Local Master Keys Are Not Secure
    :class: important
 
-   To ensure that the master key cannot be compromised, do not use a local master key in a production environment. Instead, use a secure `Key Management System <https://en.wikipedia.org/wiki/Key_management#Key_management_system>`_ such as `AWS KMS <https://aws.amazon.com/kms/>`_.
+   Do not use a local master key in a production environment to ensure that it cannot be compromised. Instead, use a secure KMS such as `AWS KMS <https://aws.amazon.com/kms/>`_.
```
nlarew: I think the previous version of this was clearer (i.e. "To ensure \<foo\>, \<bar\>"). Here, "it" is ambiguous and could refer to either the master key _or_ the production env. (technically compromising the key also compromises the environment, but still)

ccho-mongodb: Ok, will change back. The reasoning for changing it was to provide emphasis to the "do not use a local master key" by placing it first, but probably not important, especially if the way I phrased it detracts from clarity.

ccho-mongodb:
Merged into https://github.com/mongodb/docs-ecosystem/pull/528

PR #526, opened 2019-08-15
FLE: intro sections

ccho-mongodb:
Intro, problem, and solution sections.

Staging: 
https://docs-mongodbcom-staging.corp.mongodb.com/ecosystem/docsworker/fle-guide-intro/use-cases/sensitive-data-encryption.html

### DIFF
--- a/source/use-cases/sensitive-data-encryption.txt
+++ b/source/use-cases/sensitive-data-encryption.txt
@@ -13,7 +13,7 @@ Sensitive Data Encryption
 Introduction
 ------------
 
-Many applications make use of sensitive data such as confidential personal details, payment information, or proprietary data. In some jurisdictions, this type of data is subject to governance, privacy, and security compliance mandates. Unauthorized access of sensitive data or a failure to comply to a mandate often results in significant reputation damage and financial penalties. Therefore, it is important to keep sensitive data secure.
+Many applications make use of sensitive data such as confidential personal details, payment information, or proprietary data. In some jurisdictions, this type of data is subject to governance, privacy, and security compliance mandates. Unauthorized access of sensitive data or a failure to comply with a mandate often results in significant reputation damage and financial penalties. Therefore, it is important to keep sensitive data secure.
 
 MongoDB offers several methods that protect your data from unauthorized access including:
  
@@ -38,8 +38,8 @@ In this scenario, we secure sensitive data on a Medical Care Management System w
 MedcoMD needs this system to satisfy the following use cases:
 
 - Doctors use the system to access Patients' medical records, insurance information, and add new vital sign measurements.
-- Receptionists use the system to verify the Patients' identity, using a combination of their contact information and last four digits of their Social Security Number (SSN).
-- Receptionists can view a Patient's insurance policy provider, but not the number.
+- Receptionists use the system to verify the Patients' identity, using a combination of their contact information and the last four digits of their Social Security Number (SSN).
+- Receptionists can view a Patient's insurance policy provider, but not their policy number.
 - Receptionists cannot access a Patient's medical records.
 
 MedcoMD is also concerned with disclosure of sensitive data through any of the following methods:
@@ -57,15 +57,15 @@ Solution
 
 The MedcoMD software engineers review the Medical Care Management System specification and research the proper solution for limiting access to the sensitive data.
 
-The first MongoDB security feature they evaluated was `Role-Based Access Control <https://docs.mongodb.com/manual/core/authorization/>`_ which allows administrators to grant and restrict collection-level permissions for users. With the appropriate role definition and assignment, this solution prevents accidental disclosure of data and access by a superuser. However, it does not prevent capture of data over an insecure network, access to data by reading the the server's memory, or access to on-disk data by reading database or backup files.
+The first MongoDB security feature they evaluated was `Role-Based Access Control <https://docs.mongodb.com/manual/core/authorization/>`_ which allows administrators to grant and restrict collection-level permissions for users. With the appropriate role definition and assignment, this solution prevents accidental disclosure of data and access. However, it does not prevent capture of data over an insecure network, direct access of data by a superuser, access to data by reading the server's memory, or access to on-disk data by reading the database or backup files.
 
 The next MongoDB security features they evaluated were `Encryption at Rest <https://docs.mongodb.com/manual/core/security-encryption-at-rest/>`_ which encrypts the database files on disk and `Transport Encryption using TLS/SSL <https://docs.mongodb.com/manual/core/security-transport-encryption/>`_ which encrypts data over the network. When applied together, these two features prevent access to on-disk database files as well as capture of the data on the network, respectively. When combined with Role-Based Access Control, these three security features offer near-comprehensive security coverage of the sensitive data, but lack a mechanism to prevent the data from being read from the server's memory.
 
 Finally, the MedcoMD developers discovered a feature that independently satisfies all the security criteria. Client-side Field Level Encryption allows the developers to specify the fields of a document that should be kept encrypted. Sensitive data is transparently encrypted/decrypted by the client and only communicated to and from the server in encrypted form. This mechanism keeps the specified data fields secure in encrypted form on both the server and the network. While all clients have access to the non-sensitive data fields, only appropriately-configured CSFLE clients are able to read and write the sensitive data fields.
 
-The Receptionists are be provided a client without CSFLE enabled and therefore cannot view the sensitive data fields nor accidentally leave them displayed on-screen in a public area. Doctors are provided a client with CSFLE enabled and therefore be able to access the sensitive data fields in the privacy of their office. 
+MedcoMD will provide Receptionists with a client that is not configured to access data encrypted with CSFLE. This will prevent them from viewing the sensitive fields and accidentally leaving them displayed on-screen in a public area. MedcoMD will provide Doctors with a client with CSFLE enabled which will allow them to access the sensitive data fields in the privacy of their own office.
 
-Equipped with CSFLE, MedcoMD can be confident that they can keep their sensitive data secure and compliant to data privacy regulations with MongoDB.
+Equipped with CSFLE, MedcoMD can keep their sensitive data secure and compliant to data privacy regulations with MongoDB.
 
 Procedure
 ---------
--- a/source/use-cases/sensitive-data-encryption.txt
+++ b/source/use-cases/sensitive-data-encryption.txt
@@ -13,18 +13,19 @@ Sensitive Data Encryption
 Introduction
 ------------
 
-If you store information in MongoDB such as confidential personal details, payment information, or proprietary data, you are probably interested in protecting your data from unauthorized access. In addition, the data you store may be subject to governance, privacy, and security compliance mandates. 
+Many applications make use of sensitive information such as confidential personal details, payment information, or proprietary data. In some jurisdictions, this data may be subject to governance, privacy, and security compliance mandates. Unauthorized access of data or a failure to comply to a mandate often results in significant reputation damage and financial penalties. Therefore, it is important to keep sensitive data secure.
 
-MongoDB offers several methods that reduce your risk of a data breach resulting from unauthorized access whether accidental or by a malicious act including:
-
+MongoDB offers several methods that protect your data from unauthorized access including:
+ 
 * `Role-based access control <https://docs.mongodb.com/manual/core/authorization/>`_
 * `Monitoring and logging <https://docs.mongodb.com/manual/core/auditing/>`_
 * `TLS/SSL network transport encryption <https://docs.mongodb.com/manual/core/security-transport-encryption/>`_
-* `Data encryption at rest <https://docs.mongodb.com/manual/core/security-encryption-at-rest/>`_
-* `Client-side field level encryption <https://docs.mongodb.com/manual/core/security-client-side-encryption/>`_
+* `Encryption at rest <https://docs.mongodb.com/manual/core/security-encryption-at-rest/>`_
 
-This document guides you through setup and implementation of Client-Side Field Level Encryption (CSFLE), a feature that encrypts document fields specified by a developer, in a example scenario to demonstrate how it can be used to secure your data.
+Another feature that enables 
 
+This document guides you through setup and implementation of `Client-Side Field Level Encryption (CSFLE) <https://docs.mongodb.com/manual/core/security-client-side-encryption/>`_, a MongoDB feature that encrypts document fields specified by a developer, 
+In the following scenario demonstrates how it can be used to secure your data.
 .. note::
 
   Client-side Field Level Encryption is available starting in MongoDB 4.2 Enterprise only
@@ -32,32 +33,45 @@ This document guides you through setup and implementation of Client-Side Field L
 Problem
 ~~~~~~~
 
-In this scenario, we secure sensitive data on a Medical Care Management System which stores patients' personal information, appointments, and vital sign measurement reports for a fictitious company, *MedcoMD*. While none of this data should ever be exposed to the public, if a malicious user gained access to data such as the patient's social security number (SSN) or their vital sign measurements, it would be disastrous and damaging to the company.
+In this scenario, we secure sensitive data on a Medical Care Management System which stores patients' personal information, appointments, and vital sign measurement reports for a fictitious company, *MedcoMD*. None of the patient data is public, and certain data such as their social security number (SSN) and vital sign measurements are particularly sensitive. It is important for the company and the patient that the data is kept private and secure.
 
-MedcoMD provides access to this system for two types of users:
+MedcoMD needs this system to satisfy the following use cases:
 
-1.  Receptionists, who use the system to schedule appointments for Patients, identified by a combination of contact information and their SSN.
-2.  Doctors, who use the system to access and enter Patients' vital sign measurements
+- Doctors use the system to access Patients' medical records and add new vital sign measurements.
+- Receptionists use the system to schedule appointments for Patients, who are identified by a combination of contact information and their SSN.
+- Receptionists cannot access nor update Patients' medical records.
 
-While MedcoMD trusts their employees never to retrieve sensitive data, there are several ways it could be exposed if left unencrypted including by leaving it exposed on a screen or print-out, by sniffing the data from network packets on an insecure network, or by obtaining unauthorized access to the database. If they store complete, encrypted documents on the server, every query would require decrypting the entire collection which is too computationally expensive, and therefore, slow.
+MedcoMD is also concerned with disclosure of private information through any of the following methods:
+
+- Printing the information or leaving it displayed on a screen where unauthorized people can view it
+- Communicating on an insecure network
+- Gaining access to the database on one of the permissioned users
 
 What can MedcoMD do to balance the functionality and access restrictions of their Medical Care Management System?
 
 Solution
 ~~~~~~~~
 
-MongoDB CSFLE allows the developer to specify the sensitive fields of a document to keep encrypted on the server. The client encrypts the specified, sensitive fields upon writes and transparently decrypts them upon access when necessary. The collections can be queried by any of the unencrypted fields without additional cost, and encrypted fields can be queried -- with restrictions documented in the following section, :ref:`DeterministicVsRandom` -- with minimal overhead.
+The MedcoMD software engineers review the Medical Care Management System specification and determine that Doctors and Receptionists require different levels of access to the data.
 
-If an attacker manages to gain access to the MongoDB database, unless they also gain access to an authorized client, they will only be able to see the sensitive fields in encrypted form.
+
+While role-based access control could separate the access, the granularity is only as fine as a collection
+
+
+In order to minimize the chance of disclosure of sensitive data, they decide encryption is the best option
+
+
+The MedcoMD developers decide this is an ideal opportunity to use CSFLE which allows them to select specific fields of a document that should be kept encrypted on the server. This feature will perform validation and encryption of the specified sensitive data fields, ensuring that they remain encrypted on the server. 
+
+be provided with a client that can access and
+ and the Receptionists will be provided
+
+
+The client encrypts the specified, sensitive fields upon writes and transparently decrypts them upon access when necessary. The collections can be queried by any of the unencrypted fields without additional cost, and encrypted fields can be queried -- with restrictions documented in the :ref:`DeterministicVsRandom` section -- with minimal overhead.
+
+If an unauthorized user manages to gain access to the MongoDB database, unless they also gain access to an authorized client, they will only be able to see the sensitive fields in encrypted form.
 
 MedcoMD can now be confident that MongoDB will keep their sensitive data secure and in compliance with data privacy regulations.
-
-.. _DeterministicVsRandom:
-
-Deterministic vs Random Encryption
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-
-TODO
 
 Procedure
 ---------
@@ -81,6 +95,11 @@ cubilia Curae; Nunc non interdum purus, ultricies laoreet tortor.
 
 B. Define a JSON Schema
 ~~~~~~~~~~~~~~~~~~~~~~~
+
+.. _DeterministicVsRandom:
+
+Deterministic vs Random Encryption
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 Proin non mi a felis luctus vulputate nec pharetra odio. Ut pretium
 scelerisque nulla, sed euismod nisi pellentesque sit amet. Proin mollis
--- a/source/use-cases/sensitive-data-encryption.txt
+++ b/source/use-cases/sensitive-data-encryption.txt
@@ -33,45 +33,37 @@ This guide provides steps for setup and implementation of CSFLE.
 Problem
 ~~~~~~~
 
-In this scenario, we secure sensitive data on a Medical Care Management System which stores patients' personal information, appointments, and vital sign measurement reports for a fictitious company, *MedcoMD*. None of the patient data is public, and certain data such as their social security number (SSN) and vital sign measurements are particularly sensitive. It is important for the company and the patient that the data is kept private and secure.
+In this scenario, we secure sensitive data on a Medical Care Management System which stores patients' personal information, insurance information, and medical records for a fictional company, *MedcoMD*. None of the patient data is public, and certain data such as their social security number (SSN), insurance policy number, and vital sign measurements are particularly sensitive and subject to privacy compliance. It is important for the company and the patient that the data is kept private and secure.
 
 MedcoMD needs this system to satisfy the following use cases:
 
-- Doctors use the system to access Patients' medical records and add new vital sign measurements.
-- Receptionists use the system to schedule appointments for Patients, who are identified by a combination of contact information and their SSN.
-- Receptionists cannot access nor update Patients' medical records.
+- Doctors use the system to access Patients' medical records, insurance information, and add new vital sign measurements.
+- Receptionists use the system to verify the Patients' identity, using a combination of their contact information and last four digits of their Social Security Number (SSN).
+- Receptionists can view a Patient's insurance policy provider, but not the number
+- Receptionists cannot access a Patient's medical records
 
 MedcoMD is also concerned with disclosure of private information through any of the following methods:
 
-- Printing the information or leaving it displayed on a screen where unauthorized people can view it
-- Communicating on an insecure network
-- Gaining access to the database on one of the permissioned users
+- Accidental disclosure of data on the Receptionist's publicly-viewable screen
+- Communication of the data over an insecure network
+- Direct access to the database by a superuser such as a database administrator
 
 What can MedcoMD do to balance the functionality and access restrictions of their Medical Care Management System?
 
 Solution
 ~~~~~~~~
 
-The MedcoMD software engineers review the Medical Care Management System specification and determine that Doctors and Receptionists require different levels of access to the data.
+The MedcoMD software engineers review the Medical Care Management System specification and determine they must decide the proper method of limiting access to the sensitive data.
 
+The first MongoDB security feature they evaluated was `Encryption at Rest<https://docs.mongodb.com/manual/core/security-encryption-at-rest/>`_ which encrypts data at the storage level. However, this solution did not address the potential issue of database administrators or Receptionists viewing or exposing the sensitive data.
 
-While role-based access control could separate the access, the granularity is only as fine as a collection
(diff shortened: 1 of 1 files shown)


source/use-cases/sensitive-data-encryption.txt
```diff
@@ -13,26 +13,51 @@ Sensitive Data Encryption
...
-Pellentesque non mi et est congue mattis pretium sit amet mi. Praesent
-nec tortor nec arcu molestie dapibus. Phasellus at nunc ut quam molestie
-finibus. Suspendisse finibus lectus eget nulla eleifend, vel consequat
-leo rhoncus. Phasellus lacinia tortor id aliquet condimentum. Morbi
-sodales est a leo rhoncus, non pretium neque dignissim.
+In this scenario, we secure sensitive data on a Medical Care Management System which stores patients' personal information, appointments, and vital sign measurement reports for a fictitious company, *MedcoMD*. While none of this data should ever be exposed to the public, if a malicious user gained access to data such as the patient's social security number (SSN) or their vital sign measurements, it would be disastrous and damaging to the company.
+
+MedcoMD provides access to this system for two types of users:
+
+1.  Receptionists, who use the system to schedule appointments for Patients, identified by a combination of contact information and their SSN.
```
ccho-mongodb: Was uncertain how to best emphasize user types (Receptionist, Doctor, Patient), so I capitalized them. Not necessary?

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -13,26 +13,51 @@ Sensitive Data Encryption
...
 ------------
 
-Cras sed rutrum nisi, vitae tristique risus. Aliquam erat volutpat.
-Donec auctor nisl eget ornare accumsan. Pellentesque lectus tortor,
-consectetur molestie pellentesque ac, blandit non libero.
+If you store information in MongoDB such as confidential personal details, payment information, or proprietary data, you are probably interested in protecting your data from unauthorized access. In addition, the data you store may be subject to governance, privacy, and security compliance mandates. 
+
+MongoDB offers several methods that reduce your risk of a data breach resulting from unauthorized access whether accidental or by a malicious act including:
+
+* `Role-based access control <https://docs.mongodb.com/manual/core/authorization/>`_
```
ccho-mongodb: Thought this was relevant to talking about security features, but open to removing it or relocating it if it obscures the main purpose of the guide.

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -13,26 +13,51 @@ Sensitive Data Encryption
...
+* `Monitoring and logging <https://docs.mongodb.com/manual/core/auditing/>`_
+* `TLS/SSL network transport encryption <https://docs.mongodb.com/manual/core/security-transport-encryption/>`_
+* `Data encryption at rest <https://docs.mongodb.com/manual/core/security-encryption-at-rest/>`_
+* `Client-side field level encryption <https://docs.mongodb.com/manual/core/security-client-side-encryption/>`_
+
+This document guides you through setup and implementation of Client-Side Field Level Encryption (CSFLE), a feature that encrypts document fields specified by a developer, in a example scenario to demonstrate how it can be used to secure your data.
+
+.. note::
+
+  Client-side Field Level Encryption is available starting in MongoDB 4.2 Enterprise only
```
ccho-mongodb: To reduce repetition, I could probably reference the Requirements section here instead: "Check the requirements section to see if CSFLE is available for your driver and server." Let me know if that makes more sense.

nlarew: I think this note is good as it is for now. Once we have the page built out a bit more fully we can revisit.

source/use-cases/sensitive-data-encryption.txt
nlarew: I like this approach!

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -13,26 +13,51 @@ Sensitive Data Encryption
...
 
 Problem
 ~~~~~~~
 
-Pellentesque non mi et est congue mattis pretium sit amet mi. Praesent
-nec tortor nec arcu molestie dapibus. Phasellus at nunc ut quam molestie
-finibus. Suspendisse finibus lectus eget nulla eleifend, vel consequat
-leo rhoncus. Phasellus lacinia tortor id aliquet condimentum. Morbi
-sodales est a leo rhoncus, non pretium neque dignissim.
+In this scenario, we secure sensitive data on a Medical Care Management System which stores patients' personal information, appointments, and vital sign measurement reports for a fictitious company, *MedcoMD*. While none of this data should ever be exposed to the public, if a malicious user gained access to data such as the patient's social security number (SSN) or their vital sign measurements, it would be disastrous and damaging to the company.
```
nlarew: **Problem** is the scenario definition, so I don't think we need to lead with "in this scenario". Instead, let's just state the situation:

- The medical company **MedcoMD** stores sensitive patient data such as...
- None of a patient's data should be public, but certain data is particularly private
- It's important for the company _and the patient_ that data is private and secured

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -13,26 +13,51 @@ Sensitive Data Encryption
...
+
+MedcoMD provides access to this system for two types of users:
+
+1.  Receptionists, who use the system to schedule appointments for Patients, identified by a combination of contact information and their SSN.
+2.  Doctors, who use the system to access and enter Patients' vital sign measurements
+
+While MedcoMD trusts their employees never to retrieve sensitive data, there are several ways it could be exposed if left unencrypted including by leaving it exposed on a screen or print-out, by sniffing the data from network packets on an insecure network, or by obtaining unauthorized access to the database. If they store complete, encrypted documents on the server, every query would require decrypting the entire collection which is too computationally expensive, and therefore, slow.
```
nlarew: nit: several ways _that_ it could...

nlarew: nit: we say "if left unencrypted" but we haven't quite introduced the case for encryption to the scenario yet. I wonder if it makes sense to say "these are some ways that data can be stolen" and then say "because of this they must have some data encryption but also it's too expensive to encrypt everything."

i.e. More or less what you already have here but phrased so that generic encryption follows as a solution to the problem.

nlarew: also is it "employees never retrieve sensitive data" or "employees only view sensitive data when it's appropriate"

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -13,26 +13,51 @@ Sensitive Data Encryption
...
+
+What can MedcoMD do to balance the functionality and access restrictions of their Medical Care Management System?
```
nlarew: Interesting use of a question to transition! I don't think we do anything rhetorical like this elsewhere but I'm down to see how it works out.

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -13,26 +13,51 @@ Sensitive Data Encryption
...
 
 Solution
 ~~~~~~~~
 
-Praesent sit amet elit pretium mi molestie porta a sit amet ipsum. Sed
-sollicitudin purus et mi blandit sodales.
+MongoDB CSFLE allows the developer to specify the sensitive fields of a document to keep encrypted on the server. The client encrypts the specified, sensitive fields upon writes and transparently decrypts them upon access when necessary. The collections can be queried by any of the unencrypted fields without additional cost, and encrypted fields can be queried -- with restrictions documented in the following section, :ref:`DeterministicVsRandom` -- with minimal overhead.
```
nlarew: We go from "here's the problem" straight to "MongoDB can do FLE". Instead, can we start this section by describing FLE as a sort of hybrid between totally unencrypted and fully encrypted that takes care of the issues we just talked about?

Then it makes sense to talk about how MongoDB specifically implements that solution.

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -13,26 +13,51 @@ Sensitive Data Encryption
...
+
+If an attacker manages to gain access to the MongoDB database, unless they also gain access to an authorized client, they will only be able to see the sensitive fields in encrypted form.
+
+MedcoMD can now be confident that MongoDB will keep their sensitive data secure and in compliance with data privacy regulations.
+
+.. _DeterministicVsRandom:
+
+Deterministic vs Random Encryption
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
nlarew: I think we can push this explanation down into the JSON schema step or wherever we decide to go over the collection schema (pre-JSON Schema encryption)

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -13,26 +13,51 @@ Sensitive Data Encryption
...
-nec tortor nec arcu molestie dapibus. Phasellus at nunc ut quam molestie
-finibus. Suspendisse finibus lectus eget nulla eleifend, vel consequat
-leo rhoncus. Phasellus lacinia tortor id aliquet condimentum. Morbi
-sodales est a leo rhoncus, non pretium neque dignissim.
+In this scenario, we secure sensitive data on a Medical Care Management System which stores patients' personal information, appointments, and vital sign measurement reports for a fictitious company, *MedcoMD*. While none of this data should ever be exposed to the public, if a malicious user gained access to data such as the patient's social security number (SSN) or their vital sign measurements, it would be disastrous and damaging to the company.
+
+MedcoMD provides access to this system for two types of users:
+
+1.  Receptionists, who use the system to schedule appointments for Patients, identified by a combination of contact information and their SSN.
+2.  Doctors, who use the system to access and enter Patients' vital sign measurements
```
nlarew: to access the Patient's medical records and enter new vital sign measurements

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -13,26 +13,51 @@ Sensitive Data Encryption
...
+What can MedcoMD do to balance the functionality and access restrictions of their Medical Care Management System?
 
 Solution
 ~~~~~~~~
 
-Praesent sit amet elit pretium mi molestie porta a sit amet ipsum. Sed
-sollicitudin purus et mi blandit sodales.
+MongoDB CSFLE allows the developer to specify the sensitive fields of a document to keep encrypted on the server. The client encrypts the specified, sensitive fields upon writes and transparently decrypts them upon access when necessary. The collections can be queried by any of the unencrypted fields without additional cost, and encrypted fields can be queried -- with restrictions documented in the following section, :ref:`DeterministicVsRandom` -- with minimal overhead.
+
+If an attacker manages to gain access to the MongoDB database, unless they also gain access to an authorized client, they will only be able to see the sensitive fields in encrypted form.
```
nlarew: nit: Doesn't necessarily have to be an "attacker". e.g. it could just be the janitor seeing data on screen non-maliciously - not an "attack" but still very uncool.

Maybe "unauthorized user"?

source/use-cases/sensitive-data-encryption.txt
nlarew: Are we essentially saying that "receptionists shouldn't see sensitive data but doctors need to be able to read and write that data"? I think being explicit about the data access requirements will help clarify the use case, i.e. different clients have different access privileges.

We should also be clear (somewhere) that this won't be accomplished by role-based access permissions (i.e. doctor & receptionist user roles) but instead with permissioned clients (i.e. a reception client and doctor client)

source/use-cases/sensitive-data-encryption.txt
nlarew: It is definitely relevant - let's see if we can frame it in a way that shows a weakness or limitation of RBAC that is covered by CSFLE.

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -13,26 +13,51 @@ Sensitive Data Encryption
...
-consectetur molestie pellentesque ac, blandit non libero.
+If you store information in MongoDB such as confidential personal details, payment information, or proprietary data, you are probably interested in protecting your data from unauthorized access. In addition, the data you store may be subject to governance, privacy, and security compliance mandates. 
+
+MongoDB offers several methods that reduce your risk of a data breach resulting from unauthorized access whether accidental or by a malicious act including:
+
+* `Role-based access control <https://docs.mongodb.com/manual/core/authorization/>`_
+* `Monitoring and logging <https://docs.mongodb.com/manual/core/auditing/>`_
+* `TLS/SSL network transport encryption <https://docs.mongodb.com/manual/core/security-transport-encryption/>`_
+* `Data encryption at rest <https://docs.mongodb.com/manual/core/security-encryption-at-rest/>`_
+* `Client-side field level encryption <https://docs.mongodb.com/manual/core/security-client-side-encryption/>`_
```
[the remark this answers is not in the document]
terakilobyte: We shouldn't frame the usage of CSFLE as an alternative to these. Rather, it is a complementary.

source/use-cases/sensitive-data-encryption.txt
terakilobyte: One nit, I'd say "malicious actor"

source/use-cases/sensitive-data-encryption.txt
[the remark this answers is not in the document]
terakilobyte: Also to add, I think we need to add encryption as rest to this list, which will confuse a point further in the content which I'll comment on, full document encryption.

source/use-cases/sensitive-data-encryption.txt
terakilobyte: We need to be careful not to confuse full document encryption with encryption at rest. Maybe something like

> While encryption at rest protects against a malicious actor from gaining data on disk, if they were to compromise valid credentials they would have access like any other user. Full document encryption would prove too slow as there is a lot of information that doesn't need to be encrypted, and every query would ...

source/use-cases/sensitive-data-encryption.txt
ccho-mongodb: @terakilobyte 
Re: line 18, after reading the sentence a few times, I think it is not as crucial to mention the intentions of the data access. What do you think?
Here's a sample replacement sentence:
```
MongoDB offers several methods that protect your data from unauthorized access including:
```

source/use-cases/sensitive-data-encryption.txt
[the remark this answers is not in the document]
ccho-mongodb: @terakilobyte Does line 23 "Data encryption at rest" cover the item you mentioned should be added?
@nlarew I think you're identifying that the list, due to its size, is too prominent and lacks explanation. I could combine line 18 with your first bullet point "MongoDB supports several features that reduce risk, like ..." and then explain that this guide will focus on CSFLE. Or perhaps CSFLE should come first and then mention the "other methods" that contribute to security.

source/use-cases/sensitive-data-encryption.txt
ccho-mongodb: Thanks. I found some grammatical ambiguity in the sentence that I will fix.

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -13,26 +13,51 @@ Sensitive Data Encryption
 Introduction
 ------------
 
-Cras sed rutrum nisi, vitae tristique risus. Aliquam erat volutpat.
-Donec auctor nisl eget ornare accumsan. Pellentesque lectus tortor,
-consectetur molestie pellentesque ac, blandit non libero.
+If you store information in MongoDB such as confidential personal details, payment information, or proprietary data, you are probably interested in protecting your data from unauthorized access. In addition, the data you store may be subject to governance, privacy, and security compliance mandates.
```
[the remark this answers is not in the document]
ccho-mongodb: Sure, sounds good.
I initially had a generic introduction, but then changed it after reading the corresponding section from https://docs.mongodb.com/ecosystem/use-cases/metadata-and-asset-management/: 
"You are designing a content management system (CMS) and you want to use MongoDB to store the content of your sites."

Will work on rewording.

source/use-cases/sensitive-data-encryption.txt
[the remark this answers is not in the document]
ccho-mongodb: I still want to emphasize that MongoDB offers these important security features.
Probably a personal issue, but I am not creative enough to form a sentence that smoothly conveys "MongoDB offers several security features to help keep your data secure such as ... which you can learn about further through the linked documents. This article provides an example implementation to show you how to implement CSFLE." 
Any suggestions on how to write this?

source/use-cases/sensitive-data-encryption.txt
ccho-mongodb: Adds clarity, thanks.

source/use-cases/sensitive-data-encryption.txt
ccho-mongodb: Makes sense, will update.

source/use-cases/sensitive-data-encryption.txt
ccho-mongodb: Agree, will remove.

source/use-cases/sensitive-data-encryption.txt
ccho-mongodb: You're right, it is abrupt. I will update this to demonstrate what is meant by specifying sensitive fields and show other fields that are kept in plaintext.

source/use-cases/sensitive-data-encryption.txt
ccho-mongodb: @terakilobyte I'm concerned that talking about encryption at rest will require explanation and shift focus away from the Problem section. Could we chat in-person about this so that I can understand what you're looking for?

source/use-cases/sensitive-data-encryption.txt
ccho-mongodb: Redoing this section to state the following in order to match the final sentence "What can MedcoMD do...":
-What are the use cases the system must satisfy?
-What are the security concerns?

Then, will try to clarify them more in the solution.

ccho-mongodb: I wonder if placing equal weight on the two problems this solves -- how to enforce RBAC with FLE and minimizing risk of disclosing sensitive data -- emphasizes the RBAC parts too much as it was not listed in any goals in the [Goals/Non-Goals Scope doc](https://docs.google.com/document/d/1Tl2ftwyUu9g6piJF0yIRdG2eeSc6YL4DV2LqZp4MgxQ/edit#heading=h.ha97b73mwsrj).

source/use-cases/sensitive-data-encryption.txt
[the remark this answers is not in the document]
ccho-mongodb: Discussed in person and decided on:
-keep the list, but leave CSFLE off of it
-introduce it immediately following

Hopefully, my revision will make sense.

source/use-cases/sensitive-data-encryption.txt
ccho-mongodb: Discussed in-person:
Will leave out explanation and describe use cases and risks in list format

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -13,26 +13,65 @@ Sensitive Data Encryption
...
+MongoDB offers several methods that protect your data from unauthorized access including:
+ 
+* `Role-based access control <https://docs.mongodb.com/manual/core/authorization/>`_
+* `Monitoring and logging <https://docs.mongodb.com/manual/core/auditing/>`_
+* `TLS/SSL network transport encryption <https://docs.mongodb.com/manual/core/security-transport-encryption/>`_
+* `Encryption at rest <https://docs.mongodb.com/manual/core/security-encryption-at-rest/>`_
+
+Another MongoDB feature that prevents unauthorized access of data is `Client-Side Field Level Encryption (CSFLE) <https://docs.mongodb.com/manual/core/security-client-side-encryption/>`_. This feature allows a developer to selectively encrypt individual fields of a document on the client-side before it is sent to the server. This keeps the encrypted data private from the providers hosting the database as well as any user that has direct access to the database.
+
+This guide provides steps for setup and implementation of CSFLE.
```
ccho-mongodb: Should this include a description of the scenario? E.g.: 
"This guide provides steps for setup and implementation of CSFLE for a real-world scenario."

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -13,26 +13,65 @@ Sensitive Data Encryption
...
+While role-based access control could separate the access, the granularity is only as fine as a collection
+
+
+In order to minimize the chance of disclosure of sensitive data, they decide encryption is the best option
+
+
+The MedcoMD developers decide this is an ideal opportunity to use CSFLE which allows them to select specific fields of a document that should be kept encrypted on the server. This feature will perform validation and encryption of the specified sensitive data fields, ensuring that they remain encrypted on the server. 
+
+be provided with a client that can access and
+ and the Receptionists will be provided
```
nlarew: Looks like this spot is where we want to describe "doctor client vs receptionist client"?

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -13,26 +13,65 @@ Sensitive Data Encryption
...
+
+
+The client encrypts the specified, sensitive fields upon writes and transparently decrypts them upon access when necessary. The collections can be queried by any of the unencrypted fields without additional cost, and encrypted fields can be queried -- with restrictions documented in the :ref:`DeterministicVsRandom` section -- with minimal overhead.
```
nlarew: nit:the client encrypts sensitive fields  _before_ write operations

nlarew: nit: In general I've found it to be good practice to explicitly specify the ref text so that we can read it inline and so that changing the ref'd heading doesn't break the grammar of the sentence that refers to it.

``` :ref:`Deterministic vs Random Encryption <DeterministicVsRandom>` ```

This does, of course, add additional overhead if we do ever decide to rename it, but forcing us to slow down and review the impact of a change isn't necessarily a bad thing.

File: source/use-cases/sensitive-data-encryption.txt [hunk 16]
```diff
@@ -13,26 +13,65 @@ Sensitive Data Encryption
...
-Praesent sit amet elit pretium mi molestie porta a sit amet ipsum. Sed
-sollicitudin purus et mi blandit sodales.
+The MedcoMD software engineers review the Medical Care Management System specification and determine that Doctors and Receptionists require different levels of access to the data.
+
 
-- Deterministic vs Non-Deterministic
+While role-based access control could separate the access, the granularity is only as fine as a collection
+
+
+In order to minimize the chance of disclosure of sensitive data, they decide encryption is the best option
```
nlarew: nit: These first three sentences could be one single paragraph.

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -13,26 +13,65 @@ Sensitive Data Encryption
...
 Solution
 ~~~~~~~~
 
-Praesent sit amet elit pretium mi molestie porta a sit amet ipsum. Sed
-sollicitudin purus et mi blandit sodales.
+The MedcoMD software engineers review the Medical Care Management System specification and determine that Doctors and Receptionists require different levels of access to the data.
+
 
-- Deterministic vs Non-Deterministic
+While role-based access control could separate the access, the granularity is only as fine as a collection
```
nlarew: why is collection-level granularity insufficient?

nlarew: we might also mention that RBAC does not prevent authorized roles from seeing/searching all data in the collection - it's an all or nothing thing, whereas we want some parts of each document to be encrypted/unsearchable and others to not be.

source/use-cases/sensitive-data-encryption.txt
nlarew: This is a short sentence and adding a few more words doesn't really hurt anything. Your suggestion gives good context, so I'm a fan of adding it.

source/use-cases/sensitive-data-encryption.txt
ccho-mongodb: Whoops, incomplete sentence/thought.

source/use-cases/sensitive-data-encryption.txt
ccho-mongodb: Personal notes. Replaced with other arguments for CSFLE

source/use-cases/sensitive-data-encryption.txt
ccho-mongodb: Removed the part about deterministic vs random since that will be explained in a later section. Also, we made the decision to use an unencrypted client for receptionists and encrypted one for doctors.

source/use-cases/sensitive-data-encryption.txt
ccho-mongodb: Updated to explain this.

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -24,54 +24,46 @@ MongoDB offers several methods that protect your data from unauthorized access i
...
 
 What can MedcoMD do to balance the functionality and access restrictions of their Medical Care Management System?
 
 Solution
 ~~~~~~~~
 
-The MedcoMD software engineers review the Medical Care Management System specification and determine that Doctors and Receptionists require different levels of access to the data.
+The MedcoMD software engineers review the Medical Care Management System specification and determine they must decide the proper method of limiting access to the sensitive data.
 
+The first MongoDB security feature they evaluated was `Encryption at Rest <https://docs.mongodb.com/manual/core/security-encryption-at-rest/>`_ which encrypts data at the storage level. However, this solution did not address the potential issue of database administrators or Receptionists viewing or exposing the sensitive data.
```
nlarew: Should we add a blurb to the second sentence that tells what it _does_ accomplish? Or do you think we shouldn't focus too much on it?

e.g. "This would prevent \<foo\>, but it does not address..."

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -24,54 +24,46 @@ MongoDB offers several methods that protect your data from unauthorized access i
...
 
-While role-based access control could separate the access, the granularity is only as fine as a collection
+The next MongoDB security feature they evaluated was `Role-Based Access Control <https://docs.mongodb.com/manual/core/authorization/>`_ which allows them to define distinct permissions for users. However, this solution did not allow them to differentiate access to specific fields within a document nor keep the data safe when transmitted over an insecure network.  
 
+Finally, the MedcoMD developers discovered Client-side Field Level Encryption which allows them to specify fields of a document that should be stored in encrypted form on the server. Before sending the data to the server, the client encrypts the specified, sensitive fields. After retrieving data from the server, the client decrypts any encrypted fields transparently and automatically. This prevents any unencrypted sensitive data from being passed over the network as well as access to the data by a user with direct access to the database.
```
nlarew: nit: "specify _individual_ fields"

nlarew: The last sentence is a little long
```rst
This prevents sensitive data from being passed over the network unencrypted. It also prevents any user
that accesses the database directly from viewing the data.
```

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -24,54 +24,46 @@ MongoDB offers several methods that protect your data from unauthorized access i
...
 
-In order to minimize the chance of disclosure of sensitive data, they decide encryption is the best option
+The Doctors will be provided with a client with CSFLE enabled and will therefore be able to access the sensitive data. The Receptionists will be provided a client without CSFLE enabled and therefore will never be able to access the sensitive data nor accidentally leave it displayed on-screen in a public area.
```
nlarew: nit: with a client _that has_ CSFLE enabled

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -39,31 +39,33 @@ MedcoMD needs this system to satisfy the following use cases:
 
 - Doctors use the system to access Patients' medical records, insurance information, and add new vital sign measurements.
 - Receptionists use the system to verify the Patients' identity, using a combination of their contact information and last four digits of their Social Security Number (SSN).
-- Receptionists can view a Patient's insurance policy provider, but not the number
-- Receptionists cannot access a Patient's medical records
+- Receptionists can view a Patient's insurance policy provider, but not the number.
```
nlarew: nit: "their policy number"

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -13,26 +13,59 @@ Sensitive Data Encryption
...
+What can MedcoMD do to balance the functionality and access restrictions of their Medical Care Management System?
 
 Solution
 ~~~~~~~~
 
-Praesent sit amet elit pretium mi molestie porta a sit amet ipsum. Sed
-sollicitudin purus et mi blandit sodales.
+The MedcoMD software engineers review the Medical Care Management System specification and research the proper solution for limiting access to the sensitive data.
+
+The first MongoDB security feature they evaluated was `Role-Based Access Control <https://docs.mongodb.com/manual/core/authorization/>`_ which allows administrators to grant and restrict collection-level permissions for users. With the appropriate role definition and assignment, this solution prevents accidental disclosure of data and access by a superuser. However, it does not prevent capture of data over an insecure network, access to data by reading the the server's memory, or access to on-disk data by reading database or backup files.
```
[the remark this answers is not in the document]
nlarew: Not sure if RBAC actually prevents superuser attacks given that a "superuser" is just someone with a role that gives them full access

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -39,31 +39,33 @@ MedcoMD needs this system to satisfy the following use cases:
...
 
-The Doctors will be provided with a client with CSFLE enabled and will therefore be able to access the sensitive data. The Receptionists will be provided a client without CSFLE enabled and therefore will never be able to access the sensitive data nor accidentally leave it displayed on-screen in a public area.
+The Receptionists are be provided a client without CSFLE enabled and therefore cannot view the sensitive data fields nor accidentally leave them displayed on-screen in a public area. Doctors are provided a client with CSFLE enabled and therefore be able to access the sensitive data fields in the privacy of their office.
```
nlarew: nit: Let's add a transition sentence that goes from "CSFLE satisfies the requirements" to "now that they have CSFLE the developers are satisfied and have decided on a solution"

nlarew: Let's reframe using active voice:

MedcoMD will provide Receptionists with a client that is not configured to access data encrypted with CSFLE...

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -13,26 +13,59 @@ Sensitive Data Encryption
...
+
+What can MedcoMD do to balance the functionality and access restrictions of their Medical Care Management System?
 
 Solution
 ~~~~~~~~
 
-Praesent sit amet elit pretium mi molestie porta a sit amet ipsum. Sed
-sollicitudin purus et mi blandit sodales.
+The MedcoMD software engineers review the Medical Care Management System specification and research the proper solution for limiting access to the sensitive data.
+
```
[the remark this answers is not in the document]
nlarew: "access to sensitive data" === generic
"access to _the_ sensitive data" === specific

We've already introduced the fields MedcoMD is concerned about in **Problem,** so we can refer to it specifically.

PR #527, opened 2019-08-16
Docsp 6483 jsonschema withscaffold

mongomoe:
Jira Ticket: https://jira.mongodb.org/browse/DOCSP-6483?filter=-1

Staged version: https://docs-mongodbcom-staging.corp.mongodb.com/ecosystem/docsworker/DOCSP-6483-JSONSCHEMA-withscaffold/use-cases/sensitive-data-encryption.html#b-define-a-json-schema

### DIFF
--- a/source/use-cases/sensitive-data-encryption.txt
+++ b/source/use-cases/sensitive-data-encryption.txt
@@ -1,0 +1,109 @@
+=========================
+Sensitive Data Encryption
+=========================
+
+.. default-domain:: mongodb
+
+.. contents:: On this page
+   :local:
+   :backlinks: none
+   :depth: 2
+   :class: singlecol
+
+Introduction
+------------
+
+Cras sed rutrum nisi, vitae tristique risus. Aliquam erat volutpat.
+Donec auctor nisl eget ornare accumsan. Pellentesque lectus tortor,
+consectetur molestie pellentesque ac, blandit non libero.
+
+Problem
+~~~~~~~
+
+Pellentesque non mi et est congue mattis pretium sit amet mi. Praesent
+nec tortor nec arcu molestie dapibus. Phasellus at nunc ut quam molestie
+finibus. Suspendisse finibus lectus eget nulla eleifend, vel consequat
+leo rhoncus. Phasellus lacinia tortor id aliquet condimentum. Morbi
+sodales est a leo rhoncus, non pretium neque dignissim.
+
+Solution
+~~~~~~~~
+
+Praesent sit amet elit pretium mi molestie porta a sit amet ipsum. Sed
+sollicitudin purus et mi blandit sodales.
+
+- Deterministic vs Non-Deterministic
+
+Procedure
+---------
+
+Requirements
+~~~~~~~~~~~~
+
+- Point to setup and install directions
+- Nullam imperdiet lorem vitae vulputate lacinia.
+- Donec eget velit tincidunt, gravida diam ac, efficitur lacus.
+
+A. Create a Local Master Key
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Aenean eu consequat lorem. Ut posuere est sed sodales pharetra. Cras
+volutpat, massa laoreet varius dictum, leo odio porttitor ante, nec
+auctor tortor orci et mi. Maecenas tempor, lacus vehicula molestie
+pulvinar, ante eros faucibus odio, sed consequat quam tellus vel arcu.
+Vestibulum ante ipsum primis in faucibus orci luctus et ultrices posuere
+cubilia Curae; Nunc non interdum purus, ultricies laoreet tortor.
+
+B. Define a JSON Schema
+~~~~~~~~~~~~~~~~~~~~~~~
+
+Proin non mi a felis luctus vulputate nec pharetra odio. Ut pretium
+scelerisque nulla, sed euismod nisi pellentesque sit amet. Proin mollis
+mauris eu libero ullamcorper rutrum. Pellentesque vitae enim sed magna
+accumsan maximus et quis mi. Ut vel laoreet turpis. Vivamus vitae odio
+sed arcu finibus pretium vel maximus nisi. Nullam gravida consequat
+porta.
+
+C. Configure the MongoDB Client
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Donec sem felis, iaculis id pharetra sed, sollicitudin ut nisl.
+Suspendisse commodo sagittis ante in ullamcorper. Etiam viverra
+pellentesque suscipit. Pellentesque interdum egestas libero, ut varius
+velit congue vel.
+
+D. Perform Encrypted Read/Write Operations
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Etiam hendrerit posuere justo, ac hendrerit lorem pellentesque sed.
+Fusce ullamcorper neque eu libero aliquam, sed placerat sapien cursus.
+Vestibulum maximus magna sed velit feugiat tempor.
+
+Ut accumsan pellentesque metus, at porta felis dapibus vulputate. Aenean
+malesuada interdum libero in feugiat. Fusce pharetra massa eget leo
+varius, et gravida massa tempus.
+
+Ut mattis eros elit, sed ultrices purus lacinia sed. Curabitur tincidunt
+tellus eget neque porta dictum ac vel orci. Vestibulum ante ipsum primis
+in faucibus orci luctus et ultrices posuere cubilia Curae.
+
+Praesent feugiat arcu non risus elementum eleifend. Nullam sed iaculis
+erat.
+
+E. Convert to a Remote Master Key
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Fusce porta lectus et enim vehicula malesuada. In porttitor ipsum eu
+dictum congue. Fusce pharetra sagittis lacus sed posuere. Donec porta
+posuere gravida. Quisque nisi massa, condimentum vitae ligula eget,
+fringilla malesuada quam. Mauris in semper nulla, non blandit tortor.
+Mauris consequat eget urna at eleifend. Suspendisse potenti. Fusce at
+justo malesuada, finibus felis non, dignissim nisl. Ut blandit nulla sit
+amet massa euismod vehicula.
+
+Summary
+-------
+
+- What we wanted to do
+- What we did
+- More reading
--- a/source/use-cases/sensitive-data-encryption.txt
+++ b/source/use-cases/sensitive-data-encryption.txt
@@ -57,12 +57,53 @@ cubilia Curae; Nunc non interdum purus, ultricies laoreet tortor.
 B. Define a JSON Schema
 ~~~~~~~~~~~~~~~~~~~~~~~
 
-Proin non mi a felis luctus vulputate nec pharetra odio. Ut pretium
-scelerisque nulla, sed euismod nisi pellentesque sit amet. Proin mollis
-mauris eu libero ullamcorper rutrum. Pellentesque vitae enim sed magna
-accumsan maximus et quis mi. Ut vel laoreet turpis. Vivamus vitae odio
-sed arcu finibus pretium vel maximus nisi. Nullam gravida consequat
-porta.
+MongoDB uses JSON Schema to to validate data which is written to a collection.
+Within the context of Field Level Encryption, we use JSON Schema to specify several
+configuration details. For instance, we can specify the form of encryption we want to use 
+- deterministic or random. We can also specify the BSON Type of the field we are encrypting. 
+In addition, we can specify the key we want to use. This is particularly important, because
+it means that we can use different keys for different fields.
+
+In the following example, we will go over creation of a schema for a collection which represents patient data
+in a medical office. Some fields use deterministic encryption, and others use random (probabilistic encryption). 
+
+The structure for our schema will be as follows:
+
+ .. code::
+
+    fullName: String
+    address: Encrypted String using random encryption
+    ssn: Encrypted Int using deterministic encryption
+    last4SSN: Int
+    appointments: Object
+    medicalRecords: Encrypted array of documents using random encryption
+
+
+* Address randomly encrypted because it is more secure for strings to be randomly encrypted to prevent hackers from decrypting by building a cipher dictionary
+* SSN is deterministically encrypted because deterministic encryption provides filtering ability 
+* Medical Records are randomly encrypted because entire array's are only encryptable with random encryption
+
+
+In this example, the same key is used for the each encrypted field in the document. The following code illustrates how to create the above schema and 
+
+.. tabs::
+
+   tabs:
+
+     - id: java-master-key-generator
+      name: "Java"
+      content: |
+
+         The following code creates a collection using the schema above and inserts it into a database using MongoDB.
+
+         .. code-block:: java
+
+            import java.io.FileOutputStream;
+            import java.io.IOException;
+            import java.security.SecureRandom;
+
+
+
 
 C. Configure the MongoDB Client
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
--- a/source/use-cases/sensitive-data-encryption.txt
+++ b/source/use-cases/sensitive-data-encryption.txt
@@ -69,40 +69,202 @@ in a medical office. Some fields use deterministic encryption, and others use ra
 
 The structure for our schema will be as follows:
 
- .. code::
+.. code::
 
     fullName: String
     address: Encrypted String using random encryption
     ssn: Encrypted Int using deterministic encryption
     last4SSN: Int
-    appointments: Object
+    patientInfo: Object
+    patientInfo.phone Encrypted Int using deterministic encryption
+    patientInfo.provider String
     medicalRecords: Encrypted array of documents using random encryption
 
-* Address randomly encrypted because it is more secure for strings to be randomly encrypted to prevent hackers from decrypting by building a cipher-dictionary
-* SSN is deterministically encrypted because deterministic encryption provides filtering ability 
-* Medical Records are randomly encrypted because entire array's are only encryptable with random encryption
+Explanation of encryption choices:
+``````````````````````````````````
+* address is randomly encrypted because it is more secure for strings to be randomly encrypted to prevent hackers from decrypting by building a cipher-dictionary
+* ssn is deterministically encrypted because deterministic encryption provides filtering ability 
+* patientInfo.phone is deterministically encrypted for the same reason as ssn, and also it provides an example of encryption with a nested document
+* medicalRecords are randomly encrypted because entire array's are only encryptable with random encryption
 
-testabc
 
 .. tabs::
 
    tabs:
 
-     - id: java-master-key-generator
-      name: "Java"
-      content: |
+     - id: java-jsonschema-generation
+       name: "Java"
+       content: |
 
          The following code creates a collection using the schema above and inserts it into a database using MongoDB.
+         For the "Define a JSON Schema" section we've abstracted the createSecureClient, to focus on creation of a schema.
+         But that function uses getJSONSchema and passes it a dataKey id in order to create a client which uses the JSONSchema
+         we want to use. 
 
          .. code-block:: java
-            import java.util.function.Consumer;
+
+            import com.mongodb.*;
+            import com.mongodb.client.MongoClient;
+            import com.mongodb.client.MongoClients;
+            import com.mongodb.client.MongoCollection;
+            import com.mongodb.client.model.vault.DataKeyOptions;
+            import com.mongodb.client.vault.ClientEncryptions;
+
+            import org.bson.BsonDocument;
+
+            import java.util.Base64;
+            import java.util.Map;
+
             import org.json.*;
 
-            public class FleApplication {
+            public class FieldLevelEncryptionDemo {
+                static String dbName = "fieldLevelEncryptionDemo001";
+                static String collectionName = "patients";
+                public static void main(String[] args) {
+                    MongoCollection secureCollection = createSecureClient().getDatabase(dbName).getCollection(collectionName);
+                    ...
+                }
+                public static JSONObject getEncryptedField(String keyId, String chosenBsonType, Boolean isDeterministic){
+                    JSONObject binaryForKey = new JSONObject();
+                    binaryForKey.put("base64", keyId);
+                    binaryForKey.put("subType", "04");
+                    JSONObject binWrapper = new JSONObject();
+                    binWrapper.put("$binary", binaryForKey);
+                    
+                    JSONObject encryptedFieldWrap = new JSONObject();
+                    JSONObject encrypt = new JSONObject();
+
+                    encrypt.put("bsonType", chosenBsonType);
+                    if(isDeterministic){
+                        encrypt.put("algorithm", "AEAD_AES_256_CBC_HMAC_SHA_512-Deterministic");
+                    }else{
+                        encrypt.put("algorithm", "AEAD_AES_256_CBC_HMAC_SHA_512-Random");
+                    }
+                    encrypt.put("keyId", new JSONArray().put(binWrapper));
+                    encryptedFieldWrap.put("encrypt", encrypt);
+                    return encryptedFieldWrap;
+                }
+                public static JSONObject getStandardField(String chosenBsonType){
+                    JSONObject standardField = new JSONObject();
+                    standardField.put("bsonType", chosenBsonType);
+                    standardField.put("description", "must be a " + chosenBsonType + " and is required");
+                    return standardField;
+                }
+                public static String getJSONSchema(String keyId){
+                    JSONObject properties = new JSONObject();
+                    properties.put("fullName", getStandardField("string"));
+                    properties.put("address", getEncryptedField(keyId, "string", false));
+                    properties.put("ssn", getEncryptedField(keyId, "int", true));
+                    properties.put("last4SSN", getStandardField("int"));
+                    JSONObject patientInfo = new JSONObject();
+                    patientInfo.put("bsonType", "object");
+                    JSONObject patientInfoProperties = new JSONObject();
+                        patientInfoProperties.put("phone", getEncryptedField(keyId, "int", true));
+                        patientInfoProperties.put("provider", getStandardField("string"));
+                    patientInfo.put("properties", patientInfoProperties);
+                    properties.put("patientInfo", patientInfo);
+
+                    JSONObject medicalRecords = getEncryptedField(keyId, "array", false);
+                    properties.put("medicalRecords",medicalRecords);
+
+                    
+                    JSONObject patientsSchema = new JSONObject();
+                    patientsSchema.put("properties", properties);
+                    patientsSchema.put("bsonType", "object");
+                    String patientsSchemaAsString = patientsSchema.toString();
+                    return patientsSchemaAsString;
+                }
+                public static MongoClient createSecureClient(){
+                    ...
+                }
             }
 
-   
-TEST FOOBAR
+
+
+Running the makeJSONSchema method will return the following JSON schema:
+
+.. code::
+
+
+    
+    {
+        "bsonType": "object",
+        "properties": {
+            "address": {
+            "encrypt": {
+                "bsonType": "string",
+                "keyId": [
+                {
+                    "$binary": {
+                    "base64": <keyId>,
+                    "subType": "04"
+                    }
+                }
+                ],
+                "algorithm": "AEAD_AES_256_CBC_HMAC_SHA_512-Random"
+            }
+            },
+            "last4SSN": {
+            "bsonType": "int",
+            "description": "must be a int and is required"
+            },
+            "medicalRecords": {
+            "encrypt": {
+                "bsonType": "array",
+                "keyId": [
+                {
+                    "$binary": {
+                    "base64": <keyId>,
+                    "subType": "04"
+                    }
+                }
+                ],
+                "algorithm": "AEAD_AES_256_CBC_HMAC_SHA_512-Random"
+            }
+            },
+            "fullName": {
+            "bsonType": "string",
+            "description": "must be a string and is required"
+            },
+            "patientInfo": {
+            "bsonType": "object",
+            "properties": {
+                "phone": {
+                "encrypt": {
+                    "bsonType": "int",
+                    "keyId": [
+                    {
+                        "$binary": {
+                        "base64": <keyId>,
+                        "subType": "04"
+                        }
+                    }
+                    ],
+                    "algorithm": "AEAD_AES_256_CBC_HMAC_SHA_512-Deterministic"
+                }
+                },
+                "provider": {
+                "bsonType": "string",
+                "description": "must be a string and is required"
+                }
+            }
+            },
+            "ssn": {
+            "encrypt": {
+                "bsonType": "int",
+                "keyId": [
+                {
+                    "$binary": {
+                    "base64": "<keyId>",
+                    "subType": "04"
+                    }
+                }
+                ],
+                "algorithm": "AEAD_AES_256_CBC_HMAC_SHA_512-Deterministic"
+            }
+            }
+        }
+    }
 
 
 C. Configure the MongoDB Client
--- a/source/use-cases/sensitive-data-encryption.txt
+++ b/source/use-cases/sensitive-data-encryption.txt
@@ -100,7 +100,36 @@ testabc
             import java.io.IOException;
             import java.security.SecureRandom;
 
+.. tabs::
 
+   tabs:
+
+     - id: java-master-key-generator
+      name: "Java"
+      content: |
+
+         The following script generates a 96-byte local master key and saves it to a file called ``master-key.txt`` in the directory from which the script is executed.
+
+         .. code-block:: java
+
+           import java.io.FileOutputStream;
+          import java.io.IOException;
+          import java.security.SecureRandom;
+
+           public class CreateMasterKeyFile {
+            public static void main(final String[] args) {
+
+               final byte[] localMasterKey = new byte[96];
+              new SecureRandom().nextBytes(localMasterKey);
+
+               String path = "master-key.txt";
+              try (FileOutputStream stream = new FileOutputStream(path)) {
+                stream.write(localMasterKey);
+              } catch (IOException e)  {
+                e.printStackTrace();
+              }
+            }
+          }
 
 
 C. Configure the MongoDB Client
--- a/source/use-cases/sensitive-data-encryption.txt
+++ b/source/use-cases/sensitive-data-encryption.txt
@@ -84,8 +84,6 @@ The structure for our schema will be as follows:
 * Medical Records are randomly encrypted because entire array's are only encryptable with random encryption
 
 
-In this example, the same key is used for the each encrypted field in the document. The following code illustrates how to create the above schema and 
-
 .. tabs::
 
    tabs:
--- a/source/use-cases/sensitive-data-encryption.txt
+++ b/source/use-cases/sensitive-data-encryption.txt
@@ -131,6 +131,8 @@ testabc
             }
           }
 
+TEST FOOBAR
+
 
 C. Configure the MongoDB Client
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
--- a/source/use-cases/sensitive-data-encryption.txt
+++ b/source/use-cases/sensitive-data-encryption.txt
@@ -95,42 +95,13 @@ testabc
          The following code creates a collection using the schema above and inserts it into a database using MongoDB.
 
          .. code-block:: java
+            import java.util.function.Consumer;
+            import org.json.*;
 
-            import java.io.FileOutputStream;
-            import java.io.IOException;
-            import java.security.SecureRandom;
-
-.. tabs::
-
-   tabs:
-
-     - id: java-master-key-generator
-      name: "Java"
-      content: |
-
-         The following script generates a 96-byte local master key and saves it to a file called ``master-key.txt`` in the directory from which the script is executed.
-
-         .. code-block:: java
-
-           import java.io.FileOutputStream;
-          import java.io.IOException;
-          import java.security.SecureRandom;
-
-           public class CreateMasterKeyFile {
-            public static void main(final String[] args) {
-
-               final byte[] localMasterKey = new byte[96];
-              new SecureRandom().nextBytes(localMasterKey);
-
-               String path = "master-key.txt";
-              try (FileOutputStream stream = new FileOutputStream(path)) {
-                stream.write(localMasterKey);
-              } catch (IOException e)  {
-                e.printStackTrace();
-              }
+            public class FleApplication {
             }
-          }
 
+   
 TEST FOOBAR
 
 
--- a/source/use-cases/sensitive-data-encryption.txt
+++ b/source/use-cases/sensitive-data-encryption.txt
@@ -78,8 +78,7 @@ The structure for our schema will be as follows:
     appointments: Object
     medicalRecords: Encrypted array of documents using random encryption
 
-
-* Address randomly encrypted because it is more secure for strings to be randomly encrypted to prevent hackers from decrypting by building a cipher dictionary
+* Address randomly encrypted because it is more secure for strings to be randomly encrypted to prevent hackers from decrypting by building a cipher-dictionary
 * SSN is deterministically encrypted because deterministic encryption provides filtering ability 
 * Medical Records are randomly encrypted because entire array's are only encryptable with random encryption
 
--- a/source/use-cases/sensitive-data-encryption.txt
+++ b/source/use-cases/sensitive-data-encryption.txt
@@ -82,6 +82,7 @@ The structure for our schema will be as follows:
 * SSN is deterministically encrypted because deterministic encryption provides filtering ability 
 * Medical Records are randomly encrypted because entire array's are only encryptable with random encryption
 
+testabc
 
 .. tabs::
 
--- a/conf.py
+++ b/conf.py
@@ -32,6 +32,7 @@ extensions = [
     'mongodb',
     'directives',
     'intermanual',
+    'tabs',
     'fasthtml'
 ]
 
--- a/conf.py
+++ b/conf.py
@@ -25,6 +25,7 @@ sys.path.append(os.path.join(conf.paths.projectroot, conf.paths.buildsystem, 'sp
 
 needs_sphinx = '1.0'
 
+
 extensions = [
     'sphinx.ext.extlinks',
     'sphinx.ext.todo',
--- a/source/index.txt
+++ b/source/index.txt
@@ -20,7 +20,7 @@ MongoDB Drivers
 For information on MongoDB licensing, see `MongoDB Licensing
 <https://www.mongodb.com/community/licensing>`_.
 
-MongoDB ODM (Object-Document-Mapper) 
+MongoDB ODM (Object-Document-Mapper)
 ------------------------------------
 
 Mongoid is the officially supported ODM (Object-Document-Mapper)
@@ -62,3 +62,4 @@ Use Cases
 
    /use-cases/metadata-and-asset-management
    /use-cases/storing-comments
+   /use-cases/sensitive-data-encryption


source/use-cases/sensitive-data-encryption.txt
```diff
@@ -0,0 +1,312 @@
...
+    ssn: Encrypted Int using deterministic encryption
+    last4SSN: Int
+    patientInfo: Object
+    patientInfo.phone Encrypted Int using deterministic encryption
+    patientInfo.provider String
+    medicalRecords: Encrypted array of documents using random encryption
+
+Explanation of encryption choices:
+``````````````````````````````````
+* address is randomly encrypted because it is more secure for strings to be randomly encrypted to prevent hackers from decrypting by building a cipher-dictionary
```
cbush: Please hard-wrap your lines.

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -0,0 +1,312 @@
...
+    fullName: String
+    address: Encrypted String using random encryption
+    ssn: Encrypted Int using deterministic encryption
+    last4SSN: Int
+    patientInfo: Object
+    patientInfo.phone Encrypted Int using deterministic encryption
+    patientInfo.provider String
+    medicalRecords: Encrypted array of documents using random encryption
+
+Explanation of encryption choices:
```
cbush: Remove colon

source/use-cases/sensitive-data-encryption.txt
cbush: I think it would look better if you capitalize these lines

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -0,0 +1,312 @@
...
+``````````````````````````````````
+* address is randomly encrypted because it is more secure for strings to be randomly encrypted to prevent hackers from decrypting by building a cipher-dictionary
+* ssn is deterministically encrypted because deterministic encryption provides filtering ability
```
cbush: ssn -> SSN or ssn in double backticks

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -0,0 +1,312 @@
...
+    patientInfo: Object
+    patientInfo.phone Encrypted Int using deterministic encryption
+    patientInfo.provider String
+    medicalRecords: Encrypted array of documents using random encryption
+
+Explanation of encryption choices:
+``````````````````````````````````
+* address is randomly encrypted because it is more secure for strings to be randomly encrypted to prevent hackers from decrypting by building a cipher-dictionary
+* ssn is deterministically encrypted because deterministic encryption provides filtering ability 
+* patientInfo.phone is deterministically encrypted for the same reason as ssn, and also it provides an example of encryption with a nested document
```
cbush: I would put code entities in double backticks, i.e.
```
``patientInfo.phone`` is deterministically...
```

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -0,0 +1,312 @@
...
+* medicalRecords are randomly encrypted because entire array's are only encryptable with random encryption
```
cbush: medicalRecords ~are~ is (referring to the field, singular)

array's -> arrays

I'm not sure of the use of the word "entire" here.

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -0,0 +1,312 @@
...
+.. tabs::
+
+   tabs:
+
+     - id: java-jsonschema-generation
+       name: "Java"
+       content: |
+
+         The following code creates a collection using the schema above and inserts it into a database using MongoDB.
+         For the "Define a JSON Schema" section we've abstracted the createSecureClient, to focus on creation of a schema.
```
cbush: Remove comma

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -0,0 +1,312 @@
...
+            import java.util.Map;
+
+            import org.json.*;
+
+            public class FieldLevelEncryptionDemo {
+                static String dbName = "fieldLevelEncryptionDemo001";
+                static String collectionName = "patients";
+                public static void main(String[] args) {
+                    MongoCollection secureCollection = createSecureClient().getDatabase(dbName).getCollection(collectionName);
+                    ...
```
cbush: I would put this in an actual code comment... People should be able to copy and paste without syntax errors.

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -0,0 +1,312 @@
...
+   tabs:
+
+     - id: java-jsonschema-generation
+       name: "Java"
+       content: |
+
+         The following code creates a collection using the schema above and inserts it into a database using MongoDB.
+         For the "Define a JSON Schema" section we've abstracted the createSecureClient, to focus on creation of a schema.
+         But that function uses getJSONSchema and passes it a dataKey id in order to create a client which uses the JSONSchema
+         we want to use.
```
cbush: Is there a way to rewrite this sentence for clarity?

I would maybe remove "But" to start.

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -0,0 +1,312 @@
...
+                    
+                    JSONObject encryptedFieldWrap = new JSONObject();
+                    JSONObject encrypt = new JSONObject();
+
+                    encrypt.put("bsonType", chosenBsonType);
+                    if(isDeterministic){
+                        encrypt.put("algorithm", "AEAD_AES_256_CBC_HMAC_SHA_512-Deterministic");
+                    }else{
+                        encrypt.put("algorithm", "AEAD_AES_256_CBC_HMAC_SHA_512-Random");
+                    }
```
cbush: You have inconsistent formatting. Some of your methods have a space before the opening { and others don't. Usually you would put space after control statement keywords (if, else, etc.) and a space before an opening brace, but either way, please be consistent.

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -0,0 +1,312 @@
...
+
+                    
+                    JSONObject patientsSchema = new JSONObject();
+                    patientsSchema.put("properties", properties);
+                    patientsSchema.put("bsonType", "object");
+                    String patientsSchemaAsString = patientsSchema.toString();
+                    return patientsSchemaAsString;
+                }
+                public static MongoClient createSecureClient(){
+                    ...
```
cbush: Comment out

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -0,0 +1,312 @@
...
+E. Convert to a Remote Master Key
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Fusce porta lectus et enim vehicula malesuada. In porttitor ipsum eu
+dictum congue. Fusce pharetra sagittis lacus sed posuere. Donec porta
+posuere gravida. Quisque nisi massa, condimentum vitae ligula eget,
+fringilla malesuada quam. Mauris in semper nulla, non blandit tortor.
+Mauris consequat eget urna at eleifend. Suspendisse potenti. Fusce at
+justo malesuada, finibus felis non, dignissim nisl. Ut blandit nulla sit
+amet massa euismod vehicula.
```
cbush: Maybe instead of Lorem ipsum just put TODO or something so someone can see at a glance what needs to be done still.

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -0,0 +1,312 @@
...
+            import org.json.*;
+
+            public class FieldLevelEncryptionDemo {
+                static String dbName = "fieldLevelEncryptionDemo001";
+                static String collectionName = "patients";
+                public static void main(String[] args) {
+                    MongoCollection secureCollection = createSecureClient().getDatabase(dbName).getCollection(collectionName);
+                    ...
+                }
+                public static JSONObject getEncryptedField(String keyId, String chosenBsonType, Boolean isDeterministic){
```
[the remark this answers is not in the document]
ccho-mongodb: Nitpick: 
TLDR; use `boolean` as the type rather than `Boolean`
in Java, there is a `Boolean` class and a `boolean` primitive. Values of `true` or `false` refer to the primitive type. Java handles conversion automatically with "autoboxing", but no need to bother with the conversion if you define the `isDeterministic` parameter as the primitive type.

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -0,0 +1,312 @@
...
+                    JSONObject binWrapper = new JSONObject();
+                    binWrapper.put("$binary", binaryForKey);
+                    
+                    JSONObject encryptedFieldWrap = new JSONObject();
+                    JSONObject encrypt = new JSONObject();
+
+                    encrypt.put("bsonType", chosenBsonType);
+                    if(isDeterministic){
+                        encrypt.put("algorithm", "AEAD_AES_256_CBC_HMAC_SHA_512-Deterministic");
+                    }else{
```
ccho-mongodb: Nitpick: use spaces around parens and curly braces. Ref: https://google.github.io/styleguide/javaguide.html#s4.1-braces

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -0,0 +1,312 @@
...
+consectetur molestie pellentesque ac, blandit non libero.
+
+Problem
+~~~~~~~
+
+Pellentesque non mi et est congue mattis pretium sit amet mi. Praesent
+nec tortor nec arcu molestie dapibus. Phasellus at nunc ut quam molestie
+finibus. Suspendisse finibus lectus eget nulla eleifend, vel consequat
+leo rhoncus. Phasellus lacinia tortor id aliquet condimentum. Morbi
+sodales est a leo rhoncus, non pretium neque dignissim.
```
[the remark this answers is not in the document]
mongomoe: Sure, so I changed the target from master to "mongodb:fle-guide" which was nick's branch that I branched off of

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -0,0 +1,312 @@
...
+volutpat, massa laoreet varius dictum, leo odio porttitor ante, nec
+auctor tortor orci et mi. Maecenas tempor, lacus vehicula molestie
+pulvinar, ante eros faucibus odio, sed consequat quam tellus vel arcu.
+Vestibulum ante ipsum primis in faucibus orci luctus et ultrices posuere
+cubilia Curae; Nunc non interdum purus, ultricies laoreet tortor.
+
+B. Define a JSON Schema
+~~~~~~~~~~~~~~~~~~~~~~~
+
+MongoDB uses JSON Schema to to validate data which is written to a collection.
```
ccho-mongodb: I wonder if it would be helpful to link to the MongoDB JSON Schema Validation page (https://docs.mongodb.com/manual/core/schema-validation/) since the reader might not be familiar with what it is.

ccho-mongodb: Nitpick: it is not completely clear whether "written to a collection" refers to data that has already been written (which JSON Schema Validation does not automatically do) or that it happens when data is written. More specificity might be useful, e.g.
```
In MongoDB, we can create and specify a JSON Schema on a collection to validate insert and update operations.
```

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -0,0 +1,312 @@
...
+- deterministic or random. We can also specify the BSON Type of the field we are encrypting. 
+In addition, we can specify the key we want to use. This is particularly important, because
+it means that we can use different keys for different fields.
+
+In the following example, we will go over creation of a schema for a collection which represents patient data
+in a medical office. Some fields use deterministic encryption, and others use random (probabilistic encryption). 
+
+The structure for our schema will be as follows:
+
+.. code::
```
ccho-mongodb: It might tie the document together better if this was a "skeletal" form of the JSON schema output that you showed later. E.g.
```
{ 
  ...
  fullName: <String, required>,
  address: <Deterministically Encrypted String, required>,
...
}

File: source/use-cases/sensitive-data-encryption.txt [hunk 17]
```diff
@@ -0,0 +1,312 @@
...
+       content: |
+
+         The following code creates a collection using the schema above and inserts it into a database using MongoDB.
+         For the "Define a JSON Schema" section we've abstracted the createSecureClient, to focus on creation of a schema.
+         But that function uses getJSONSchema and passes it a dataKey id in order to create a client which uses the JSONSchema
+         we want to use. 
+
+         .. code-block:: java
+
+            import com.mongodb.*;
```
ccho-mongodb: Quick question: is this import necessary?

source/use-cases/sensitive-data-encryption.txt
ccho-mongodb: Question: Should we steer the user to creating the JSON Schema on both the collection (on the server) as well as the client and show them how to accomplish both here? Or do you think that should belong in another section?

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -0,0 +1,312 @@
...
+                    encryptedFieldWrap.put("encrypt", encrypt);
+                    return encryptedFieldWrap;
+                }
+                public static JSONObject getStandardField(String chosenBsonType){
+                    JSONObject standardField = new JSONObject();
+                    standardField.put("bsonType", chosenBsonType);
+                    standardField.put("description", "must be a " + chosenBsonType + " and is required");
+                    return standardField;
+                }
+                public static String getJSONSchema(String keyId){
```
ccho-mongodb: Similar comment on the method name prefix `get`; this could be `createJSONSchema`.

source/use-cases/sensitive-data-encryption.txt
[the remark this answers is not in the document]
mongomoe: Good catch, this is fixed now

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -0,0 +1,312 @@
...
+
+B. Define a JSON Schema
+~~~~~~~~~~~~~~~~~~~~~~~
+
+MongoDB uses JSON Schema to to validate data which is written to a collection.
+Within the context of Field Level Encryption, we use JSON Schema to specify several
+configuration details. For instance, we can specify the form of encryption we want to use 
+- deterministic or random. We can also specify the BSON Type of the field we are encrypting. 
+In addition, we can specify the key we want to use. This is particularly important, because
+it means that we can use different keys for different fields.
```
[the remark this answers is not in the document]
mongomoe: Cool, I switched all the 'we' use to 'you' use.

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -0,0 +1,312 @@
...
+
+
+.. tabs::
+
+   tabs:
+
+     - id: java-jsonschema-generation
+       name: "Java"
+       content: |
+
```
ccho-mongodb: At this point in the guide, the reader will have generated data encryption keys and have the corresponding keyIds. While that section has not been written yet, you can assume that the random and deterministic keys were generated and that those values should be inserted in the JSON Schema definition.

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -0,0 +1,312 @@
...
+auctor tortor orci et mi. Maecenas tempor, lacus vehicula molestie
+pulvinar, ante eros faucibus odio, sed consequat quam tellus vel arcu.
+Vestibulum ante ipsum primis in faucibus orci luctus et ultrices posuere
+cubilia Curae; Nunc non interdum purus, ultricies laoreet tortor.
+
+B. Define a JSON Schema
+~~~~~~~~~~~~~~~~~~~~~~~
+
+MongoDB uses JSON Schema to to validate data which is written to a collection.
+Within the context of Field Level Encryption, we use JSON Schema to specify several
```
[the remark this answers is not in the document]

source/use-cases/sensitive-data-encryption.txt
mongomoe: Do you mean comment out " MongoCollection secureCollection = createSecureClient().getDatabase(dbName).getCollection(collectionName);" in the main method?

ccho-mongodb: On line 99, you mentioned:

>  The following code creates a collection using the schema above and inserts it into a database using MongoDB.

I think we should make sure that the description and code match.

source/use-cases/sensitive-data-encryption.txt
mongomoe: Fixed in the latest commit, check it out when you have time and let me know if it still sounds unclear

source/use-cases/sensitive-data-encryption.txt
mongomoe: Good catch, fixed in latest commit

source/use-cases/sensitive-data-encryption.txt
mongomoe: Fixed in latest commit

source/use-cases/sensitive-data-encryption.txt
mongomoe: I'd actually prefer it as is. Since mostly Chris Cho & I are working on this, I think it should be fine & is already clear what needs to be done. But if the team would prefer "to do" to the lorem ipseum, I'll change it.  @ccho-mongodb & @nlarew what do you think?

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -0,0 +1,312 @@
...
+configuration details. For instance, we can specify the form of encryption we want to use 
+- deterministic or random. We can also specify the BSON Type of the field we are encrypting. 
+In addition, we can specify the key we want to use. This is particularly important, because
+it means that we can use different keys for different fields.
+
+In the following example, we will go over creation of a schema for a collection which represents patient data
```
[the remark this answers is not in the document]
mongomoe: Fixed in latest commit

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -0,0 +1,312 @@
...
+in a medical office. Some fields use deterministic encryption, and others use random (probabilistic encryption).
```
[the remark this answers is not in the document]
mongomoe: I wrote it how I wanted to: 'random encryption' is the same as 'probabilistic encryption'. But I realize how that might be unclear, so I've fixed this in the latest commit

source/use-cases/sensitive-data-encryption.txt
cbush: No, I mean the `...`s

source/use-cases/sensitive-data-encryption.txt
cbush: Fine with me!

source/use-cases/sensitive-data-encryption.txt
mongomoe: fixed, sorry about that

mongomoe: The first letter of each bullet point represents a field name which is in camel case, so capitalizing wouldn't make sense in this scenario

source/use-cases/sensitive-data-encryption.txt
mongomoe: Good idea with the double backticks, I've used that style in the latest commit

source/use-cases/sensitive-data-encryption.txt

source/use-cases/sensitive-data-encryption.txt
mongomoe: I switched "array's" to "arrays". As for the second suggestion, "entire" makes sense technically. This is because, technically, array items are encryptable just not via automatic encryption with the driver. For instance, if I wanted to add the item "hello world" and encrypt it manually myself and push it into an unencrypted array, that is completely possible. But this example is going over auto encrypting the entire array.

source/use-cases/sensitive-data-encryption.txt
mongomoe: Changed this whole sentence. Now it reads: " The following code creates a collection using the schema above and inserts it into a database using MongoDB.
The createSecureClient is a method that calls getJSONSchema and passes it a dataKey id in order to create a client that uses the JSONSchema returned from getJSONSchema. However, the createSecureClient is abstracted from this section for focus on the creation of schema. "

source/use-cases/sensitive-data-encryption.txt
mongomoe: Gotcha, I actually just commented out both because the MongoCollection secureCollection ... will throw an error if the createSecureClient() doesn't actually return a mongoclient (which it doesn't because its abstracted in this doc).

source/use-cases/sensitive-data-encryption.txt
nlarew: Might be a little more direct/clear to say "array fields cannot be encrypted deterministically"

i.e. array field vs array item/element

source/use-cases/sensitive-data-encryption.txt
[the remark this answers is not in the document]
> changed getEncryptedField to createEncryptedFieldType
> changed chosenBsonType to bsonType
> changed Boolean to boolean

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -0,0 +1,312 @@
...
+
+
+
+Running the makeJSONSchema method will return the following JSON schema:
+
+.. code::
+
+
+    
+    {
```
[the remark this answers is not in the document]
mongomoe: Fixed in latest commit

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -0,0 +1,312 @@
...
+
+                    encrypt.put("bsonType", chosenBsonType);
+                    if(isDeterministic){
+                        encrypt.put("algorithm", "AEAD_AES_256_CBC_HMAC_SHA_512-Deterministic");
+                    }else{
+                        encrypt.put("algorithm", "AEAD_AES_256_CBC_HMAC_SHA_512-Random");
+                    }
+                    encrypt.put("keyId", new JSONArray().put(binWrapper));
+                    encryptedFieldWrap.put("encrypt", encrypt);
+                    return encryptedFieldWrap;
```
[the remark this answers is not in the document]
mongomoe: I don't necessarily think it would be clearer, and I also think it would be an extra & tedious step of switching over to JSONObjectBuilder. I spent some time considering the various options for generating JSON with java, and there aren't any great options, so if I had to make an executive decision on this - I say we keep it as is

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -0,0 +1,312 @@
...
+                }
+                public static JSONObject getStandardField(String chosenBsonType){
```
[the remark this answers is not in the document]
> changed getStandardField to createStandardFieldType

source/use-cases/sensitive-data-encryption.txt
mongomoe: It throws the following error without it:
`le/fle/CleanFleApp.java:120: error: cannot find symbol
                var keyVaultSettings = ClientEncryptionSettings.builder()
                                       ^
  symbol:   variable ClientEncryptionSettings
  location: class CleanFleApp
/Users/mohammad.hunan/Code/docs-fle/examples/javaexample/src/main/java/com/example/fle/CleanFleApp.java:121: error: cannot find symbol
                                .keyVaultMongoClientSettings(MongoClientSettings.builder()
                                                             ^
  symbol:   variable MongoClientSettings
  location: class CleanFleApp
/Users/mohammad.hunan/Code/docs-fle/examples/javaexample/src/main/java/com/example/fle/CleanFleApp.java:132: error: cannot find symbol
                                AutoEncryptionSettings.builder()
                                ^
  symbol:   variable AutoEncryptionSettings
  location: class CleanFleApp
/Users/mohammad.hunan/Code/docs-fle/examples/javaexample/src/main/java/com/example/fle/CleanFleApp.java:141: error: cannot find symbol
                var clientSettings = MongoClientSettings.builder()
                                     ^
  symbol:   variable MongoClientSettings
  location: class CleanFleApp
Note: /Users/mohammad.hunan/Code/docs-fle/examples/javaexample/src/main/java/com/example/fle/CleanFleApp.java uses unchecked or unsafe operations.
Note: Recompile with -Xlint:unchecked for details.
4 errors
`

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -57,12 +57,222 @@ cubilia Curae; Nunc non interdum purus, ultricies laoreet tortor.
...
+    patientInfo.phone Encrypted Int using deterministic encryption
+    patientInfo.provider String
+    medicalRecords: Encrypted array of documents using random encryption
+
+Explanation of encryption choices
+``````````````````````````````````
+* ``address`` is randomly encrypted because it is more secure for strings to be randomly encrypted to prevent hackers from decrypting by building a cipher-dictionary
+* ``ssn`` is deterministically encrypted because deterministic encryption provides filtering ability 
+* ``patientInfo.phone`` is deterministically encrypted for the same reason as ssn, and also it provides an example of encryption with a nested document
+* ``medicalRecords`` are randomly encrypted because entire arrays are only encryptable with random encryption
```
nlarew: We should introduce the (unencrypted) schema in the **Problem** section of the guide intro so that we have immediate context. As part of that, we should call out which fields we want to encrypt and _why_ we want to encrypt it.

These bullets mostly cover the _how_ of encryption - e.g. we randomly encrypt `address`, but why does `address` need to be private in the first place?

source/use-cases/sensitive-data-encryption.txt
mongomoe: @nlarew

source/use-cases/sensitive-data-encryption.txt

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -57,12 +57,222 @@ cubilia Curae; Nunc non interdum purus, ultricies laoreet tortor.
...
+.. code::
+
+    fullName: String
+    address: Encrypted String using random encryption
+    ssn: Encrypted Int using deterministic encryption
+    last4SSN: Int
+    patientInfo: Object
+    patientInfo.phone Encrypted Int using deterministic encryption
+    patientInfo.provider String
+    medicalRecords: Encrypted array of documents using random encryption
```
nlarew: Can we make it a bit clearer that these bucket into three types of fields:

- Unencrypted fields
- Randomly encrypted fields
- Deterministically encrypted fields

The type information isn't all too helpful except to discuss why we must use random on array fields and we can cover it in the **Problem** section (see my comment below)

source/use-cases/sensitive-data-encryption.txt
mongomoe: How about: "The following code creates a JSONSchema object that enforces encryption on various fields"

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -57,12 +57,222 @@ cubilia Curae; Nunc non interdum purus, ultricies laoreet tortor.
...
+       content: |
+       
+        The following code creates a collection using the schema above and inserts
+        it into a database using MongoDB.
+        The createSecureClient is a method that calls getJSONSchema and passes it
+        a dataKey id in order to create a client that uses the JSONSchema returned
+        from getJSONSchema. However, the createSecureClient is abstracted from
+        this section for focus on the creation of schema. 
+
+        .. code-block:: java
```
nlarew: This code is nice as a template for implementing JSON Schema, but it's really long and leaves this section with little explanation of what's actually happening. We should explain what the `encrypt` field looks like generically and then demonstrate how we define it for each field to match our requirements.

Right now I have to parse through the code and the output to infer this stuff

nlarew: This might mean _not_ including the script as a single code block and instead splitting it up into individual pieces that we combine.

e.g.
- This is the generic schema for an encrypted field
  ```
  "<field name>": {
    "bsonType": "<BSON Type>",
    "encrypt": {
      "bsonType": "<BSON Type>",
      "algorithm": "<Encryption Algorithm>",
      "keyId": [
        {
          "$binary": {
            "base64": <keyId>,
            "subType": "04"
          }
        }
      ],
    }
  }
  ```
- You can generate the schema for an encrypted field using the following function...

source/use-cases/sensitive-data-encryption.txt
ccho-mongodb: Ok makes sense, it's in a part of the app that isn't shown.

source/use-cases/sensitive-data-encryption.txt
nlarew: Also, can we use the Java BSON package to construct the Binary instead of manually doing it? (legit question idk if there's some limitation here)

source/use-cases/sensitive-data-encryption.txt
mongomoe: Good choice @nlarew, fixed in the latest commit

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -0,0 +1,312 @@
...
+    {
+        "bsonType": "object",
+        "properties": {
+            "address": {
+            "encrypt": {
+                "bsonType": "string",
+                "keyId": [
+                {
+                    "$binary": {
+                    "base64": <keyId>,
```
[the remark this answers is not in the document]
mongomoe: Fixed in latest commit

source/use-cases/sensitive-data-encryption.txt
mongomoe: Hm, so I updated this paragraph to:
"        The following code creates a JSONSchema object using the above structure. 
        The createSecureClient is a method that calls createJSONSchema and passes it
        a dataKey id in order to create a client that uses the JSONSchema returned
        from createJSONSchema. However, the createSecureClient is abstracted from
        this section for focus on the creation of schema. "

Let me know if that reads better

source/use-cases/sensitive-data-encryption.txt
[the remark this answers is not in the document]
ccho-mongodb: Ok, to me the builder pattern (https://en.wikipedia.org/wiki/Builder_pattern) is cleaner because you can define everything in a chain which resembles the structure of a hierarchical document. 

A builder pattern is also used for the client settings in the Java documentation of CSFLE (http://mongodb.github.io/mongo-java-driver/3.11/driver/tutorials/client-side-encryption/) and will likely be used in the section that precedes this one to prevent the code from being tedious. I am happy to help with rewriting this block if you need.

I do not mind releasing as-is, but would consider changing if we want to brush up the code to be easier to read. If you're not certain about the benefits, could you find someone on the Java driver team and see what they think?

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -0,0 +1,312 @@
...
+                    JSONObject standardField = new JSONObject();
+                    standardField.put("bsonType", chosenBsonType);
+                    standardField.put("description", "must be a " + chosenBsonType + " and is required");
+                    return standardField;
+                }
+                public static String getJSONSchema(String keyId){
+                    JSONObject properties = new JSONObject();
+                    properties.put("fullName", getStandardField("string"));
+                    properties.put("address", getEncryptedField(keyId, "string", false));
+                    properties.put("ssn", getEncryptedField(keyId, "int", true));
```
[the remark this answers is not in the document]
ccho-mongodb: If we decide on `ssn_last4`, could we stick to snake case for the rest of the fields to stay consistent?

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -57,12 +57,222 @@ cubilia Curae; Nunc non interdum purus, ultricies laoreet tortor.
...
-porta.
+MongoDB uses JSON Schema to validate data that is written to a collection.
+Within the context of Field Level Encryption, JSONSchema is used to
+specify several configuration details.
+For instance, you can specify the form of encryption you want to use  -
+deterministic or random.
+You can also specify the BSON Type of the field you are encrypting. In
+addition, you can 
+specify the key you want to use. This is particularly important, because
+it means that you can use different keys for different fields.
```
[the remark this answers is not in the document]
mongomoe: Fixed in the latest commit:
- made it more opinionated
- cut out the bsonType (you're right it's not specific to FLE)

As for "You can use a single data key for all encrypted fields or you can use individual data keys that are specific to each field." - I almost that's true. I'm fairly certain that's an added security measure that we've talked about before.  @ccho-mongodb  can you verify?

mongomoe:
@nlarew about your top level comment:

your comment: "I'd suggest ssn and ssn_last4"
my answer: I don't think ssn_last4 is any more clear than last4SSN, and ssn_last4 doesn't follow the convention of camelcase which the rest of the doc uses

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -57,12 +57,222 @@ cubilia Curae; Nunc non interdum purus, ultricies laoreet tortor.
...
+                    return patientsSchemaAsString;
+                }
+                public static MongoClient createSecureClient() {
+                    // ...
+                }
+            }
+
+
+
+Running the makeJSONSchema method will return the following JSON schema:
```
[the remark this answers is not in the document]
mongomoe: Sorry it used to be called makeJSONSchema and is now called createJSONSchema. I've fixed that reference in the recent commit

source/use-cases/sensitive-data-encryption.txt
ccho-mongodb: Just to help me update the "Introduction/Problem/Solution" section, can you describe the use case for encrypting the address? In our planning, we did not mention that as one of the fields we would encrypt so what I wrote in the intro was under the assumption that only SSN and Medical Records were being encrypted:
https://docs.google.com/document/d/16fdXj-wTLeBNqe4FUwfc6dv_D4C3k2u7YZb_k2vcgMY/edit

source/use-cases/sensitive-data-encryption.txt
mongomoe: I think these bullets cover the "why" rather than "how". These bullets are answering the question of "why are we choosing a certain form of encryption for a certain field", rather than "why are we encrypting". I think it makes more sense to answer why we are choosing to use one form of encryption over another. I also think it's the better question to answer. 

It's fair to assume our readers are smart enough to draw the conclusion that "address is encrypted because people might not want that info being in plain text", but we can't make the assumption that our readers know why address should be randomly encrypted instead of deterministically encrypted.

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -60,12 +60,13 @@ B. Define a JSON Schema
...
-You can also specify the BSON Type of the field you are encrypting. In
-addition, you can 
-specify the key you want to use. This is particularly important, because
-it means that you can use different keys for different fields.
+
+* Drivers can automatically encrypt and decrypt fields
+* To do so, you need to specify two things in JSON Schema for each encrypted field:
+    a) The encryption type (deterministic or random)
+    b) The data key
+* You can use a single data key for all encrypted fields or you can use individual data keys that are specific to each field.
```
nlarew: FYI: I meant these bullet points as a reference for the structure of the previous paragraph moreso than a verbatim bulleted list

```rst
MongoDB natively supports data validation by allowing you to specify
a JSON Schema for each collection. In addition to standard schema
validation, you can specify field-level encryption details in a
collection's JSON Schema to allow MongoDB drivers to automatically
encrypt and decrypt fields.

To configure encryption for a field in a JSON Schema, you must specify
the following in the `encrypt` field of the field's schema:

- The encryption algorithm (i.e. deterministic or random)
- The data key used to encrypt and decrypt the field
- The BSON type of the field

.. code-block:: json
   
   "<field name>": {
     "encrypt": {
       "bsonType": "<BSON Type>",
       "algorithm": "<Encryption Algorithm>",
       "keyId": [
         {
           "$binary": {
             "base64": <keyId>,
             "subType": "04"
           }
         }
       ],
     }
   }
```

source/use-cases/sensitive-data-encryption.txt
ccho-mongodb: The statement you're proposing is fine, but what I was thinking that we should show in this section is:

- Create the JSONSchema object
- Add it to a collection (or create the collection with it specified, if that is easier) using `runCommand`. Explain why you add the validator to the collection
- Mention that the JSONSchema object will be used in the AutoEncryptionSettings in the client setup (upcoming section). Explain why you would add it to the autoencryption settings.

Does that work for you?

source/use-cases/sensitive-data-encryption.txt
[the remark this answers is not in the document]
ccho-mongodb: Yes, you can use a single data key for all encrypted fields. We could provide the opinion that using different keys per field for deterministic algorithms should be used to protect against frequency analysis attacks.

source/use-cases/sensitive-data-encryption.txt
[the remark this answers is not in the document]
nlarew: I suggested snake case here to demonstrate that `ssn_last4` is a modified version of the `ssn` field rather than a distinct, independent field. That said if it isn't clear feel free to disregard.

source/use-cases/sensitive-data-encryption.txt
nlarew: Hmmm not sure tbh. For now let's focus on getting it done client-side and then when the rest of the guide is filled out we can decide where the best home for server-side JSON schema is.

source/use-cases/sensitive-data-encryption.txt
nlarew: Agreed that "why are we choosing a certain form of encryption for a certain field" is important.

I still think it's important that we talk about "why are we encrypting this field at all" because it's part of the logical narrative (we have this data, we want to protect these fields because reasons) and because it helps non-US users who might not understand what a social security number even is, let alone why I would encrypt the full thing but not the last four digits.

nlarew: Is there a use case for filtering an SSN? Full SSN is very private and nobody should really ever search based on that afaict.

Similarly, I can imagine a use case where I would want to filter based on address (e.g. "show me only MedcoMD patients that live in Cleveland.") which would require deterministic encryption.

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -57,12 +57,220 @@ cubilia Curae; Nunc non interdum purus, ultricies laoreet tortor.
 B. Define a JSON Schema
 ~~~~~~~~~~~~~~~~~~~~~~~
 
-Proin non mi a felis luctus vulputate nec pharetra odio. Ut pretium
-scelerisque nulla, sed euismod nisi pellentesque sit amet. Proin mollis
-mauris eu libero ullamcorper rutrum. Pellentesque vitae enim sed magna
-accumsan maximus et quis mi. Ut vel laoreet turpis. Vivamus vitae odio
-sed arcu finibus pretium vel maximus nisi. Nullam gravida consequat
-porta.
+MongoDB uses JSON Schema to validate data that is written to a collection.
```
terakilobyte: Let's link this to https://docs.mongodb.com/manual/reference/operator/query/jsonSchema/#json-schema

source/use-cases/sensitive-data-encryption.txt
mongomoe: I added a reference to the schema validation link in that sentence, and I completely agree with the second comment as well. Both suggestions have been implemented and pushed in the most recent commit.

source/use-cases/sensitive-data-encryption.txt
mongomoe: Done

source/use-cases/sensitive-data-encryption.txt
mongomoe: I've already assumed that the keys were generated and that those values are inserted into the JSONSchema. Here is (a piece of) the json schema generated, which has the data keys. 

`{
...
     "medicalRecords": {
        "encrypt": {
            "bsonType": "array",
            "keyId": [
            {
                "$binary": {
                "base64": <keyId>,
                "subType": "04"
                }
            }
            ],
            "algorithm": "AEAD_AES_256_CBC_HMAC_SHA_512-Random"
        }
        }
...
}`

source/use-cases/sensitive-data-encryption.txt
mongomoe: Those three points works for me to add to the doc. Would you suggest putting those after that above line ^ ("The following code creates a collection using the schema above and inserts it into a database using MongoDB.")

source/use-cases/sensitive-data-encryption.txt
mongomoe: Done

source/use-cases/sensitive-data-encryption.txt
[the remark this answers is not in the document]
mongomoe: I'm going to executive decision this and say we stick to camel case. I think we could go back and fourth of snake case vs camel case forever and never reach a conclusion. But ultimately, my reasons for sticking to camel case boils down to:
1. Consistency among this page (if we change this to snake case, this entire page will have to be snake case)2
2. Consistency amongst docs-ecosystem in general
3. last4SSN/ssn_last4 isn't really a modified version of ssn (as in, the reader isn't going to interact with ssn to get last4SSN, they are just two separate fields which are put in a document)

source/use-cases/sensitive-data-encryption.txt
mongomoe: Cool

mongomoe: 💯

source/use-cases/sensitive-data-encryption.txt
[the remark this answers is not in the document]
mongomoe: Sure, let's keep it as is for the PR while we're awaiting input on the java drivers team, and then if they wholeheartedly agree to use the builder pattern we can draw up another ticket to switch to that pattern.

source/use-cases/sensitive-data-encryption.txt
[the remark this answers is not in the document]
mongomoe: I agree, however, I don't see how that inherently relates to JSONSchema, and hence isn't part of this section. Perhaps we can create another ticket to add that.

source/use-cases/sensitive-data-encryption.txt
mongomoe: On second thought, you're right. SSN makes more sense as randomly encrypted because it doesn't require filtering, and address should be deterministically encrypted because they do require filtering. Fixed in the most recent commit.

source/use-cases/sensitive-data-encryption.txt
mongomoe: I agree that adding a simple example of what an encrypted field looks like would be helpful. I've added that to the most recent commit. 

As for your question: 
"Also, can we use the Java BSON package to construct the Binary instead of manually doing it?"
I don't know the answer to that, but it's unrelated to this exact section - which is specific to JSONSchema.

source/use-cases/sensitive-data-encryption.txt
mongomoe: I actually like the a) ... b) ... format. Also, like you said, since bsonType isn't relevant to jsonSchema I've abstracted it for this piece of doc. I did take your suggestion on having that code block implemented.

This is implemented in the latest commit.

source/use-cases/sensitive-data-encryption.txt
mongomoe: Done

source/use-cases/sensitive-data-encryption.txt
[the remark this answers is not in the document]
ccho-mongodb: Did you purposely add more than one blank line between the `code` directive and the start of the JSON schema result?

source/use-cases/sensitive-data-encryption.txt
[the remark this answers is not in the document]
ccho-mongodb: The spacing is still not correct on lines that follow opened arrays like 224. Here's an example of what is output by the code snippet I provided:

```
        "keyId": [
          {
            "$binary": {
              "base64": "keyId",
              "subType": "04"
            }
          }
        ],
        "algorithm": "AEAD_AES_256_CBC_HMAC_SHA_512-Random"
      }
...
```

source/use-cases/sensitive-data-encryption.txt
ccho-mongodb: This actually changes the scenario which is currently:

```
In this scenario, we secure sensitive data on a Medical Care Management System which stores patients' personal information, appointments, and vital sign measurement reports for a fictitious company, *MedcoMD*. None of the patient data is public, and certain data such as their social security number (SSN) and vital sign measurements are particularly sensitive. It is important for the company and the patient that the data is kept private and secure.

MedcoMD needs this system to satisfy the following use cases:

- Doctors use the system to access Patients' medical records and add new vital sign measurements.
- Receptionists use the system to schedule appointments for Patients, who are identified by a combination of contact information and their SSN.
- Receptionists cannot access nor update Patients' medical records.

MedcoMD is also concerned with disclosure of private information through any of the following methods:

- Printing the information or leaving it displayed on a screen where unauthorized people can view it
- Communicating on an insecure network
- Gaining access to the database on one of the permissioned users

What can MedcoMD do to balance the functionality and access restrictions of their Medical Care Management System?```

Let's get together with Nick and figure out the best schema to use that fits a reasonable scenario as this is currently blocking work on the intro.

File: source/use-cases/sensitive-data-encryption.txt (hunk 29 above)
Comment (nlarew) [reply source/use-cases/sensitive-data-encryption.txt]: If we build documents imperatively then using the official BSON package to construct BSON objects is very relevant and should be our preferred approach if feasible

Comment (nlarew) [reply source/use-cases/sensitive-data-encryption.txt]: I'm wondering if the imperative, code-based approach we're taking here is actually the most effective approach. Right now we have two monolithic code blocks (one for the Java and one for the output). We also have content that focuses on the Java code but loses focus on actually explaining the "why" of each output schema.

This means I have to scroll back and forth a lot to map the context of each field to the schema. i.e. "which fields are encrypted?", "what encryption type does each field use?", and "what does the schema for a particular field/encryption type look like?" are all related but only the first two are grouped currently.

I think it would instead make sense to define the JSON Schema in an external `.json` file and then read & parse that file into our code. This has several advantages:

- It's way cleaner than imperative object creation
- It's consistent across all drivers (no need to write custom schema creation code for each driver)
- It makes it easier to copy-paste a completed schema that we provide into a file
- It allows us to put a small schema example in a code block right next to each field rather than separating the fields from the schema with a big code block. We can then give users a link to see/download the full JSON Schema without taking up a ton of screen real estate.

Let's sync on this in person / over slack. cc: @ccho-mongodb

Comment (nlarew) [reply source/use-cases/sensitive-data-encryption.txt]: ^ Essentially, the size of the "address" schema example in the most recent version is ideal - we should aim for all of the fields to use bite-size code blocks like that. We can also add line highlights to point out the key things that we mention in accompanying text.

File: source/use-cases/sensitive-data-encryption.txt (hunk 24 above)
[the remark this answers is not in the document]
Comment (mongomoe) [reply source/use-cases/sensitive-data-encryption.txt]: fixed

File: source/use-cases/sensitive-data-encryption.txt (hunk 30 above)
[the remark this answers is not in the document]
Comment (mongomoe) [reply source/use-cases/sensitive-data-encryption.txt]: I think it's good now

File: source/use-cases/sensitive-data-encryption.txt (hunk 34 above)
Comment (mongomoe) [reply source/use-cases/sensitive-data-encryption.txt]: Do you mean, the first bullet point as in "The encryption type (deterministic or random)" is re-iterating? If we're getting rid of that we might as well get rid of the data-key one

File: source/use-cases/sensitive-data-encryption.txt [hunk 36]
```diff
@@ -57,12 +57,242 @@ cubilia Curae; Nunc non interdum purus, ultricies laoreet tortor.
...
-Proin non mi a felis luctus vulputate nec pharetra odio. Ut pretium
-scelerisque nulla, sed euismod nisi pellentesque sit amet. Proin mollis
-mauris eu libero ullamcorper rutrum. Pellentesque vitae enim sed magna
-accumsan maximus et quis mi. Ut vel laoreet turpis. Vivamus vitae odio
-sed arcu finibus pretium vel maximus nisi. Nullam gravida consequat
-porta.
+In MongoDB, we can create and specify a `JSON Schema <https://docs.mongodb.com/manual/reference/operator/query/jsonSchema/#json-schema>`_ on a collection to `validate data 
+<https://docs.mongodb.com/manual/core/schema-validation/>`_ during insert and update operations.
+Within the context of Field Level Encryption, JSONSchema is used to
+specify several configuration details.
```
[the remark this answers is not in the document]
mongomoe: How's this: "JSONSchema is used to validate data written to a MongoDB collection, within the context of Field Level Encryption it is used to specify encryption-related configuration details"

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -57,12 +57,242 @@ cubilia Curae; Nunc non interdum purus, ultricies laoreet tortor.
...
+
+* Drivers can automatically encrypt and decrypt fields
+* To do so, you need to specify two things in JSON Schema for each encrypted field:
+    a) The encryption type (deterministic or random)
+    b) The data key
+* You can use a single data key for all encrypted fields or you can use individual data keys that are specific to each field.
+
+
+To create a field that is encrypted, you first have to define it's schema. The following illustrates a quick example of a JSONSchema of a deterministically encrypted field called address:
```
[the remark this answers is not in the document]
mongomoe: Fixed in the latest commit

source/use-cases/sensitive-data-encryption.txt
mongomoe: I agree that having large monolithic code blocks causes a lot of annoying scrolling. This guide (meaning the entire page, not just JSON schema section) is meant to be a complete example of an implementation of field-level encryption that could happen in real life. The one thing that I think can help in terms of breaking up the monolithic nature of the code blocks is separating the java code. So for instance, creating one code block with createEncryptedFieldType and then an explanation of what that results in, and then another code block of the method createStandardFieldType and then an explanation of that and then the final code block of getJSONSchema which uses the other functions to create the entire schema and the results of the schema in JSON form (so the second monolithic code-block, the output, would remain)


As for making the JSON schema an external file, I’ve considered this. Java generally doesn’t have any “great” ways of creating JSON like javascript does, so even if we pivot to another option, I am certain it will be both time-consuming and non-fruitful in terms of adding clarity or decreasing bulk to the page. 

So the three original options were:
External file 
As a string
Using a JSONObject creation import (our current use-case)

Number two would be the most absurd option unless the schema we want was really limited and small (this is the route the java drivers team took in their doc, which made sense for them since their example was just one field). Number one is usable but the issue here is that a JSON file can’t be dynamic. I think most readers will probably use a JSONObject creation library of some sort (there’s at least a few out there). In my opinion, we should keep using number 3 with the caveat that we split the huge code chunk into three code chunks per method that each has an explanation. 

Also note: I have a bias to number 3 because I’ve just spent the past week writing and rewriting the doc using the JSONObject method.

PR #531, opened 2019-08-26
Docsp 6483 json schema fle

mongomoe:
Jira Ticket: https://jira.mongodb.org/browse/DOCSP-6483?filter=-1
Staged: https://docs-mongodbcom-staging.corp.mongodb.com/ecosystem/docsworker/DOCSP-6483-json-schema-fle/use-cases/sensitive-data-encryption.html


source/use-cases/sensitive-data-encryption.txt
```diff
@@ -57,12 +57,127 @@ cubilia Curae; Nunc non interdum purus, ultricies laoreet tortor.
 B. Define a JSON Schema
 ~~~~~~~~~~~~~~~~~~~~~~~
 
-Proin non mi a felis luctus vulputate nec pharetra odio. Ut pretium
-scelerisque nulla, sed euismod nisi pellentesque sit amet. Proin mollis
-mauris eu libero ullamcorper rutrum. Pellentesque vitae enim sed magna
-accumsan maximus et quis mi. Ut vel laoreet turpis. Vivamus vitae odio
-sed arcu finibus pretium vel maximus nisi. Nullam gravida consequat
-porta.
+An important concept in database interaction is `data validation
```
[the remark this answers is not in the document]
nlarew: - Let's de-emphasize the data validation aspect of JSON schema. We should note that the use case and process of defining a validation schema is different than FLE so that our users are not confused, but otherwise we should focus on encryption (I realize this is a shift from some of my earlier comments)
- Let's emphasize why we use JSON schema to configure encryption (i.e. answer the question: "why did MongoDB engineer the feature this way?") and give some quick insight into how it works under the hood without giving the full technical breakdown.

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -57,12 +57,127 @@ cubilia Curae; Nunc non interdum purus, ultricies laoreet tortor.
...
+`JSON Schema <https://docs.mongodb.com/manual/reference/operator/query/jsonSchema/#json-schema>`_ 
+for validation during insert and update operations. Within the context of Field Level Encryption,
+JSON Schema is used to specify several configuration details, allowing the drivers
+to automatically encrypt and decrypt fields. 
+
+* To do so, you need to specify two things in JSON Schema for each encrypted field:
+    a) The encryption type (deterministic or random)
+    b) The data key
+    c) The BSON Type
+* You can use a single data key for all encrypted fields or you can use individual data keys that are specific to each field.
```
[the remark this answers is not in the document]
ccho-mongodb: I was asking whether line 72 was meant to be part of the bulleted list (ll. 69-71) because the outer list (ll. 68, 72) does not have a unifying theme. My suggestion was to provide a unifying theme "there are several decisions...". This could be accompanied by a description of the table and what the JSON Schema column names are before line 79.

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -57,12 +57,127 @@ cubilia Curae; Nunc non interdum purus, ultricies laoreet tortor.
...
+                    "keyId": <keyID>,
+                    "algorithm": "AEAD_AES_256_CBC_HMAC_SHA_512-Deterministic"
+                }
+            }
+        }
+    }    
+
+
+
+- View the full JSON Schema for this MedcoMD guide `here <https://raw.githubusercontent.com/mongodb/docs-assets/DOCSP-json-schema-helper-and-json/MedcoMDSchema.json>`_
```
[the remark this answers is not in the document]
ccho-mongodb: I think we are talking about two different comments here. This one asks to change the name of "MedcoMD guide" to "CSFLE Use Case Guide" or "MedcoMD Medical Care Management System" which would read:
```
View the full JSON Schema for this CSFLE Use Case Guide
```
or
```
View the full JSON Schema for the Medco Medical Management System
```

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -57,12 +57,127 @@ cubilia Curae; Nunc non interdum purus, ultricies laoreet tortor.
...
+            "properties": {
+                // copy and paste your fields here 
+            }
+        }
+
+For the MedcoMD system, we will create JSON objects for each field and paste them into the properties 
+object of the JSON Schema above.
+
+
+.. code::
```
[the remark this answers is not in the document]
ccho-mongodb: I am suggesting the same format as the one you linked, but maybe it was not clear enough. In my example, here are the fields to which each line belongs:

```
Property name: JSON Schema for Social Security Number (SSN)
```
```
Description: SSN is a 9-digit identification number that can be used to verify a patient's identity.

Note: keyId can be specified, but is omitted because it inherits the property from encryptMetadata ...
```
```
Code example: <Code example>
```

What you currently have:
```
Code comment with a combination of describing deterministic encryption and how it applies to the field:
// Deterministic encryption supports querying...  of the encrypted data.
```

A couple reasons I think formatting it the way I suggested is helpful:
1) By including the description in the code block, it will have to be re-written as a comment in every language example
2) You can use the "Note" directive to separate the description from any other context you want to provide

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -57,12 +57,126 @@ cubilia Curae; Nunc non interdum purus, ultricies laoreet tortor.
...
+    | Blood Type                  | Random              |  String                                      |
+    +-----------------------------+---------------------+----------------------------------------------+
+    | Medical Records             | Random              |  Array                                       |
+    +-----------------------------+---------------------+----------------------------------------------+
+    | Insurance: Policy Number    | Deterministic       |  Int (embedded inside `insurance` object)    |
+    +-----------------------------+---------------------+----------------------------------------------+
+    | Insurance: Provider         | Non-Encrypted       |  String (embedded inside `insurance` object) |
+    +-----------------------------+---------------------+----------------------------------------------+
+
+You will use the secure client, created in an earlier section, by passing it a JSON object that contains your schema.
```
ccho-mongodb: Nitpick: [style guide: use present tense](https://docs.mongodb.com/meta/style-guide/writing/use-present-tense/)

E.g.
```
You will use the secure client... -> Pass the schema as a JSON object to the secure client to specify the encrypted field properties.
```

source/use-cases/sensitive-data-encryption.txt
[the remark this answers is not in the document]
mongomoe: @ccho-mongodb @nlarew 

Thanks for your suggestions with this. I've created a brief paragraph that I think satisfies the requirements of: 
* explaining briefly what data validation and JSON schema briefly and linking to a doc with further explanation
* not using reductive phrases (like 'important' or 'simple')
* using active voice

Paragraph:

B. Define a JSON Schema

"While designing your database you may want to enforce specific datatypes on specific fields. MongoDB works with JSON Schema to reduce the complexity in data validation during insert and update operations. Within the context of Field Level Encryption, JSON Schema is used to specify several configuration details, allowing the drivers to automatically encrypt and decrypt fields."

What do you think?

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -57,12 +57,127 @@ cubilia Curae; Nunc non interdum purus, ultricies laoreet tortor.
...
+                    "algorithm": "AEAD_AES_256_CBC_HMAC_SHA_512-Deterministic"
+                }
+            }
+        }
+    }    
+
+
+
+- View the full JSON Schema for this MedcoMD guide `here <https://raw.githubusercontent.com/mongodb/docs-assets/DOCSP-json-schema-helper-and-json/MedcoMDSchema.json>`_
+- View the `helper code <https://raw.githubusercontent.com/mongodb/docs-assets/DOCSP-json-schema-helper-and-json/HelperCode.java>`_ to generate the full JSON Schema for this MedcoMD guide in java. Note: this helper code is not the full helper code for this guide, it only contains the code to generate a JSON Schema for this section.
```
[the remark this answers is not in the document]
ccho-mongodb: (I am guessing that github markdown broke your ordered list formatting and that the last three items are actually 3, 4, and 5)

1 - Thanks!
2a - Thanks!
2b - The comment was to use `createJSONSchema`, not `getJSONSchema` as `get` and `set` are usually used in the getter/setter pattern. Also, did you get a chance to talk to the Java drivers team about the builder pattern for the JSON Object that holds the schema?
2c - Sorry I meant *data key*. I was referencing this line:
```System.out.println(getJSONSchema("<paste_your_key_id_here>"));```

3 - Thanks!
5 - Ok, thanks. Is the code that loads the JSON Schema from the file be in the "Configure the Client" section? Or are we going to use `FieldLevelEncryptionJSONSchema.java` for that section?

File: source/use-cases/sensitive-data-encryption.txt [hunk 7]
```diff
@@ -57,12 +57,126 @@ cubilia Curae; Nunc non interdum purus, ultricies laoreet tortor.
...
+            "bsonType" : "object",
+            "encryptMetadata" : {
+                "keyId" : // copy and paste your keyID generated here
+            },
+            "properties": {
+                // copy and paste your fields here 
+            }
+        }
+
+For the MedcoMD system, we will create JSON objects for each field and paste them into the properties
```
ccho-mongodb: Nitpick: [style guide: use present tense](https://docs.mongodb.com/meta/style-guide/writing/use-present-tense/)

E.g.
```
Create JSON objects for each field and append them to the `properties` map.
```

source/use-cases/sensitive-data-encryption.txt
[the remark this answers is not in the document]
ccho-mongodb: This version is much better, but could still be more direct. Examples:
1) "works with JSON Schema to reduce..." - could be revised to explain how it uses JSON Schema rather than just "working with" it.
2) "within the context of..." - we could instead use this opportunity to mention that it is required rather than that it has some meaning within the context of CSFLE.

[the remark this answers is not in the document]
ccho-mongodb: Nitpick: "specific datatypes on specific fields" seems repetitive and maybe should only be used once.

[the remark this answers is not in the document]
mongomoe: Good ideas. I've updated it, I think you'll like the updated paragraph. It resolves your two points and the nitpick.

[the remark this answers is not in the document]
mongomoe: @ccho-mongodb

source/use-cases/sensitive-data-encryption.txt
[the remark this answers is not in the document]
mongomoe: I've switched the second bullet to a note and clarified the first bullet (nested list). I've updated the staging page. Let me know what you think @ccho-mongodb

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -57,12 +57,127 @@ cubilia Curae; Nunc non interdum purus, ultricies laoreet tortor.
...
+to automatically encrypt and decrypt fields. 
+
+* To do so, you need to specify two things in JSON Schema for each encrypted field:
+    a) The encryption type (deterministic or random)
+    b) The data key
+    c) The BSON Type
+* You can use a single data key for all encrypted fields or you can use individual data keys that are specific to each field.
+
+
+The following table represents the fields of data, MedcoMD will be using in their Medical Care Management System. Each field
```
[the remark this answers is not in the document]
mongomoe: How's this (it's most similar to option 1 in terms of using the medco example ): "The MedcoMD engineers receive specific requirements for the fields of data. Each field will be either unencrypted or encrypted. Encrypted fields will be encrypted through `Deterministic {link}` Encryption or `Random{link}` encryption. The following table illustrates the data model of the Medco Management System."


I decided not to go with #2 for several reasons:
* if we talk about filterability here we also have to talk about the impact of cardinality of a field in deciding what encryption to use
* the section with the specific field (syntax and explanation), explains why we made specific choices and talks about filterability and cardinality

I decided to not go w/ # 3 because we need to link to ravind's deterministic & random explanation and the table needs some explanation beyond only having a title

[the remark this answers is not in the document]
mongomoe: I also like this new paragraph because it uses an active voice. I've updated the staging link, check it out and let me know what you think @ccho-mongodb

source/use-cases/sensitive-data-encryption.txt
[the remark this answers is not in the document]
mongomoe: Sounds good. I went with "View the full JSON Schema for the Medco Medical Management System"

source/use-cases/sensitive-data-encryption.txt
mongomoe: Great idea, I've fixed it now

source/use-cases/sensitive-data-encryption.txt
mongomoe: good catch - fixed now

source/use-cases/sensitive-data-encryption.txt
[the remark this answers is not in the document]
mongomoe: 2b. I've switched it to createJSONSchema now. Also, I messaged Jeff from Java drivers team about builder pattern and waiting for his response.
2c. I see. I've add in an additional comment in the helper code (" // replace the "paste_your_key_id_here" with your data key id") in case "<paste_your_key_id_here>" was not clear enough
5. Going to defer to @nlarew on this one

source/use-cases/sensitive-data-encryption.txt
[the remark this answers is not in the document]
ccho-mongodb: This thread got marked outdated, so will comment on the new paragraph.

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -57,12 +57,119 @@ cubilia Curae; Nunc non interdum purus, ultricies laoreet tortor.
 B. Define a JSON Schema
 ~~~~~~~~~~~~~~~~~~~~~~~
 
-Proin non mi a felis luctus vulputate nec pharetra odio. Ut pretium
-scelerisque nulla, sed euismod nisi pellentesque sit amet. Proin mollis
-mauris eu libero ullamcorper rutrum. Pellentesque vitae enim sed magna
-accumsan maximus et quis mi. Ut vel laoreet turpis. Vivamus vitae odio
-sed arcu finibus pretium vel maximus nisi. Nullam gravida consequat
-porta.
+While designing your database, you may want to enforce particular datatypes in specific fields. MongoDB uses JSON Schema to reduce the complexity of data validation during insert and update operations. Field Level Encryption with MongoDB requires the use of JSON Schema to specify several configuration options, allowing the drivers to encrypt and decrypt fields automatically.
```
ccho-mongodb: I know what I suggested is very similar to what you have written, but it still seems focused on JSON schema for data validation and not enough on automatic encryption/decryption. The last sentence is also focused on "several configuration options" rather than directly "encryption algorithm/scheme".

How about something like this:
```
MongoDB uses JSON Schema to enforce particular datatypes for specific fields during insert and update operations. MongoDB CSFLE requires JSON Schema to identify which fields to automatically encrypt and decrypt, the data encryption key id, and the type of encryption algorithm to use.
```

source/use-cases/sensitive-data-encryption.txt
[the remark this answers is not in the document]
ccho-mongodb: 2c. That seems more verbose, but I'll leave it up to you.

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -57,12 +57,127 @@ cubilia Curae; Nunc non interdum purus, ultricies laoreet tortor.
...
+
+The following table represents the fields of data, MedcoMD will be using in their Medical Care Management System. Each field
+will be either unencrypted or encrypted. Encrypted fields will be encrypted through `Deterministic Encryption <https://docs.mongodb.com/manual/core/security-client-side-encryption/#deterministic-encryption>`_
+or `Random encryption <https://docs.mongodb.com/manual/core/security-client-side-encryption/#random-encryption>`_.
+
+.. table:: 
+    :widths: 50 50
+
+    +-----------------------------+---------------------+---------------------+
+    |Field type                   |Encryption Algorithm | BSON Type           |
```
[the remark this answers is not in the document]
ccho-mongodb: Ok, thanks.

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -57,12 +57,127 @@ cubilia Curae; Nunc non interdum purus, ultricies laoreet tortor.
...
+    +-----------------------------+---------------------+---------------------+
+    | Medical Records             | Random              |  Array              |
+    +-----------------------------+---------------------+---------------------+
+    | Insurance: Policy Number    | Deterministic       |  (Nested) Int       |
+    +-----------------------------+---------------------+---------------------+
+    | Insurance: Provider         | Non-Encrypted       |  (Nested) String    |
+    +-----------------------------+---------------------+---------------------+
+
+
+You will use the secure client, created in an earlier, by passing it a JSON object that contains your schema.
```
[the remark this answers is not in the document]
ccho-mongodb: Thanks (for also using active voice)!

source/use-cases/sensitive-data-encryption.txt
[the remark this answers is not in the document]
ccho-mongodb: I meant a formatted note, e.g.:
```
.. note::
   This helper code is not the full...
```
Also, a nitpick: avoid using "helper code" twice in a row in the same sentence.

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -57,12 +57,119 @@ cubilia Curae; Nunc non interdum purus, ultricies laoreet tortor.
...
+    | Blood Type                  | Random              |  String                                      |
+    +-----------------------------+---------------------+----------------------------------------------+
+    | Medical Records             | Random              |  Array                                       |
+    +-----------------------------+---------------------+----------------------------------------------+
+    | Insurance: Policy Number    | Deterministic       |  Int (embedded inside `insurance` object)    |
+    +-----------------------------+---------------------+----------------------------------------------+
+    | Insurance: Provider         | Non-Encrypted       |  String (embedded inside `insurance` object) |
+    +-----------------------------+---------------------+----------------------------------------------+
+
+Pass the schema as a JSON object to the secure client to specify the encrypted field properties.
```
ccho-mongodb: This transition is abrupt -- sorry if I missed that while doing a spot check in another review. In the prior sentence, you describe the Medco Management System data model, and then here, ask the reader to do something with JSON schema without connecting the two. You could try rearranging some of the content you have, e.g.:
```
line 93:
In order to ensure fields are encrypted with the appropriate encryption algorithm, data key, and BSON type, create a JSON Schema using the information in the data model and pass it to the secure client as shown below:


.. note::
   The encryptMetadata is used in this example to allow the encryption-specific properties to be inherited by fields defined in the properties section.
```

source/use-cases/sensitive-data-encryption.txt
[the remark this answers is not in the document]
ccho-mongodb: Since this is a single bullet, could it just be a top-level item? Also, I'm not sure if you intended to do this, but the formatting came out in bold from what I see on staging.
Screen Shot 2019-09-03 at 4 29 51 PM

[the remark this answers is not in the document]
mongomoe: Fixed, I removed the bold first bullet and made the rest of the bullets top-level; it now reads:

While designing your database, you may want to enforce particular datatypes in specific fields. MongoDB uses JSON Schema to reduce the complexity of data validation during insert and update operations. Client-Side Field Level Encryption with MongoDB requires the use of JSON Schema for automatic encryption. Encrypted fields require specifying the following configuration options:

The encryption algorithm (deterministic or random)
The data key
The BSON Type

Is this okay? @ccho-mongodb

source/use-cases/sensitive-data-encryption.txt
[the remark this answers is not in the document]
mongomoe: I see, I fixed it into a formatted note and removed one of the 'helper code' word uses

source/use-cases/sensitive-data-encryption.txt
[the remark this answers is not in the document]
mongomoe: I'm going to keep it "<paste_your_key_id_here>". I'd take clarity > less verbosity in this use case

source/use-cases/sensitive-data-encryption.txt
mongomoe: Fixed to your paragraph roughly:

"MongoDB uses JSON Schema to enforce particular datatypes for specific fields during insert and update operations. MongoDB CSFLE requires JSON Schema to identify fields to automatically encrypt and decrypt, the data encryption key id, and the type of encryption algorithm to use."

source/use-cases/sensitive-data-encryption.txt
mongomoe: Some things I agree with in your comment:
* The transition is abrupt
* I haven't connected the previous bit (data model) with the following bit (json schema config)

Some things I don't agree with and why:
* Specifying that we're using an encryption algorithm, data key, and BSON Type. I don't agree with this because we've already mentioned those 3 things that we need and it's demonstrated right below, at that point mentioning it here would be repetitive.
* Putting a note right after we say 'pass it to the secure client as shown below:' because it's a continuity issue where we say we're doing something below and then show a note instead of the thing we said we're going to show.

Here's what I chose to replace it with to address the transition and connection issue:
"
To automatically encrypt and decrypt create a JSON Schema using the information in the data model and pass the schema as a JSON object to the secure client to specify the encrypted field properties.
In the JSON, you can specify `encryptMetadata <https://docs.mongodb.com/manual/reference/security-client-side-automatic-json-schema/#encryptmetadata-schema-keyword>`_ that is passed down to the fields under the ``properties`` nested object. 
"

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -57,12 +57,122 @@ cubilia Curae; Nunc non interdum purus, ultricies laoreet tortor.
...
+    }    
+
+
+
+- View the full JSON Schema for the Medco Medical Management System `here <https://raw.githubusercontent.com/mongodb/docs-assets/DOCSP-json-schema-helper-and-json/MedcoMDSchema.json>`_
+- View the `helper code <https://raw.githubusercontent.com/mongodb/docs-assets/DOCSP-json-schema-helper-and-json/FieldLevelEncryptionJSONSchema.java>`_ to generate the full JSON Schema for this MedcoMD guide in Java. 
+
+.. note::
+
+    this helper code is not the full code for this guide; it only contains the code to generate a JSON Schema for this section
```
ccho-mongodb: In your previous `.. note::` directive, you started the sentence with a capitalized letter and ended with a period -- it would be great if we could standardize. I prefer the caps + period at the end, but can adjust all the instances of my own for this guide if there are any reasons to do it differently.

ccho-mongodb: Also, do you think we need this note? It's clear to me from the mention of "View the helper code to generate the full JSON Schema...".

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -57,12 +57,122 @@ cubilia Curae; Nunc non interdum purus, ultricies laoreet tortor.
...
+
+
+.. code:: javascript
+
+    // Random encryption does not support filtering and is best used on fields with low cardinality (very few unique fields) to prevent against frequency analysis attacks. Since there are only 8 blood types, this field is randomly encrypted. To decrypt this data, a privileged user would have to query on another field via a secure collection, and this data would be returned unencrypted.
+
+    "bloodType": {
+        "encrypt": {
+            "bsonType": "string",
+            "keyId": <keyId>,
```
ccho-mongodb: Nitpick: the other instances use `<keyID>`, would be good to standardize.

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -57,12 +57,122 @@ cubilia Curae; Nunc non interdum purus, ultricies laoreet tortor.
...
+            "bsonType": "int",
+            "keyId": <keyID>,
+            "algorithm": "AEAD_AES_256_CBC_HMAC_SHA_512-Deterministic"
+        }
+    }
+
+
+.. code:: javascript
+
+    // Random encryption does not support filtering and is best used on fields with low cardinality (very few unique fields) to prevent against frequency analysis attacks. Since there are only 8 blood types, this field is randomly encrypted. To decrypt this data, a privileged user would have to query on another field via a secure collection, and this data would be returned unencrypted.
```
ccho-mongodb: The last sentence is unclear to me, especially the part about "via a secure collection". Did you mean to say:
```
This field is decrypted by a CSFLE-enabled client with appropriate access once the document is retrieved.
```

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -57,12 +57,122 @@ cubilia Curae; Nunc non interdum purus, ultricies laoreet tortor.
...
+            "properties": {
+                // copy and paste your fields here 
+            }
+        }
+
+Create JSON objects for each field and append them to the `properties` map.
+
+.. code::
+
+    // Deterministic encryption supports querying and is equally as secure as random encryption for high cardinality (very unique) fields such as SSN. A privileged user in the MedcoMD Medical Care Management System might want to query on SSN to verify a patient's identity. One thing to note is that you do not have to include keyId if you specified it in the top-level encryptMetadata field. Though you could include it if you want different users with specific keys to be able to access a specific subset of the encrypted data.
```
ccho-mongodb: We should standardize `keyId` (whether `keyID`, `keyId`, or other). The last sentence is not complete -- perhaps you can fix it by removing "Though". See https://english.stackexchange.com/questions/120981/using-though-at-the-beginning-of-the-following-sentence

source/use-cases/sensitive-data-encryption.txt
ccho-mongodb: Will comment on the updated commit.

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -57,12 +57,122 @@ cubilia Curae; Nunc non interdum purus, ultricies laoreet tortor.
...
+    }    
+
+
+
+- View the full JSON Schema for the Medco Medical Management System `here <https://raw.githubusercontent.com/mongodb/docs-assets/DOCSP-json-schema-helper-and-json/MedcoMDSchema.json>`_
+- View the `helper code <https://raw.githubusercontent.com/mongodb/docs-assets/DOCSP-json-schema-helper-and-json/FieldLevelEncryptionJSONSchema.java>`_ to generate the full JSON Schema for this MedcoMD guide in Java. 
+
+.. note::
+
+    This helper code is not the full code for this guide; it only contains the code to generate a JSON Schema for this section.
```
ccho-mongodb: From a prior comment:
Do you think we need this note? It's clear to me from the mention of "View the helper code to generate the full JSON Schema...".

source/use-cases/sensitive-data-encryption.txt
mongomoe: * To address your first concern: valid point, I've changed this note to be capitalized and end with a period. 
* To address your second concern: I'm neutral to the note, but it was your suggestion originally "I meant a formatted note, e.g.:

.. note::
   This helper code is not the full..."

 perhaps @nlarew could weigh his opinion here.

source/use-cases/sensitive-data-encryption.txt
mongomoe: Most actually use <keyId>, fair point on standardizing. I've switched all to "keyId" since that is what the property is as well. I think that is the most clear

source/use-cases/sensitive-data-encryption.txt
mongomoe: I did not mean that, but that's fair so I switched the last sentence to yours.

source/use-cases/sensitive-data-encryption.txt
mongomoe: I am neutral on this, it was your suggestion to include the note originally.

source/use-cases/sensitive-data-encryption.txt
mongomoe: Fixed. It now reads:
* "One thing to note is that you do not have to include keyId if you specified it in the top-level encryptMetadata field. You could include keyId if you want different users with specific keys to be able to access a specific subset of the encrypted data."

source/use-cases/sensitive-data-encryption.txt
ccho-mongodb: Yes, my current suggestion is to omit the note completely.

source/use-cases/sensitive-data-encryption.txt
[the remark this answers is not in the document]
ccho-mongodb: This is really weird -- review comment threads seem to be breaking. My comment "I meant a formatted note, e.g..." has nothing to do with my comment here. I think what is happening is that changes are being committed that have a timestamp before my comment was posted.

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -57,12 +57,122 @@ cubilia Curae; Nunc non interdum purus, ultricies laoreet tortor.
...
+    +-----------------------------+---------------------+----------------------------------------------+
+    | Medical Records             | Random              |  Array                                       |
+    +-----------------------------+---------------------+----------------------------------------------+
+    | Insurance: Policy Number    | Deterministic       |  Int (embedded inside `insurance` object)    |
+    +-----------------------------+---------------------+----------------------------------------------+
+    | Insurance: Provider         | Non-Encrypted       |  String (embedded inside `insurance` object) |
+    +-----------------------------+---------------------+----------------------------------------------+
+
+To automatically encrypt and decrypt create a JSON Schema using the information in the data model and pass the schema as a JSON object to the secure client to specify the encrypted field properties.
+In the JSON, you can specify `encryptMetadata <https://docs.mongodb.com/manual/reference/security-client-side-automatic-json-schema/#encryptmetadata-schema-keyword>`_ that is passed down to the fields under the ``properties``
```
ccho-mongodb: Thanks for revising this sentence. I think there are a few minor grammatical errors in this version that make this less clear than it could be:

1) "...you can specify encryptMetadata that..."

It is not clear that you are talking about JSON key-value pairs inside of the encryptMetadata object. If I said "In the office, you can specify desk", it's unclear whether I meant "select a desk to sit at" or "choose what materials it should be made of". 

2) Another point that makes it unclear what is being "passed down" is that you are using the singular helper verb "is" which implies the object `encryptMetadata` and not the properties within it.

Here's my sample sentence again:
```   
The encryptMetadata is used in this example to allow the encryption-specific properties to be inherited by fields defined in the properties section.
```

source/use-cases/sensitive-data-encryption.txt
ccho-mongodb: I'm not asking you to write something that you did not mean. What did you mean to write? What does it mean to "query on another field via a secure collection"?

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -57,12 +57,122 @@ cubilia Curae; Nunc non interdum purus, ultricies laoreet tortor.
 B. Define a JSON Schema
 ~~~~~~~~~~~~~~~~~~~~~~~
 
-Proin non mi a felis luctus vulputate nec pharetra odio. Ut pretium
-scelerisque nulla, sed euismod nisi pellentesque sit amet. Proin mollis
-mauris eu libero ullamcorper rutrum. Pellentesque vitae enim sed magna
-accumsan maximus et quis mi. Ut vel laoreet turpis. Vivamus vitae odio
-sed arcu finibus pretium vel maximus nisi. Nullam gravida consequat
-porta.
+MongoDB uses JSON Schema to enforce particular datatypes for specific fields during insert and update operations. MongoDB CSFLE requires JSON Schema to identify fields to automatically encrypt and decrypt, the data encryption key id, and the type of encryption algorithm to use:
```
ccho-mongodb: Thanks for updating this sentence. I am not sure exactly what I had recommended, but I think to make it work with the list on lines 62-64, you could update the second sentence to something like this:
```
MongoDB CSFLE requires the following properties in the JSON Schema to identify the fields that should be automatically encrypted and decrypted:
```

source/use-cases/sensitive-data-encryption.txt
ccho-mongodb: If it's clear without it, can we remove it? Otherwise, keep it. Just let me know what you decide here.

source/use-cases/sensitive-data-encryption.txt
mongomoe: I meant that I did not mean those exact words. What I was trying to capture in that last sentence was:
* A privileged user, such as a doctor, can decrypt sensitive data only on deterministically encrypted fields
* A non-privileged user, such as a receptionist, cannot decrypt sensitive data at all

Perhaps a more clear way of stating this would be:

"While a CSFLE-enabled client cannot filter on a randomly encrypted field, it can perform a query on a deterministic field, or a field that is not encrypted at all. Doing so would return the decrypted results of all the fields including randomly encrypted fields, such as blood type."


So the full comment paragraph would read:
```
    // Random encryption does not support filtering and is best used on fields with low cardinality (very few unique fields) to prevent against frequency analysis attacks. Since there are only 8 blood types, this field is randomly encrypted. While a CSFLE-enabled client cannot filter on a randomly encrypted field, it can perform a query on a deterministic field, or a field that is not encrypted at all. Doing so would return the decrypted results of all the fields including randomly encrypted fields, such as blood type.
```

source/use-cases/sensitive-data-encryption.txt
mongomoe: Sure it's removed now

source/use-cases/sensitive-data-encryption.txt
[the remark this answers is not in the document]
mongomoe: Are you okay w/ resolving this discussion if you feel your concerns are addressed? I ask because it's easier for me to see what's left to fix in a PR when the only remaining unresolved comments are stuff I have yet to fix 

@ccho-mongodb

source/use-cases/sensitive-data-encryption.txt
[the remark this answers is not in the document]
mongomoe: Are you okay w/ resolving this discussion if you feel your concerns are addressed? Let me know if there's something left to fix w/ this comment thread

@ccho-mongodb

source/use-cases/sensitive-data-encryption.txt
mongomoe: Are you okay w/ resolving this discussion if you feel your concerns are addressed? Let me know if there's something left to fix w/ this comment thread

@ccho-mongodb

source/use-cases/sensitive-data-encryption.txt
mongomoe: Fixed, it's omitted now

source/use-cases/sensitive-data-encryption.txt
mongomoe: What do you think of the updated version ^ @ccho-mongodb

source/use-cases/sensitive-data-encryption.txt
mongomoe: How's this:

"To automatically encrypt and decrypt create a JSON Schema using the information in the data model and pass the schema as a JSON object to the secure client to specify the encrypted field properties.
In the JSON, you can specify `encryptMetadata <https://docs.mongodb.com/manual/reference/security-client-side-automatic-json-schema/#encryptmetadata-schema-keyword>`_ to specify top-level encryption Configuration options such as a keyId or encryption algorithm. Fields under the ``properties`` key inherit their encryption properties from encryptMetadata."

I think this clears up what is being inherited from encryptMetadata.

mongomoe: what do you think @ccho-mongodb

source/use-cases/sensitive-data-encryption.txt

ccho-mongodb:
I cleaned up the Java code to generate the JSON schema and posted it here: https://gist.github.com/ccho-mongodb/87caa1d19d9cec80f2c77f4b925be035

We may want to modify the code to write to a file since that is how I think we decided it would get set up in the Client setup step.

I forgot to add the comment from your file, but feel free to add it back:
```
// replace the "paste_your_key_id_here" with your data key id
```

ccho-mongodb:
Updated gist which writes the schema to a file "fle-schema.json"
https://gist.github.com/ccho-mongodb/088176b1bed3b9e54cdc0c2c3c537d1b

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -57,12 +57,116 @@ cubilia Curae; Nunc non interdum purus, ultricies laoreet tortor.
...
+    | Insurance: Policy Number    | Deterministic       |  Int (embedded inside `insurance` object)    |
+    +-----------------------------+---------------------+----------------------------------------------+
+    | Insurance: Provider         | Non-Encrypted       |  String (embedded inside `insurance` object) |
+    +-----------------------------+---------------------+----------------------------------------------+
+
+To automatically encrypt and decrypt create a JSON Schema using the information in the data model and pass the schema as a JSON object to the secure client to specify the encrypted field properties.
+In the JSON, you can specify `encryptMetadata <https://docs.mongodb.com/manual/reference/security-client-side-automatic-json-schema/#encryptmetadata-schema-keyword>`_ that is passed down to the fields under the ``properties``
+nested object. 
+
+.. code::
```
nlarew: We prefer to use a custom `code-block` directive instead of the `code` directive. It should work more or less the same as `code` but adds a copy button and some styling. You'll also need to specify the language of the code block (`none` is a valid language, if appropriate)

```
.. code-block:: json
   
   {
     ...
   }
```

source/use-cases/sensitive-data-encryption.txt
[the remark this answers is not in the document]
ccho-mongodb: @mongomoe Just wanted to surface this since I haven't seen any responses to this thread.

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -166,7 +166,7 @@ Create JSON objects for each field and append them to the `properties` map.
 
 
 - View the full JSON Schema for the Medco Medical Management System `here <https://raw.githubusercontent.com/mongodb/docs-assets/DOCSP-json-schema-helper-and-json/MedcoMDSchema.json>`_
-- View the `helper code <https://raw.githubusercontent.com/mongodb/docs-assets/DOCSP-json-schema-helper-and-json/FieldLevelEncryptionJSONSchema.java>`_ to generate the full JSON Schema for this MedcoMD guide in Java. 
+- View the `helper code <https://raw.githubusercontent.com/mongodb/docs-assets/DOCSP-json-schema-helper-and-json/FieldLevelEncryptionJSONSchema.java>`_ to generate the full JSON Schema for this MedcoMD guide in Java
```
ccho-mongodb: I have been using a period to close out bullet points that are complete sentences, similar to how it is done here: 
https://docs.mongodb.com/ecosystem/drivers/specs/

What do you think about following that pattern (also followed this in my CSFLE PRs)?

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -57,12 +57,116 @@ cubilia Curae; Nunc non interdum purus, ultricies laoreet tortor.
...
+            "bsonType" : "object",
+            "encryptMetadata" : {
+                "keyId" : // copy and paste your keyID generated here
+            },
+            "properties": {
+                // copy and paste your fields here 
+            }
+        }
+
+Create JSON objects for each field and append them to the `properties` map.
```
nlarew: Let's turn each one of these code blocks into a little subsection.
- Heading: Describe the field
- Text: Blurb that maps the schema to the choices we list in the table above. This could probably just be the first line comment from each code block.
- Code
e.g.
```rst
Social Security Number
++++++++++++++++++++++

Paragraph explaining what an SSN is, that we want to query on it, and that
it's high cardinality. This means we'll use deterministic encryption.

.. code-block:: json
   
   {
     ...
   }
```

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -57,12 +57,116 @@ cubilia Curae; Nunc non interdum purus, ultricies laoreet tortor.
...
+
+.. code::
+
+    // Deterministic encryption supports querying and is equally as secure as random encryption for high cardinality (very unique) fields such as SSN. A privileged user in the MedcoMD Medical Care Management System might want to query on SSN to verify a patient's identity. One thing to note is that you do not have to include keyId if you specified it in the top-level encryptMetadata field. You could include keyId if you want different users with specific keys to be able to access a specific subset of the encrypted data.
```
nlarew: - This content should not be in a code comment (applies to each schema in this section)
  - JSON doesn't support comments so this would hurt copy-pastability
  - Comments should add clarity specific to the code
- The "one thing to note" shouldn't be in prose, let's put it in an actual `.. note::`
- Let's be opinionated about key ids. We have them define encryptMetadata, so no need to have the field in each subschema.

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -57,12 +57,116 @@ cubilia Curae; Nunc non interdum purus, ultricies laoreet tortor.
...
+            "bsonType": "int",
+            "keyId": <keyId>,
+            "algorithm": "AEAD_AES_256_CBC_HMAC_SHA_512-Deterministic"
+        }
+    }
+
+
+.. code:: javascript
+
+    // Random encryption does not support filtering and is best used on fields with low cardinality (very few unique fields) to prevent against frequency analysis attacks. Since there are only 8 blood types, this field is randomly encrypted. This field is decrypted by a CSFLE-enabled client with appropriate access once the document is retrieved.
```
nlarew: Let's change up the order of this to focus less on random encryption initially and instead lead to that with the "why" for this/each field.

1. Blood types only have 8 varieties (low cardinality)
2. We don't want to search on blood type
3. Random encryption fits these requirements

source/use-cases/sensitive-data-encryption.txt
ccho-mongodb: First sentence:
- Missing a comma: "To automatically encrypt and decrypt, create a JSON Schema..."
- This might be a run-on sentence, even with the comma.
- You also might consider mentioning the MongoDB driver since it executes the encryption/decryption based on the data in the JSON Schema.
- Since this paragraph is right before the code samples, maybe this should signal that the reader should be taking action.
- By putting together all of these comments, here's an example pair of sentences (to avoid the run-on): "Create a JSON Schema corresponding to the data model above. The MongoDB driver uses the schema to automatically manage the encryption and decryption of the specified fields."

Second sentence:
- I think `Configuration` should be lowercase.
- Parallel structure is needed. Corrected: "...such as a keyId or an encryption algorithm."

Third sentence:
- I think this could use more standardized language ("in objects" rather than "under ... key"). E.g. "Fields in the `properties` object inherit any encryption properties specified in the `encryptMetadata` object."

source/use-cases/sensitive-data-encryption.txt
ccho-mongodb: It is much more clear. However, there are a couple minor issues:

- First sentence: subject should be "Fields that are randomly encrypted". Otherwise, you are referring to "Random encryption" which does not necessarily have anything to do with MongoDB or filters. Also, it might not be apparent to the reader that you are referring to "query filters".

- Would skip the third and fourth sentences and either replace with a sentence that straightforward such as "While randomly encrypted fields are not queryable, they are decrypted by the client when the document is retrieved."

- There is an extra comma in the second sentence. Corrected: "... on a deterministic field or a field that is not encrypted at all."

- Nitpick: (personal preference and what I was taught) Any quantity less than 10 should be written out; i.e. "eight" instead of 8.

- Nitpick: in written English, we should avoid starting sentences with a gerund if possible ("Doing").

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -57,12 +57,116 @@ cubilia Curae; Nunc non interdum purus, ultricies laoreet tortor.
...
-scelerisque nulla, sed euismod nisi pellentesque sit amet. Proin mollis
-mauris eu libero ullamcorper rutrum. Pellentesque vitae enim sed magna
-accumsan maximus et quis mi. Ut vel laoreet turpis. Vivamus vitae odio
-sed arcu finibus pretium vel maximus nisi. Nullam gravida consequat
-porta.
+MongoDB uses JSON Schema to enforce particular datatypes for specific fields during insert and update operations. MongoDB CSFLE requires JSON Schema to identify fields to automatically encrypt and decrypt, the data encryption key id, and the type of encryption algorithm to use:
+
+* The encryption algorithm (deterministic or random)
+* The data key
+* The BSON Type
```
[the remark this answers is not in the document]
mongomoe: I didn't realize that originally, but I've just tested it out and you're correct - only deterministically encrypted fields through an error when you don't give them a bsonType.

I've altered the bullet to be:
* The BSON Type (only required by deterministically encrypted fields)

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -57,12 +57,116 @@ cubilia Curae; Nunc non interdum purus, ultricies laoreet tortor.
...
+
+.. note::
+
+    A single data key can be used for all encrypted fields, or you can use individual data keys that are specific to each field.
```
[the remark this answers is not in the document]
mongomoe: I haven't updated the staging bc I'm torn on this one. I really like the general structure of your sentence, but I think the vagueness of the fragment "a combination of both" can be improved upon.

How's this:

"You can use a single data key for all encrypted fields, or individual data keys for each field. You can also use a single data key for some keys and a different data key for others."

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -57,12 +57,116 @@ cubilia Curae; Nunc non interdum purus, ultricies laoreet tortor.
...
+
+
+The MedcoMD engineers receive specific requirements for the fields of data. `Deterministic Encryption <https://docs.mongodb.com/manual/core/security-client-side-encryption/#deterministic-encryption>`_ and `Random encryption <https://docs.mongodb.com/manual/core/security-client-side-encryption/#random-encryption>`_ are the two different methods of encrypting sensitive fields. The following table illustrates the data model of the Medco Management System.
+
+.. table::
```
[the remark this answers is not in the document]

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -57,12 +57,116 @@ cubilia Curae; Nunc non interdum purus, ultricies laoreet tortor.
...
+* The encryption algorithm (deterministic or random)
+* The data key
+* The BSON Type
+
+.. note::
+
+    A single data key can be used for all encrypted fields, or you can use individual data keys that are specific to each field.
+
+
+The MedcoMD engineers receive specific requirements for the fields of data. `Deterministic Encryption <https://docs.mongodb.com/manual/core/security-client-side-encryption/#deterministic-encryption>`_ and `Random encryption <https://docs.mongodb.com/manual/core/security-client-side-encryption/#random-encryption>`_ are the two different methods of encrypting sensitive fields. The following table illustrates the data model of the Medco Management System.
```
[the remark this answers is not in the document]
mongomoe: Makes sense. I've moved the links out of the table's explanation and into the first occurrence of determin+rand encryption. 

The introduction of the table now says:

"The MedcoMD engineers receive specific requirements for the fields of data and their encryption strategies. The following table illustrates the data model of the Medco Management System." 

Let me know what you think

source/use-cases/sensitive-data-encryption.txt

source/use-cases/sensitive-data-encryption.txt
nlarew: Generally context dependent on something like this. The most important thing is to be internally consistent in the list, i.e. if one ends with a period they all should.

PR #532, opened 2019-08-28
FLE Guide - Requirements section (DOCSP-6419)

ccho-mongodb:
**JIRA:**
https://jira.mongodb.org/browse/DOCSP-6419

**Staging:**
https://docs-mongodbcom-staging.corp.mongodb.com/ecosystem/docsworker/fle-guide-reqs/use-cases/sensitive-data-encryption.html

### DIFF
--- a/source/use-cases/sensitive-data-encryption.txt
+++ b/source/use-cases/sensitive-data-encryption.txt
@@ -65,19 +65,7 @@ AWS Key Management Service (KMS)
     Support for other KMS providers may be added in the future.
 
 
-MongoDB Authorization Privileges
-  * In order to create data encryption keys, the client needs the
-    :manual:`ReadWrite </reference/built-in-roles/#readWrite>` 
-    built-in role on the key vault collection which is specified in the client 
-    configuration.
-
-  * If no additional data encryption keys need to be created, the client can 
-    be downgraded to the 
-    :manual:`listCollections </reference/privilege-actions/#listCollections>`
-    privilege on the key vault collection.
-
-
-Server or VM Permissions
+File System Permissions
   * The client application or a privileged user needs permissions to start the 
     :manual:`mongocryptd </core/security-client-side-encryption/#field-level-encryption-mongocryptd>`
     process on the host.
--- a/source/use-cases/sensitive-data-encryption.txt
+++ b/source/use-cases/sensitive-data-encryption.txt
@@ -40,29 +40,41 @@ Procedure
 Requirements
 ~~~~~~~~~~~~
 
-Server Software
-^^^^^^^^^^^^^^^
+MongoDB Server 4.2 Enterprise
+  * For installation instructions, refer to the 
+    `Enterprise Edition Installation Tutorials <https://docs.mongodb.com/manual/installation/#mongodb-enterprise-edition-installation-tutorials>`_.
 
-* **MongoDB Server 4.2 Enterprise**. `Installation guide <https://docs.mongodb.com/manual/installation/>`_.
 
-Client development
-^^^^^^^^^^^^^^^^^^
+MongoDB Driver Compatible with CSFLE 
+  * For a list of drivers that support CSFLE, refer to the
+    `driver compatibility table <https://docs.mongodb.com/manual/core/security-client-side-encryption/#driver-compatibility-table>`_.
 
-* **MongoDB Driver compatible with CSFLE**. Refer to the `driver compatibility table <https://docs.mongodb.com/manual/core/security-client-side-encryption/#driver-compatibility-table>`_.
 
-External Services
-^^^^^^^^^^^^^^^^^
-* **AWS Key Management Service**
-  - Access to an existing master key or to create a new one (`AWS KMS key creation guide <https://docs.aws.amazon.com/kms/latest/developerguide/create-keys.html>`_.
-  - IAM user with list and read access to the master key
+AWS Key Management Service (KMS)
+  * AWS KMS is required for the secure deployment of CSFLE.
 
-Permissions
-^^^^^^^^^^^
+  * Compatible drivers currently include integration with AWS KMS. Support for
+    other KMS providers may be added in the future.
 
-* **MongoDB**
-  - `ReadWrite access <https://docs.mongodb.com/manual/reference/built-in-roles/#readWrite>`_ for the client on the key vault collection. After initial setup, if no additional data keys will be created, only `listCollections <https://docs.mongodb.com/manual/reference/command/listCollections/#dbcmd.listCollections>`_ is necessary.
+  * The client application will need AWS KMS permissions to the GenerateDataKey,
+    Encrypt, and Decrypt actions.
+ 
 
-* Client application or a privileged user needs permissions to start the **mongocryptd** process on the client-host
+MongoDB Authorization Privileges
+  * In order to create data encryption keys, the client needs the
+    `ReadWrite <https://docs.mongodb.com/manual/reference/built-in-roles/#readWrite>`_ 
+    built-in role on the key vault collection which is specified in the client 
+    configuration.
+
+  * If no additional data encryption keys need to be created, the client can 
+    be downgraded to the `listCollections <https://docs.mongodb.com/manual/reference/privilege-actions/#listCollections>`_
+    privilege on the key vault collection.
+
+
+Server or VM Permissions
+  * The client application or a privileged user needs permissions to start the 
+    `mongocryptd <https://docs.mongodb.com/manual/core/security-client-side-encryption/#field-level-encryption-mongocryptd>`_
+    process on the host. While the drivers contain code to spawn the process.
 
 
 A. Create a Master Key
--- a/source/use-cases/sensitive-data-encryption.txt
+++ b/source/use-cases/sensitive-data-encryption.txt
@@ -57,7 +57,7 @@ AWS Key Management Service (KMS)
     * :aws-docs:`Encrypt <kms/latest/APIReference/API_Encrypt.html>`
     * :aws-docs:`Decrypt <kms/latest/APIReference/API_Decrypt.html>`
 
-    To learn how to create an IAM user to assign these
+    To learn how to create an IAM user and assign these
     permissions, see :aws-docs:`AWS IAM Users </IAM/latest/UserGuide/id_users.html>`
     .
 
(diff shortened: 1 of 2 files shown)


source/use-cases/sensitive-data-encryption.txt
```diff
@@ -40,9 +40,42 @@ Procedure
...
 
-- Point to setup and install directions
-- Nullam imperdiet lorem vitae vulputate lacinia.
-- Donec eget velit tincidunt, gravida diam ac, efficitur lacus.
+MongoDB Server 4.2 Enterprise
+  * For installation instructions, refer to the 
+    `Enterprise Edition Installation Tutorials <https://docs.mongodb.com/manual/installation/#mongodb-enterprise-edition-installation-tutorials>`_.
+
+
+MongoDB Driver Compatible with CSFLE
```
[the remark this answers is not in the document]
nlarew: We introduce and define CSFLE in the intro section, it's just not included in this PR. https://github.com/mongodb/docs-ecosystem/pull/526/files#diff-c5c354e57498148e8ef37f38712ce945R25

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -40,9 +40,42 @@ Procedure
...
+
+
+AWS Key Management Service (KMS)
+  * AWS KMS is required for the secure deployment of CSFLE.
+
+  * Compatible drivers currently include integration with AWS KMS. Support for
+    other KMS providers may be added in the future.
+
+  * The client application will need AWS KMS permissions to the GenerateDataKey,
+    Encrypt, and Decrypt actions.
```
nlarew: I think we can focus this a bit more on the _what_ rather than _why_. Comments for each AWS KMS bullet:

- "AWS KMS is required for the secure deployment of CSFLE."
  - We can remove the first bullet since we explain the _why_ of a requirement in the guide. 

- "Compatible drivers currently include integration with AWS KMS. Support for other KMS providers may be added in the future."
  - "Compatible _MongoDB_ drivers currently support key management through AWS KMS. Support for other KMS providers may be added in the future."

- "The client application will need AWS KMS permissions to the GenerateDataKey, Encrypt, and Decrypt actions."
  - I believe these are, specifically, "IAM role permissions" so let's call out that they need an IAM user with these permissions
  - Let's move this to be the first bullet point since it's the only actual requirement here.
  - nit: thoughts on monospacing the permission names? e.g. ``` ``GenerateDataKey`` ```

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -40,9 +40,42 @@ Procedure
...
+
+MongoDB Authorization Privileges
+  * In order to create data encryption keys, the client needs the
+    `ReadWrite <https://docs.mongodb.com/manual/reference/built-in-roles/#readWrite>`_ 
+    built-in role on the key vault collection which is specified in the client 
+    configuration.
+
+  * If no additional data encryption keys need to be created, the client can 
+    be downgraded to the `listCollections <https://docs.mongodb.com/manual/reference/privilege-actions/#listCollections>`_
+    privilege on the key vault collection.
```
nlarew: nit: Let's not frame this as "downgraded" since they likely wouldn't ever have the elevated privileges in the first place

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -40,9 +40,42 @@ Procedure
...
+
+
+Server or VM Permissions
+  * The client application or a privileged user needs permissions to start the 
+    `mongocryptd <https://docs.mongodb.com/manual/core/security-client-side-encryption/#field-level-encryption-mongocryptd>`_
+    process on the host. While the drivers contain code to spawn the process.
```
nlarew: > While the drivers contain code to spawn the process.

Looks like an incomplete fragment slipped in here

source/use-cases/sensitive-data-encryption.txt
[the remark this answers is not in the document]
ccho-mongodb: I can probably reword this to "Compatible MongoDB Driver" if we want to avoid using any abbreviations, but that is likely less clear. I would like to avoid any conceptual explanations for a "Requirements" section, but am happy to change it to whatever works best.

source/use-cases/sensitive-data-encryption.txt
ccho-mongodb: Whoops, will remove.

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -40,9 +40,42 @@ Procedure
...
+    `Enterprise Edition Installation Tutorials <https://docs.mongodb.com/manual/installation/#mongodb-enterprise-edition-installation-tutorials>`_.
+
+
+MongoDB Driver Compatible with CSFLE 
+  * For a list of drivers that support CSFLE, refer to the
+    `driver compatibility table <https://docs.mongodb.com/manual/core/security-client-side-encryption/#driver-compatibility-table>`_.
+
+
+AWS Key Management Service (KMS)
+  * AWS KMS is required for the secure deployment of CSFLE.
```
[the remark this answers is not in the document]
ccho-mongodb: Thanks for the link! I will reword this to match the article content.

source/use-cases/sensitive-data-encryption.txt
ccho-mongodb: Thanks for the suggestions, will incorporate the first two.

I will omit the third bullet point and rely on the link Moe posted (https://docs.atlas.mongodb.com/security-aws-kms/) to explain the setup including permissions. This might be a hint that there should be guides that instruct both DBA/DeevOps and Developers to do an end-to-end setup (if AWS KMS is used).

It does not bother me to call these IAM role permissions, but I think at least Encrypt/Decrypt are "key policies" that are defined on specific customer master keys that are then added to an IAM role. Pretty confusing. See: https://docs.aws.amazon.com/kms/latest/developerguide/kms-api-permissions-reference.html

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -40,9 +40,42 @@ Procedure
...
+    `ReadWrite <https://docs.mongodb.com/manual/reference/built-in-roles/#readWrite>`_ 
+    built-in role on the key vault collection which is specified in the client 
+    configuration.
+
+  * If no additional data encryption keys need to be created, the client can 
+    be downgraded to the `listCollections <https://docs.mongodb.com/manual/reference/privilege-actions/#listCollections>`_
+    privilege on the key vault collection.
+
+
+Server or VM Permissions
```
ccho-mongodb: Uncertain whether this is conventional terminology as well as whether it is clear. The next closest term I could think of is Client-Host. Any suggestions?

source/use-cases/sensitive-data-encryption.txt
ccho-mongodb: I was trying to identify the difference between what they permissions they would need for the data key creation step and everything following that.

On second thought, this auth requirement might be trivial to even include in this guide, especially if we assume the user is running through the guide using a local MongoDB replica set in which there are no auth restrictions.

Do you think it would be better to move this requirement into one of the steps (e.g. "Make sure your user has ReadWrite permissions for this Data Key setup step") or remove it?

source/use-cases/sensitive-data-encryption.txt
ccho-mongodb: On second thought, since that link is for encryption-at-rest, the references to Atlas might be confusing. Is it ok to copy and paste a couple of the bullet points from that article?

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -42,39 +42,45 @@ Requirements
...
 
 MongoDB Driver Compatible with CSFLE 
   * For a list of drivers that support CSFLE, refer to the
-    `driver compatibility table <https://docs.mongodb.com/manual/core/security-client-side-encryption/#driver-compatibility-table>`_.
+    :manual:`driver compatibility table </core/security-client-side-encryption/#driver-compatibility-table>`.
 
 
 AWS Key Management Service (KMS)
-  * AWS KMS is required for the secure deployment of CSFLE.
+  * The client application will need the following AWS KMS permissions to the
```
ccho-mongodb: Updated list, confirmed by Kenn White.

source/use-cases/sensitive-data-encryption.txt
ccho-mongodb: Let me know what you think about the updates.

source/use-cases/sensitive-data-encryption.txt
nlarew: How about **File System Permissions**?

source/use-cases/sensitive-data-encryption.txt
nlarew: We definitely should include the required permissions since I presume enterprise users will deploy to production at some point which won't use localhost. We should also include it in the steps as an admonition.

A few things about the mongodb roles:

- Can we reframe the bullet points to be more independent?
   - e.g.
     ```rst
     - To create data encryption keys...
     - To encrypt and decrypt data...
     ```
- These bullets seem inconsistent:
   - `readWrite` is a built-in database user role
   - `listCollections` is a privilege/action granted by both the `read` and `readWrite` built-in database user roles
  - Built-in roles apply to entire databases, so you would have a built-in role _on the database that contains the keyvault collection._

source/use-cases/sensitive-data-encryption.txt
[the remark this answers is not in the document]
nlarew: I think this is fine with CSFLE. The context of the guide should make it clear and the link out can clarify any confusion.

source/use-cases/sensitive-data-encryption.txt
nlarew: These look good to me

source/use-cases/sensitive-data-encryption.txt
ccho-mongodb: Discussed in-person; will remove this section and create a note in the Data Encryption Key section that explains what permissions are needed to create the key only. FLE team still waffling on where best to keep keys secure and standard naming across examples.

source/use-cases/sensitive-data-encryption.txt
ccho-mongodb: Works for me, thanks.

PR #534, opened 2019-09-03
Docsp 6484 crud field level encryption

mongomoe:
staging: https://docs-mongodbcom-staging.corp.mongodb.com/ecosystem/docsworker/DOCSP-6484-CRUD-FieldLevelEncryption/use-cases/sensitive-data-encryption.html#d-perform-encrypted-read-write-operations
jira: https://jira.mongodb.org/browse/DOCSP-6484?filter=-1

### DIFF
--- a/source/use-cases/sensitive-data-encryption.txt
+++ b/source/use-cases/sensitive-data-encryption.txt
@@ -478,21 +478,136 @@ following procedure to configure and instantiate the MongoDB client:
 E. Perform Encrypted Read/Write Operations
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-Etiam hendrerit posuere justo, ac hendrerit lorem pellentesque sed.
-Fusce ullamcorper neque eu libero aliquam, sed placerat sapien cursus.
-Vestibulum maximus magna sed velit feugiat tempor.
+The MedcoMD engineers now have a CSFLE-enabled client and can test that
+the client can perform queries that meet the requirements. Doctors
+should be able to read and write to all fields, and receptionists should
+only be allowed to read and write to non-sensitive fields. 
 
-Ut accumsan pellentesque metus, at porta felis dapibus vulputate. Aenean
-malesuada interdum libero in feugiat. Fusce pharetra massa eget leo
-varius, et gravida massa tempus.
+Insert a Document with Encrypted Fields
++++++++++++++++++++++++++++++++++++++++
 
-Ut mattis eros elit, sed ultrices purus lacinia sed. Curabitur tincidunt
-tellus eget neque porta dictum ac vel orci. Vestibulum ante ipsum primis
-in faucibus orci luctus et ultrices posuere cubilia Curae.
+MedcoMD engineers write a function to create a new patient record:
 
-Praesent feugiat arcu risus elementum eleifend. Nullam sed iaculis
-erat.
 
+.. tabs::
+
+   tabs:
+
+     - id: java-jsonschema-generation
+       name: "Java"
+       content: |
+       
+        .. code-block:: java
+
+            public static void insertPatient(MongoCollection collection, String name, int ssn, String bloodType, ArrayList<Document> medicalRecords, int policyNumber, String provider) {
+                Document patient = new Document()
+                    .append("name", name)
+                    .append("ssn", ssn)
+                    .append("bloodType", bloodType)
+                    .append("medicalRecords", medicalRecords);
+                Document insurance = new Document()
+                    .append("policyNumber", policyNumber)
+                    .append("provider", provider)
+                    .append("insurance", insurance);
+                collection.insertOne(patient);
+            }
+
+When a CSFLE-enabled client inserts a new patient record into the Medical Care 
+Management System, it automatically encrypts the fields. This operation 
+produces a document similar to the following:
+
+.. code-block:: json 
+
+    {
+        "_id": "5d7a7bbe6d58fd263b6d7315",
+        "name": "Jon Doe",
+        "ssn": "Ac+ZbPM+sk7gl7CJCcIzlRAQUJ+uo/0WhqX+KbTNdhqCszHucqXNiwqEUjkGlh7gK8pm2JhIs/P3//nkVP0dWu8pSs6TJnpfUwRjPfnI0TURzQ==",
+        "bloodType": "As+ZbPM+sk7gl7CJCcIzlRACESwHCTCtK/lQV9kF6/LRoL3mh59gzBVA42vGBVfLIycYWpfAy7ZCi2eRGEgMX5CrGl259Wfu6Zf/ELBVqQDnyQ==",
+        "medicalRecords": "As+ZbPM+sk7gl7CJCcIzlRAEFt249toVYOlvlC/79cAtQ5jvE/ukF1ZLxRZn1g0zBBtPnf6L0AFTKMVdNJnjMGPMTszYU58qRE9uMvCU05DVHYl8DJnbtGXXFRLJ7ElQOc=",
+        "insurance": {
+            "provider": "MaestCare",
+            "policyNumber": "Ac+ZbPM+sk7gl7CJCcIzlRAQm7kFhN1hy3l7Wt3BSpBMbvVSuiaDsf3UPF9bvJLTEcC+Ka+3kZI4SVZinj4tyc5uDYeyh6+7phpKrQo4CHWyg=="
+        }
+    }
+
+.. note:: 
+    Clients that do not have CSFLE configured will insert unencrypted data. We 
+    recommend using :manual:`server-side schema validation </core/schema-validation/>`
+    to enforce encrypted writes for fields that should be encrypted. 
+
+Query for Documents on a Deterministically Encrypted Field
++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
+
+Queries on encrypted fields can be done using traditional MongoDB driver
+methods as well. When a query is made using a client configured to use
+CSFLE, such as when a doctor in the MedcoMD management system searches
+for a patient with their SSN, the patient's data is returned
+unencrypted:
+
+.. code-block:: json 
+
+    {
+        "_id": "5d6ecdce70401f03b27448fc",
+        "name": "Jon Doe",
+        "ssn": 241014209,
+        "bloodType": "AB+",
+        "medicalRecords": [
+            {
+                "weight": 180,
+                "bloodPressure": "120/80"
+            }
+        ],
+        "insurance": {
+            "provider": "MaestCare",
+            "policyNumber": 123142
+        }
+    }
+
+
+.. note:: 
+    For queries using a client that is not configured to use CSFLE, such
+    as when receptionists in the MedcoMD management system search for a
+    patient with their ``ssn``, a null value is returned. A client without CSFLE 
+    configured cannot query on a sensitive field.
+
+Query for Documents on a Randomly Encrypted Field
++++++++++++++++++++++++++++++++++++++++++++++++++
+
+.. warning::
+    You cannot directly query for documents on a randomly encrypted field, however 
+    you can use another field to find the document that contains an approximation of
+    the randomly encrypted field data.
+
+MedcoMD engineers determined that the fields they randomly encrypted
+would not be used to find patients records. Had this been required, for example,
+if the patient's ``ssn`` was randomly encrypted, MedcoMD engineers
+could have included another plaintext field called ``last4ssn`` that
+contains the last 4 digits of the ``ssn`` field. They could then query
+on this field as a proxy for the ``ssn``.
+
+
+
+.. code-block:: json 
+   :emphasize-lines: 5
+
+    {
+        "_id": "5d6ecdce70401f03b27448fc",
+        "name": "Jon Doe",
+        "ssn": 241014209,
+        "last4ssn": 4209,
+        "bloodType": "AB+",
+        "medicalRecords": [
+            {
+                "weight": 180,
+                "bloodPressure": "120/80"
+            }
+        ],
+        "insurance": {
+            "provider": "MaestCare",
+            "policyNumber": 123142
+        }
+    }
+ 
 .. _fle-convert-to-a-remote-master-key:
 
 F. Convert to a Remote Master Key


source/use-cases/sensitive-data-encryption.txt
```diff
@@ -75,20 +75,61 @@ velit congue vel.
...
+   tabs:
+
+     - id: java-jsonschema-generation
+       name: "Java"
+       content: |
+       
+        .. code-block:: java
+
+            public static void insertPatient(MongoCollection collection, String name, int ssn, String bloodType, ArrayList<Document> medicalRecords, int policyNumber, String provider) {
+                Document patient = new Document();
```
ccho-mongodb: Could you use the `append` pattern from the driver docs to construct this? The insertOne example shows how to create one with a subdocument too:
 https://mongodb.github.io/mongo-java-driver/3.11/driver/tutorials/perform-write-operations/

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -75,20 +75,61 @@ velit congue vel.
...
-Fusce ullamcorper neque eu libero aliquam, sed placerat sapien cursus.
-Vestibulum maximus magna sed velit feugiat tempor.
+Read and write operations on encrypted fields use the same methods and have the capabilities of traditional MongoDB read and write operations with the added security of field-level encryption.
 
-Ut accumsan pellentesque metus, at porta felis dapibus vulputate. Aenean
-malesuada interdum libero in feugiat. Fusce pharetra massa eget leo
-varius, et gravida massa tempus.
+This section continues the sample guide of the MedcoMD Management System. Now that their schemas are defined, MedcoMD engineers are tasked with creating a system that allows doctors, who have access to the FLE Enabled client, to read and write to documents in the patient’s collection without restriction. Receptionists, who have access to the non-FLE enabled client, have restricted access to documents in the patient’s collection that prevents them from reading and writing to sensitive fields. 
+By following this section, you learn the difference in output when using a secure client with FLE enabled, and a client without FLE enabled.
```
nlarew: - Can we make this the second sentence of the section? Feels like it belongs with the content before this MedcoMD paragraph.
- "By following this section" -> "in this section"

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -75,20 +75,61 @@ velit congue vel.
...
+A patient is inserted into the patient’s collection using the following code.
```
nlarew: passive -> active

```rst
The following code inserts a document into the ``patients`` collection:
```

Also let's make this sentence its own line separate form the paragraph

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -75,20 +75,61 @@ velit congue vel.
...
+                patient.put("bloodType", bloodType);
+                patient.put("medicalRecords", medicalRecords);
+                Document insurance = new Document();
+                insurance.put("policyNumber", policyNumber);
+                insurance.put("provider", provider);
+                patient.put("insurance", insurance);
+                collection.insertOne(patient);
+            }
+
+A doctor starts entering her new patient’s information in the Medco Management System. When she hits submit the above code is executed using the FLE enabled client. The new patient is inserted into the database, with the sensitive fields encrypted and looks like this:
```
nlarew: This paragraph has some mixed tense and passive voice. We might want to frame this a little bit less narratively, i.e. not from the perspective of a specific doctor.

- Any client can create a patient document
- Doctors can read and modify any field in the patient document
- Doctors don't need to think about encryption because the client handles it automatically

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -75,20 +75,61 @@ velit congue vel.
...
+
+.. code::
+
+    Document{{_id=5d696fff112d8c439bae7068, name=Peter Parker, ssn=org.bson.types.Binary@afbb5147, bloodType=org.bson.types.Binary@713daeda, medicalRecords=org.bson.types.Binary@113f42c8, insurance=Document{{policyNumber=org.bson.types.Binary@806ab88d, provider=MaestCare}}}}
```
nlarew: - I think this would be more effective if we displayed a valid JSON interpretation of the document rather than this Java representation (essentially for the same reasons as the JSON schema section)
- Let's prettify this to split it across multiple lines. Horizontal scrolling code blocks kinda suck.

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -75,20 +75,61 @@ velit congue vel.
...
+
+A receptionist submits another patient’s information in the Medco Management System using the insertPatient method, and the results look like this:
```
nlarew: Similar to my previous comment, let's see if we can frame this in terms of "Receptionists" rather than "A receptionist". Also let's make this a read example and move the unencrypted insert into the note below.

- Receptionists can read any document but cannot see the values of encrypted fields
- Receptionists can't query based on an encrypted field like `ssn`
- Receptionists need to query based on an unencrypted field, like `ssnLast4`
- Here's an example of a document from the `patients` collection as Receptionists read them, where the encrypted fields are all asterisks `{ "foo": *********** }`

In the note:
- If CSFLE is only configured client-side, Receptionists can insert patient documents with unencrypted fields. Beware!
- Here's how you could solve this

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -75,20 +75,61 @@ velit congue vel.
...
+
+.. code::
+
+    Document{{_id=5d696fff112d8c439bae7068, name=Gwen Stacey, ssn=1112223333, bloodType=a-, medicalRecords=[{weight: 120, bloodPressure: 120/80}], insurance=Document{{policyNumber=4410, provider=PlatinumCare}}}}
+
+.. note::
+    The information that the receptionist inserted is unencrypted even for sensitive data, such as ssn You may want to logic in your application code to prevent from writing to a sensitive field with a client that does not have FLE enabled. Another alternative approach is to use server side field level encryption in addition to CSLFE, this would throw an error if plaintext is attempted to be inserted into a sensitive field.
```
nlarew: - "you may want to ____ logic"
- "to prevent clients not configured for CSFLE from writing to a sensitive field"
- Server-side FLE should probably be the main suggestion rather than an alternative IMO.

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -75,20 +75,61 @@ velit congue vel.
...
+
+You can read the document(s) using traditional MongoDB driver methods as well. In the following example, a doctor in the MedcoMD management system scans a patient’s wristband. A function is called to search via the SSN of the scanned wristband and the following results are returned:
```
nlarew: - We're switching between "MedcoMD/Doctors/Receptionists" and "you". Let's be consistent throughout this section.
- Let's draw a contrast with the reads from the Receptionist client
  - Doctors can search on encrypted fields like `ssn`
  - Doctors can't search on randomly encrypted fields like `bloodType`

nlarew: It may even make sense to show Doctor reads before Receptionist reads. i.e.

1. Doctors can insert documents
2. Doctors can query on encrypted fields (except random)
3. Receptionists can't query on encrypted fields
4. Warning note about servers-side FLE.

Eventually we will want SSFLE as a step in this guide and at that point we could delete the warning note and add a replacement step:


1. Doctors can insert documents
2. **Receptionists can't insert documents (SSFLE returns an error)**
3. Doctors can query on encrypted fields (except random)
4. Receptionists can't query on encrypted fields

source/use-cases/sensitive-data-encryption.txt

Switched put to append

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -75,20 +75,61 @@ velit congue vel.
...
 
-Etiam hendrerit posuere justo, ac hendrerit lorem pellentesque sed.
-Fusce ullamcorper neque eu libero aliquam, sed placerat sapien cursus.
-Vestibulum maximus magna sed velit feugiat tempor.
+Read and write operations on encrypted fields use the same methods and have the capabilities of traditional MongoDB read and write operations with the added security of field-level encryption.
 
-Ut accumsan pellentesque metus, at porta felis dapibus vulputate. Aenean
-malesuada interdum libero in feugiat. Fusce pharetra massa eget leo
-varius, et gravida massa tempus.
+This section continues the sample guide of the MedcoMD Management System. Now that their schemas are defined, MedcoMD engineers are tasked with creating a system that allows doctors, who have access to the FLE Enabled client, to read and write to documents in the patient’s collection without restriction. Receptionists, who have access to the non-FLE enabled client, have restricted access to documents in the patient’s collection that prevents them from reading and writing to sensitive fields.
```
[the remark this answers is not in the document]

source/use-cases/sensitive-data-encryption.txt

source/use-cases/sensitive-data-encryption.txt
mongomoe: So I altered this whole paragraph because I thought it was too bulky and not inline with the rest of the page, let me know what you think of the updated paragraph

source/use-cases/sensitive-data-encryption.txt
mongomoe: Hey @nlarew I've altered it a bit, but I'm having a hard time deciding on phrasing it without referencing the doctor since the doc has access to the CSFLE configured client. How would you phrase it?

source/use-cases/sensitive-data-encryption.txt
mongomoe: In the latest commit, I've switched over to JSON (partially because a prettified java Document still doesn't look great and also because of Nick's first bullet).

Let me know what you think

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -75,20 +75,106 @@ velit congue vel.
...
+            }
+        ],
+        "insurance": {
+            "provider": "PlatinumCare North East",
+            "policyNumber": 4410
+        }
+    }
+
+.. note::
+    All the data the receptionist inserts is unencrypted even if it is a
```
ccho-mongodb: What do you think about using "Any data" instead of "All the data"?

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -75,20 +75,106 @@ velit congue vel.
...
+    sensitive field, such as SSN. You may want to include application
+    logic preventing clients without CSFLE configured from writing to
+    sensitive fields. An alternative approach is to use Server Side
```
ccho-mongodb: Do we support server-side field level encryption? It's not clear to me how to implement that if so. Did you mean JSON Schema validation on the collection?

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -75,20 +75,106 @@ velit congue vel.
...
+    Field Level Encryption in combination with CSFLE. Doing so adds
+    further security by throwing an error if an attempt is made to
+    insert plaintext into a sensitive field.
```
ccho-mongodb: On line 131, you're using "plain-text". It would be great to consolidate to a single term. Personally, I prefer "plaintext".

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -75,20 +75,106 @@ velit congue vel.
...
+and receptionists should only be allowed to read and write to non-sensitive fields. 
+
+MedcoMD engineers create a function to insert a new patient:
+
+
+.. tabs::
+
+   tabs:
+
+     - id: java-jsonschema-generation
```
ccho-mongodb: We will likely address this during the full review, but I think the id needs to be the same as all the other tab so that when we have a second tab (e.g. Node), switching tabs will activate all the Java tabs or all the Node ones. I also think that we're using "tabid" and am not certain whether id provides any functionality in this context. Here is the directive I see in the main doc:
```
:tabid: use-cases-fle-guide-java
```

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -75,20 +75,106 @@ velit congue vel.
...
+       name: "Java"
+       content: |
+       
+        .. code-block:: java
+
+            public static void insertPatient(MongoCollection collection, String name, int ssn, String bloodType, ArrayList<Document> medicalRecords, int policyNumber, String provider) {
+                Document patient = new Document();
+                patient.append("name", name);
```
ccho-mongodb: What do you think about chaining these `append` statements to make it less verbose? Other example code used in this guide uses inline declarations and/or chaining for Documents.

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -75,20 +75,106 @@ velit congue vel.
...
+.. note::
+    All the data the receptionist inserts is unencrypted even if it is a
+    sensitive field, such as SSN. You may want to include application
+    logic preventing clients without CSFLE configured from writing to
+    sensitive fields. An alternative approach is to use Server Side
+    Field Level Encryption in combination with CSFLE. Doing so adds
+    further security by throwing an error if an attempt is made to
+    insert plaintext into a sensitive field. 
+
+You can read the document(s) using traditional MongoDB driver methods as
```
ccho-mongodb: How about "You can read documents containing encrypted fields using ..."

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -75,20 +75,106 @@ velit congue vel.
...
+well. In the following example, a doctor in the MedcoMD management
+system scans a patient’s wristband. A function is called to search via
+the SSN of the scanned wristband, and the following results are returned:
```
ccho-mongodb: Could we reword this to use active voice (instead of "a function is called...")? I think you could combine a phrase with the previous sentence to link the action ("scan a wristband") and the results ("queries MongoDB which returns..."). 
 
Also, did you mean "SSN on the wristband"? If you use "of", it implies an SSN belongs to a wristband rather than a patient.

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -75,20 +75,106 @@ velit congue vel.
...
+    All the data the receptionist inserts is unencrypted even if it is a
+    sensitive field, such as SSN. You may want to include application
+    logic preventing clients without CSFLE configured from writing to
+    sensitive fields. An alternative approach is to use Server Side
+    Field Level Encryption in combination with CSFLE. Doing so adds
+    further security by throwing an error if an attempt is made to
+    insert plaintext into a sensitive field. 
+
+You can read the document(s) using traditional MongoDB driver methods as
+well. In the following example, a doctor in the MedcoMD management
```
ccho-mongodb: Is the subject "A doctor in the MedcoMD Management system"? Or did you mean that the doctor scans a wristband which gets entered into the MedcoMD management system?

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -75,20 +75,106 @@ velit congue vel.
...
+        "ssn": "Ac+ZbPM+sk7gl7CJCcIzlRAQUJ+uo/0WhqX+KbTNdhqCszHucqXNiwqEUjkGlh7gK8pm2JhIs/P3//nkVP0dWu8pSs6TJnpfUwRjPfnI0TURzQ==",
+        "bloodType": "As+ZbPM+sk7gl7CJCcIzlRACESwHCTCtK/lQV9kF6/LRoL3mh59gzBVA42vGBVfLIycYWpfAy7ZCi2eRGEgMX5CrGl259Wfu6Zf/ELBVqQDnyQ==",
+        "medicalRecords": "As+ZbPM+sk7gl7CJCcIzlRAEFt249toVYOlvlC/79cAtQ5jvE/ukF1ZLxRZn1g0zBBtPnf6L0AFTKMVdNJnjMGPMTszYU58qRE9uMvCU05DVHYl8DJnbtGXXFRLJ7ElQOc=",
+        "insurance": {
+            "provider": "MaestCare",
+            "policyNumber": "Ac+ZbPM+sk7gl7CJCcIzlRAQm7kFhN1hy3l7Wt3BSpBMbvVSuiaDsf3UPF9bvJLTEcC+Ka+3kZI4SVZinj4tyc5uDYeyh6+7phpKrQo4CHWyg=="
+        }
+    }
+
+When a new record is inserted using a client that is not configured to
```
ccho-mongodb: Could we make it more prominent to the reader that this as a hypothetical situation that they should avoid?

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -75,20 +75,106 @@ velit congue vel.
...
+system scans a patient’s wristband. A function is called to search via
+the SSN of the scanned wristband, and the following results are returned:
+
+
+
+.. code::
+
+    Document{{_id=5d6ecdce70401f03b27448fc, name=Peter Parker, ssn=241014209, bloodType=AB+, medicalRecords=[Document{{weight=180, bloodPressure=120/80}}], insurance=Document{{policyNumber=123142, provider=PlatinumCare}}}}
+
+However, if a receptionist tried to scan the same patient’s wristband,
```
ccho-mongodb: Should we maintain tense here? It looks like it is in past tense, but the rest of the document is in present tense.

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -75,20 +75,106 @@ velit congue vel.
...
+                insurance.append("policyNumber", policyNumber);
+                insurance.append("provider", provider);
+                patient.append("insurance", insurance);
+                collection.insertOne(patient);
+            }
+
+
+When a new record is inserted using a client configured to use CSFLE,
+such as when a doctor inserts a new patient in the MedcoMD Management
+System, the sensitive fields become encrypted:
```
ccho-mongodb: I could be mistaken, but I think "become" implies "will become" which is passive voice. Instead, you could use "are".

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -75,20 +75,106 @@ velit congue vel.
...
+the SSN of the scanned wristband, and the following results are returned:
+
+
+
+.. code::
+
+    Document{{_id=5d6ecdce70401f03b27448fc, name=Peter Parker, ssn=241014209, bloodType=AB+, medicalRecords=[Document{{weight=180, bloodPressure=120/80}}], insurance=Document{{policyNumber=123142, provider=PlatinumCare}}}}
+
+However, if a receptionist tried to scan the same patient’s wristband,
+they would receive a null result. This is because a client without CSFLE
```
ccho-mongodb: Could we combine these two sentences to ensure the reader connects them mentally? Something like the following:
"If a receptionist attempted to query a patient by their SSN, they would not retrieve any results because their client application does not have access to the encrypted, sensitive fields."

source/use-cases/Untitled-1
```diff
@@ -0,0 +1 @@
+mgeneratejs '{"name": "$name", "address": "$address", "phone": { "$integer": {"min": 3471111111, "max":8009999999 }, "ssn":{"$integer": {"min": 111111111, "max": 999999999}}, "ssnLast4": "0000", "bloodType": { "$choose": { "from": ["ab-", "b-","ab+","a-"], weight: [1,1,1,1]}} } ,"insurance": "policyNumber": { "$integer": {"min": 11111, "max":99999 } , "provider": {"$choose": { "from": ["MedSilver Insure", "HealthAid Pro", "Caresimple Co", "HealthConnect Insure"], "weights": [1,1,1,1] } } }, "medicalRecords": {"$array": {"of": {"weight": "{{chance.natural({min: 160, max: 180})}}", "heartRate": "{{chance.natural({min: 70, max: 100})}}", "bloodPressure": { "$choose": { "from": ["120/80", "115/75", "120/70"], weight: [1,1,1]}} }, "number": "{{chance.natural({ min: 1, max: 10})}}" }} }'  -n 1000
```
ccho-mongodb: Did you mean to commit this file? I was not able to find it referenced in the article.

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -75,20 +75,106 @@ velit congue vel.
...
-
-Praesent feugiat arcu non risus elementum eleifend. Nullam sed iaculis
-erat.
+Read and write operations on encrypted fields use the same methods and have the capabilities of traditional MongoDB read and write operations with the added security of field-level encryption.
+
+MedcoMD engineers continue the development of the Medical Care Management 
+System. In the system, doctors should be able to read and write to all fields, 
+and receptionists should only be allowed to read and write to non-sensitive fields. 
+
+MedcoMD engineers create a function to insert a new patient:
```
ccho-mongodb: Would it make more sense to isolate the CRUD terminology by limiting use of the verbs?

"Medco engineers write a function to **create** a new patient record" (bold is optional, just used for demonstration)

source/use-cases/sensitive-data-encryption.txt
mongomoe: This suggestion, I've added in the latest commit ^

source/use-cases/sensitive-data-encryption.txt
mongomoe: Fixed!

source/use-cases/sensitive-data-encryption.txt
mongomoe: Fixed!

source/use-cases/sensitive-data-encryption.txt
mongomoe: That sounds good, fixed in latest commit

source/use-cases/sensitive-data-encryption.txt
mongomoe: *  I'm not entirely sure how to reword this sentence while retaining clarity and ease of understanding. Can you give any suggestions? 
* Fixed, the sentence now reads: "A function is called to search via the SSN on the scanned wristband ..."

source/use-cases/sensitive-data-encryption.txt
mongomoe: Addressed all these comments in the latest commit

source/use-cases/sensitive-data-encryption.txt
mongomoe: Resolving since this is well outdated, and we've talked about this in person

source/use-cases/sensitive-data-encryption.txt
mongomoe: Resolving since this is outdated & that sentence is no longer in the guide

source/use-cases/sensitive-data-encryption.txt
mongomoe: Resolving this since this is outdated & that sentence is no longer in the guide

source/use-cases/sensitive-data-encryption.txt
mongomoe: Fixed, I switched this over to a note

source/use-cases/sensitive-data-encryption.txt
mongomoe: Resolving since this is outdated & no longer in the guide. The current version of this uses present tense

source/use-cases/sensitive-data-encryption.txt
mongomoe: Resolving since this is outdated & no longer in the guide

source/use-cases/Untitled-1
mongomoe: Sorry about that, I've removed it now

source/use-cases/sensitive-data-encryption.txt
mongomoe: Makes sense, fixed in latest commit

PR #535, opened 2019-09-03
DOCSP-6562 FLE guide data keys

ccho-mongodb:
JIRA: https://jira.mongodb.org/browse/DOCSP-6562

Staging: https://docs-mongodborg-staging.corp.mongodb.com/ecosystem/docsworker/fle-guide-data-keys/use-cases/sensitive-data-encryption.html

### DIFF
--- a/source/includes/steps-create-data-encryption-key.yaml
+++ b/source/includes/steps-create-data-encryption-key.yaml
@@ -103,8 +103,15 @@ title: Verify the data encryption key was created
 ref: verify-data-key-created
 level: 4
 content: |
-  Verify that the key was created in your MongoDB replica set in the specified
-  key vault namespace. The document should resemble the following:
+  Verify that the key document was created in your MongoDB replica set in the 
+  key vault collection. This document contains the following data:
+  
+  * Data encryption key, in encrypted form.
+  * Data encryption key UUID.
+  * KMS provider information for the master key.
+  * Other metadata such as creation and last modified date.
+ 
+  The document in the key vault collection should resemble the following:
 
   .. code-block:: none
   
--- a/source/includes/steps-create-data-encryption-key.yaml
+++ b/source/includes/steps-create-data-encryption-key.yaml
@@ -2,8 +2,8 @@ title: Read the Local Master Key from a File
 ref: read-local-master-key-from-file
 level: 4
 content: |
-  First, read the file containing the local master key that you generated in 
-  the :ref:`Create a Master Key <fle-create-a-master-key>` section:
+  First, retrieve the contents of the local master key file that you generated
+  in the :ref:`Create a Master Key <fle-create-a-master-key>` section:
 
   .. tabs::
 
@@ -34,8 +34,9 @@ title: Specify KMS Provider Settings
 ref: specify-kms-provider-settings
 level: 4
 content: |
-  Next, specify the KMS provider settings. Set the provider name to ``local``
-  when using a local master key:
+  Next, specify the KMS provider settings. The client uses these settings to 
+  discover the master key. Set the provider name to ``local`` when using a local 
+  master key:
 
   .. tabs::
 
--- a/source/includes/steps-create-data-encryption-key.yaml
+++ b/source/includes/steps-create-data-encryption-key.yaml
@@ -1,9 +1,9 @@
-title: Read the local master key from a file
+title: Read the Local Master Key from a File
 ref: read-local-master-key-from-file
 level: 4
 content: |
-  First, read the file containing the local master key, generated in the
-  :ref:`Create a Master Key <fle-create-a-master-key>` section:
+  First, read the file containing the local master key that you generated in 
+  the :ref:`Create a Master Key <fle-create-a-master-key>` section:
 
   .. tabs::
 
@@ -30,7 +30,7 @@ content: |
 
 
 ---
-title: Specify KMS provider settings
+title: Specify KMS Provider Settings
 ref: specify-kms-provider-settings
 level: 4
 content: |
@@ -58,13 +58,13 @@ content: |
 
 
 ---
-title: Construct the client
-ref: construct-the-client
+title: Create a Data Encryption Key
+ref: create-a-data-encryption-key
 level: 4
 content: |
-  Create the client using the MongoDB connection string and key vault 
-  namespace. In this example, we specify the ``encryption`` database and the 
-  ``__keyVault`` collection:
+  Construct a client with the MongoDB connection string and key vault 
+  namespace configuration, and create a data encryption key. The key vault in 
+  this example uses the ``encryption`` database and ``__keyVault`` collection.
 
   .. tabs::
 
--- a/source/includes/steps-create-data-encryption-key.yaml
+++ b/source/includes/steps-create-data-encryption-key.yaml
@@ -1,0 +1,120 @@
+title: Read the local master key from a file
+ref: read-local-master-key-from-file
+level: 4
+content: |
+  First, read the file containing the local master key, generated in the
+  :ref:`Create a Master Key <fle-create-a-master-key>` section:
+
+  .. tabs::
+
+     tabs:
+
+       - id: use-cases-fle-guide-java
+         name: "Java"
+         content: |
+
+           .. code-block:: java
+              :emphasize-lines: 11
+
+              String path = "master-key.txt";
+
+              byte[] fileBytes = new byte[96];
+
+              try (FileInputStream fis = new FileInputStream(path)) {
+                fileBytes = fis.readAllBytes();
+              } catch (IOException e) {
+                e.printStackTrace();
+              }
+
+              final byte[] localMasterKey = Arrays.copyOf(fileBytes, 96);
+
+
+---
+title: Specify KMS provider settings
+ref: specify-kms-provider-settings
+level: 4
+content: |
+  Next, specify the KMS provider settings. Set the provider name to ``local``
+  when using a local master key:
+
+  .. tabs::
+
+     tabs:
+
+       - id: use-cases-fle-guide-java
+         name: "Java"
+         content: |
+
+           .. code-block:: java
+              :emphasize-lines: 4,5
+
+              String kmsProvider = "local";
+
+              Map<String, Map<String, Object>> kmsProviders = new HashMap<String, Map<String, Object>>() {{
+                put(kmsProvider, new HashMap<String, Object>() {{
+                  put("key", localMasterKey);
+                }});
+              }};
+
+
+---
+title: Construct the client
+ref: construct-the-client
+level: 4
+content: |
+  Create the client using the MongoDB connection string and key vault 
+  namespace. In this example, we specify the ``encryption`` database and the 
+  ``__keyVault`` collection:
+
+  .. tabs::
+
+     tabs:
+
+       - id: use-cases-fle-guide-java
+         name: "Java"
+         content: |
+
+           .. code-block:: java
+              :emphasize-lines: 8,9,12
+
+              String connectionString = "mongodb://localhost:27017";
+              String keyVaultNamespace = "encryption.__keyVault";
+
+              ClientEncryptionSettings clientEncryptionSettings = ClientEncryptionSettings.builder()
+                .keyVaultMongoClientSettings(MongoClientSettings.builder()
+                  .applyConnectionString(new ConnectionString(connectionString))
+                  .build())
+                .keyVaultNamespace(keyVaultNamespace)
+                .kmsProviders(kmsProviders)
+                .build();
+
+              ClientEncryption clientEncryption = ClientEncryptions.create(clientEncryptionSettings);
+
+
+  .. note::
+
+     Ensure that the client has `ReadWrite
+     <https://docs.mongodb.com/manual/reference/built-in-roles/#readWrite>`_
+     permissions on the specified key vault namespace.
+
+
+---
+title: Verify the data encryption key was created
+ref: verify-data-key-created
+level: 4
+content: |
+  Verify that the key was created in your MongoDB replica set in the specified
+  key vault namespace. The document should resemble the following:
+
+  .. code-block:: none
+  
+     {
+       "_id" : UUID("..."),
+       "keyMaterial" : BinData(0,"..."),
+       "creationDate" : ISODate("..."),
+       "updateDate" : ISODate("..."),
+       "status" : 0,
+       "masterKey" : {
+         "provider" : "local"
+       }
+     }
--- a/source/use-cases/sensitive-data-encryption.txt
+++ b/source/use-cases/sensitive-data-encryption.txt
@@ -131,20 +131,17 @@ In this step, we generate a local master key to expedite setup of our developmen
 B. Create a Data Encryption Key
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-In this section, we configure the MongoDB driver generate a data
+In this section, we configure the MongoDB driver to generate a data
 encryption key using the master key. The data encryption key, used to 
 automatically encrypt and decrypt fields in a document, is stored in a MongoDB 
 collection called the *key vault collection*.
 
-The following configuration values are required by the client to generate a new
-data encryption key:
+The client requires the following configuration values to generate a new data
+encryption key:
 
 * The local master key or AWS KMS master key access settings.
 * A MongoDB connection string that authenticates on a running server.
 * The key vault namespace (database and collection).
-
-In this guide, we use the ``encryption`` database and ``__keyvault`` 
-collection as our key vault collection.
 
 The MongoDB driver performs the following to generate the data encryption key:
 
@@ -153,7 +150,8 @@ The MongoDB driver performs the following to generate the data encryption key:
 #.  Returns the ``_id`` field of the inserted document as a `UUID <https://en.wikipedia.org/wiki/Universally_unique_identifier>`_.
 
 Follow the steps below to generate a single data encryption key from the
-local master key.
+local master key. You can also download the `complete code example on
+GitHub <https://raw.githubusercontent.com/mongodb/docs-assets/DOCSP-csfle-data-encryption-key/DataEncryptionKeyGenerator.java>`_.
 
 .. note::
 
--- a/source/use-cases/sensitive-data-encryption.txt
+++ b/source/use-cases/sensitive-data-encryption.txt
@@ -143,15 +143,14 @@ data encryption key:
 * A MongoDB connection string that authenticates on a running server.
 * The key vault namespace (database and collection).
 
-In this example, we use the ``encryption`` database and ``__keyvault`` 
+In this guide, we use the ``encryption`` database and ``__keyvault`` 
 collection as our key vault collection.
 
-Upon creating a data encryption key, the driver returns the ``_id`` field
-(as a `UUID <https://en.wikipedia.org/wiki/Universally_unique_identifier>`_)
-of a document that stores information about the key including:
+The MongoDB driver performs the following to generate the data encryption key:
 
-* The data encryption key, encrypted by the master key.
-* The key management service provider data.
+1.  Creates a data encryption key.
+#.  Inserts a new document into the key vault collection.
+#.  Returns the ``_id`` field of the inserted document as a `UUID <https://en.wikipedia.org/wiki/Universally_unique_identifier>`_.
 
 Follow the steps below to generate a single data encryption key from the
 local master key.
--- a/source/use-cases/sensitive-data-encryption.txt
+++ b/source/use-cases/sensitive-data-encryption.txt
@@ -153,12 +153,6 @@ Follow the steps below to generate a single data encryption key from the
 local master key. You can also download the `complete code example on
 GitHub <https://raw.githubusercontent.com/mongodb/docs-assets/DOCSP-csfle-data-encryption-key/DataEncryptionKeyGenerator.java>`_.
 
-.. note::
-
-  If using AWS KMS to provide the master key, refer to the :ref:`Convert
-  to a Remote Master Key <fle-convert-to-a-remote-master-key>` section for
-  the appropriate settings.
-
 .. include:: /includes/steps/create-data-encryption-key.rst
 
 .. _fle-define-a-json-schema:
--- a/source/use-cases/sensitive-data-encryption.txt
+++ b/source/use-cases/sensitive-data-encryption.txt
@@ -161,9 +161,6 @@ GitHub <https://raw.githubusercontent.com/mongodb/docs-assets/DOCSP-csfle-data-e
 
 .. include:: /includes/steps/create-data-encryption-key.rst
 
-`Download the source code <https://raw.githubusercontent.com/mongodb/docs-assets/DOCSP-csfle-data-encryption-key/DataEncryptionKeyGenerator.java>`_
-used to generate the data encryption key in this section.
-
 .. _fle-define-a-json-schema:
 
 C. Define a JSON Schema
--- a/source/use-cases/sensitive-data-encryption.txt
+++ b/source/use-cases/sensitive-data-encryption.txt
@@ -131,10 +131,9 @@ In this step, we generate a local master key to expedite setup of our developmen
 B. Create a Data Encryption Key
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-In this section, we configure the MongoDB driver to generate a data
-encryption key. The data encryption key, used to automatically encrypt and 
-decrypt fields in a document, is stored in a MongoDB collection called the
-*key vault collection*.
+In this section, we generate a data encryption key. The MongoDB driver stores 
+the key in a key vault collection where CSFLE-enabled clients can access the
+key for automatic encryption and decryption.
 
 The client requires the following configuration values to generate a new data
 encryption key:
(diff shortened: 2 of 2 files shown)


source/use-cases/sensitive-data-encryption.txt
```diff
@@ -124,6 +124,151 @@ In this step, we generate a local master key to expedite setup of our developmen
...
+* A MongoDB connection string that authenticates to a running server.
+* The key vault database and collection name.
+
+Upon creating a data encryption key, the driver returns the ``id`` field of
+a document that stores information about the encrypted key including:
+
+* The data encryption key, encrypted by the master key.
+* The key management service provider data.
+
+In this section, we generate a single data encryption key from the local
```
terakilobyte: "In this section" is used two paragraphs above. Suggest rewording, but not a blocker.

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -124,6 +124,151 @@ In this step, we generate a local master key to expedite setup of our developmen
...
+In this section, we use the MongoDB driver to create and store a data 
+encryption key, used by CSFLE to automatically encrypt and decrypt data on 
+document fields. The following configuration values are required by the client 
+to generate a new data encryption key:
+
+* The local master key or AWS KMS master key access settings.
+* A MongoDB connection string that authenticates to a running server.
+* The key vault database and collection name.
+
+Upon creating a data encryption key, the driver returns the ``id`` field of
```
terakilobyte: Shouldn't it be ``_id`` instead of ``id``? It also might be worth adding an admonition that the _id field of a key is *not* an ObjectID.

source/use-cases/sensitive-data-encryption.txt
ccho-mongodb: Thanks for identifying that, will reword.

source/use-cases/sensitive-data-encryption.txt
ccho-mongodb: I agree, I should use `_id`. Should I identify this as a `UUID` for the reader or is that too much detail?

terakilobyte: I think you should call it out for sure. It will be surprising that is a UUID instead of an ObjectID.

ccho-mongodb: Fixed, let me know if it looks ok now.

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -124,7 +124,153 @@ In this step, we generate a local master key to expedite setup of our developmen
           }
 
 
-B. Define a JSON Schema
+B. Create a Data Encryption Key
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+In this section, we use the MongoDB driver to create and store a data
+encryption key, used by CSFLE to automatically encrypt and decrypt data on
+document fields. The following configuration values are required by the client
```
nlarew: nit: IMO the comma before "used" makes it read like an appositive. Instead we could do two sentences:

In this section we use the MongoDB driver to create and store a data encryption key _in the keyvault collection_. Drivers that have access to the keyvault can use the encryption key to encrypt and decrypt fields in a document.

nlarew: nit: passive -> active

The driver requires the following configuration values to generate a new data encryption key:

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -124,7 +124,153 @@ In this step, we generate a local master key to expedite setup of our developmen
...
+to generate a new data encryption key:
+
+* The local master key or AWS KMS master key access settings.
+* A MongoDB connection string that authenticates on a running server.
+* The key vault database and collection name.
+
+Upon creating a data encryption key, the driver returns the ``_id`` field
+(as a `UUID <https://en.wikipedia.org/wiki/Universally_unique_identifier>`_)
+of a document that stores information about the key including:
```
nlarew: Are we certain that this is a UUID? I would imagine that it's a 12-byte ObjectId value unless this specifically does not use the default.

nlarew: I might interpret this as "the _id field stores information about the key" rather than the document. I think we can also explicitly say that the driver:

1.  creates the key
2. inserts a new document into the keyvault with info about the key
3. returns the _id of the inserted document

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -124,7 +124,153 @@ In this step, we generate a local master key to expedite setup of our developmen
...
+* The data encryption key, encrypted by the master key.
+* The key management service provider data.
+
+Follow the steps below to generate a single data encryption key from the
+local master key.
+
+.. note::
+
+  If using AWS KMS to provide the master key, refer to the "`F. Convert to
+  a Remote Master Key`_" section for the appropriate settings.
```
nlarew: nit: the "F" in the step title is a bit weird. Instead we could add an inline reference above the step heading and then use a `:ref:` here instead

**Inline reference above the step heading**
```
.. _fle-convert-to-a-remote-master-key:

F. Convert to a Remote Master Key
```

**Reference in this note**
```
If using AWS KMS to provide the master key, refer to the :ref:`Convert
to a Remote Master Key <fle-convert-to-a-remote-master-key>` section of
this guide for the appropriate settings.
```

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -124,7 +124,153 @@ In this step, we generate a local master key to expedite setup of our developmen
...
+
+First, read the file containing the local master key generated in the "`A.
+Create a Master Key`_" section:
```
nlarew: See my previous comment

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -124,7 +124,153 @@ In this step, we generate a local master key to expedite setup of our developmen
...
+
+          byte[] fileBytes = new byte[96];
+
+          try (FileInputStream fis = new FileInputStream(path)) {
+            fileBytes = fis.readAllBytes();
+          } catch (IOException e) {
+            e.printStackTrace();
+          }
+
+          final byte[] localMasterKey = Arrays.copyOf(fileBytes, 96);
```
nlarew: - The code here looks misaligned though it doesn't seem to cause a rendering issue. The first character of each line of code should be even with the "c" in `code-block`.
- It'd be useful to highlight the line where we actually have the master key in memory. e.g.

```rst
.. code-block:: java
   :emphasize-lines: 11
   
   // ... the rest of the code
```

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -124,7 +124,153 @@ In this step, we generate a local master key to expedite setup of our developmen
...
+
+
+Next, specify the KMS provider settings. Since we are using a local master
+key, the provider name is ``local``:
+
+.. tabs::
+
+  tabs:
+
+    - id: java-kms-settings-local
```
nlarew: All of our custom tab id values should be consistent, otherwise they won't stay in sync. (Eventually we should probably switch to pre-defined tabsets, but this is fine for now if we fix the id)

i.e. all the Java tabs should have the same ID, etc.

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -124,7 +124,153 @@ In this step, we generate a local master key to expedite setup of our developmen
...
+            fileBytes = fis.readAllBytes();
+          } catch (IOException e) {
+            e.printStackTrace();
+          }
+
+          final byte[] localMasterKey = Arrays.copyOf(fileBytes, 96);
+
+
+Next, specify the KMS provider settings. Since we are using a local master
+key, the provider name is ``local``:
```
nlarew: How do we specify the KMS provider settings? Looks like we make a hashmap in the code but why? What does it contain?

We should add a little bit more here to explain

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -124,7 +124,153 @@ In this step, we generate a local master key to expedite setup of our developmen
...
+
+          Map<String, Map<String, Object>> kmsProviders = new HashMap<String, Map<String, Object>>() {{
+             put(kmsProvider, new HashMap<String, Object>() {{
+                 put("key", localMasterKey);
+             }});
+          }};
+
+
+Create the client with the MongoDB connection string and key vault namespace
+(database and collection name):
```
nlarew: Let's give specific guidance about what the namespace should be. If we have a best practice, we should say it. Otherwise, we should explicitly say that the names are up to the developer.

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -124,7 +124,153 @@ In this step, we generate a local master key to expedite setup of our developmen
...
+    "creationDate" : ISODate("..."),
+    "updateDate" : ISODate("..."),
+    "status" : 0,
+    "masterKey" : {
+      "provider" : "local"
+    }
+  }
+
+`Click here <https://raw.githubusercontent.com/mongodb/docs-assets/DOCSP-csfle-data-encryption-key/DataEncryptionKeyGenerator.java>`_
+for the source code used to generate the data encryption key in this section.
```
nlarew: nit: [Don’t provide links from ambiguous phrases such as Click here or More information](https://docs.mongodb.com/meta/style-guide/style/links-and-cross-references/)

source/use-cases/sensitive-data-encryption.txt
nlarew: Also these "next..." phrases are really small relative to the code blocks they're sandwiched between. Given that this step is procedural, maybe we can convert to a stepfile?

source/use-cases/sensitive-data-encryption.txt
ccho-mongodb: Would it be ok if I just make a note to the user that is the arbitrary namespace we will be using for now? I discussed with Kenn and Ravind yesterday and the conclusion was that they would work on something more standardized during "beta cleanup". Conversation in #fle on Slack starts here:
https://mongodb.slack.com/archives/CDM9QEGUT/p1567530749015300

source/use-cases/sensitive-data-encryption.txt
ccho-mongodb: Thanks for pointing that out, will fix.

source/use-cases/sensitive-data-encryption.txt
ccho-mongodb: Oops. Up to this point, I thought they should be unique. Will need to fix across multiple PRs. I'll create a JIRA ticket to do cleanup on the guide once all the sections have been merged and add this as an item to address. For now, I will use `use-cases-fle-guide-java` (use-cases section, fle guide document, language)

source/use-cases/sensitive-data-encryption.txt
ccho-mongodb: > Why.....
I'd imagine this has to do with the crypto library and someone's decision not to convert it to an ObjectId. It does not seem that important to the user, however, since it would only be defined in the JSON Schema.

Per the second comment, I will try to reformat as an ordered list.

source/use-cases/sensitive-data-encryption.txt
ccho-mongodb: Thanks, I was looking for that directive, but couldn't find it for some reason. Will fix.

source/use-cases/sensitive-data-encryption.txt
ccho-mongodb: Thanks, will fix.

source/use-cases/sensitive-data-encryption.txt
ccho-mongodb: Thanks, didn't know about the emphasize-lines directive. Will fix.

source/use-cases/sensitive-data-encryption.txt
ccho-mongodb: > Looks like we make a hashmap in the code but why? What does it contain?
>
> We should add a little bit more here to explain

Can you explain what you're looking for? Each driver has its own way of specifying settings whether hashmap, json, associative array, etc. Did you want to have explicit information about how drivers specify their configurations? Would that belong in the scope of this document?

source/use-cases/sensitive-data-encryption.txt
ccho-mongodb: Will fix.

source/use-cases/sensitive-data-encryption.txt
ccho-mongodb: Agree, it's awkward. I see that you want to emphasize/include discussion of the key vault collection. Since it's the first two sentences, would your goal be met if key vault collection was only mentioned in the second one?

source/use-cases/sensitive-data-encryption.txt
ccho-mongodb: > Given that this step is procedural, maybe we can convert to a stepfile?

Good idea, will move to a step file.

ccho-mongodb: Question about tabs in a step file:
Should the id use the same naming scheme as the document? If so, then they are not reusable in different documents, if there is ever that need. Should we make note of that somewhere?

ccho-mongodb: Another question -- if I have a `:ref:` link in the step file, will it resolve to the anchors set in the document that includes the step file?

source/use-cases/sensitive-data-encryption.txt
nlarew: I think that's probably fine. The main reason I had it in both is to ensure that readers understand the keyvault is just a collection.

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -124,7 +126,49 @@ In this step, we generate a local master key to expedite setup of our developmen
           }
 
 
-B. Define a JSON Schema
+.. _fle-create-a-data-encryption-key:
+
+B. Create a Data Encryption Key
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+In this section, we configure the MongoDB driver generate a data
```
nlarew: configure the MongoDB driver _**and/to**_ generate a data key...

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -124,7 +126,49 @@ In this step, we generate a local master key to expedite setup of our developmen
...
+encryption key using the master key. The data encryption key, used to 
+automatically encrypt and decrypt fields in a document, is stored in a MongoDB 
+collection called the *key vault collection*.
+
+The following configuration values are required by the client to generate a new
+data encryption key:
```
nlarew: passive -> active

the client requires the following configuration values...

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -124,7 +126,49 @@ In this step, we generate a local master key to expedite setup of our developmen
...
+* A MongoDB connection string that authenticates on a running server.
+* The key vault namespace (database and collection).
+
+In this guide, we use the ``encryption`` database and ``__keyvault`` 
+collection as our key vault collection.
+
+The MongoDB driver performs the following to generate the data encryption key:
+
+1.  Creates a data encryption key.
+#.  Inserts a new document into the key vault collection.
```
nlarew: What's in the document? Looks like we give an example document in the final substep - maybe we can elaborate on what's in the document there?

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -124,7 +126,49 @@ In this step, we generate a local master key to expedite setup of our developmen
...
+.. note::
+
+  If using AWS KMS to provide the master key, refer to the :ref:`Convert
+  to a Remote Master Key <fle-convert-to-a-remote-master-key>` section for
+  the appropriate settings.
+
+.. include:: /includes/steps/create-data-encryption-key.rst
+
+`Download the source code <https://raw.githubusercontent.com/mongodb/docs-assets/DOCSP-csfle-data-encryption-key/DataEncryptionKeyGenerator.java>`_
+used to generate the data encryption key in this section.
```
nlarew: Seems like this might get lost here at the bottom of the step. Maybe we could combine with the "follow the steps" sentence above the stepfile?

Follow the steps below to generate a single data encryption key from the local master key. You can also find the complete code example on GitHub.

source/includes/steps-create-data-encryption-key.yaml
```diff
@@ -0,0 +1,120 @@
+title: Read the local master key from a file
```
nlarew: nit: Let's use title case for step titles.

See [Enforce a Document Schema](https://docs.mongodb.com/stitch/mongodb/enforce-a-document-schema/) for an example

source/includes/steps-create-data-encryption-key.yaml
```diff
@@ -0,0 +1,120 @@
...
+ref: read-local-master-key-from-file
+level: 4
+content: |
+  First, read the file containing the local master key, generated in the
+  :ref:`Create a Master Key <fle-create-a-master-key>` section:
```
nlarew: nit: the local master key _that you_ generated

source/includes/steps-create-data-encryption-key.yaml
```diff
@@ -0,0 +1,120 @@
...
+              final byte[] localMasterKey = Arrays.copyOf(fileBytes, 96);
+
+
+---
+title: Specify KMS provider settings
+ref: specify-kms-provider-settings
+level: 4
+content: |
+  Next, specify the KMS provider settings. Set the provider name to ``local``
+  when using a local master key:
```
nlarew: I think we could have a little more detail on the what the KMS provider settings actually are / what they configure.

source/includes/steps-create-data-encryption-key.yaml
```diff
@@ -0,0 +1,120 @@
...
+     permissions on the specified key vault namespace.
+
+
+---
+title: Verify the data encryption key was created
+ref: verify-data-key-created
+level: 4
+content: |
+  Verify that the key was created in your MongoDB replica set in the specified
+  key vault namespace. The document should resemble the following:
```
nlarew: It'd be nice to have a quick explanation of what the document actually is/contains/represents beyond just the example

source/includes/steps-create-data-encryption-key.yaml
```diff
@@ -0,0 +1,120 @@
...
+
+              Map<String, Map<String, Object>> kmsProviders = new HashMap<String, Map<String, Object>>() {{
+                put(kmsProvider, new HashMap<String, Object>() {{
+                  put("key", localMasterKey);
+                }});
+              }};
+
+
+---
+title: Construct the client
```
nlarew: The end state of this step isn't just a constructed client, it's actually a created key. Maybe the title here should be **Create the Data Key** and then in the text below we explain that the client creates and inserts the key.

source/use-cases/sensitive-data-encryption.txt
nlarew: Thoughts on moving this process into the content of step three? It's most relevant there IMO.

source/use-cases/sensitive-data-encryption.txt
ccho-mongodb: Oops, good catch, will fix.

source/use-cases/sensitive-data-encryption.txt
ccho-mongodb: Thanks, will fix.

source/includes/steps-create-data-encryption-key.yaml
ccho-mongodb: Will do, thanks for the example.

source/use-cases/sensitive-data-encryption.txt
ccho-mongodb: Sounds good, will update.

source/includes/steps-create-data-encryption-key.yaml
ccho-mongodb: Good point, will fix.

source/includes/steps-create-data-encryption-key.yaml
ccho-mongodb: Thanks, that sounds better, will fix. Just for my own understanding, is "generated in..." an example of a non-restrictive clause? And we prefer to use which/that over commas? Or am I completely wrong?

source/includes/steps-create-data-encryption-key.yaml
ccho-mongodb: I removed that description from the beginning of the section because I did not think it provided very useful information. Which of the fields listed here https://github.com/mongodb/docs-ecosystem/pull/535/files#diff-54590a1d9915ddb3aea9c0163a7a6c8cR111 would you consider important? Though I did not directly refer to the names, I described the document contains _id, masterKey, and keyMaterial fields.
Should I also mention keyAltNames (which I have not attempted to generate)?

source/includes/steps-create-data-encryption-key.yaml
ccho-mongodb: Whoops, lost my comment here (created a review instead of single comment, and then subsequently discarded the review). I prefer your version, but the question was just about whether "generated in the..." is an example of a non-restrictive clause and that we prefer that/which over commas, or whether it is not even a non-restrictive clause.

source/use-cases/sensitive-data-encryption.txt
ccho-mongodb: I'm fine with moving it to step three, I was just trying to centralize "pre-" and "post-conditions" in the initial section. Would it also make sense to then move "The client requires the following..." on line 139 to steps 2 and 3 when they are used in the code examples?

source/includes/steps-create-data-encryption-key.yaml
ccho-mongodb: I was planning to mention the AWS KMS-specific ones in the final section of the use-case guide, but are there other settings and are they important to mention?

source/includes/steps-create-data-encryption-key.yaml
nlarew: I don't think we necessarily need to mention specific fields. Maybe just "includes the key (or KMS info) and metadata that describes when the key was created and last updated".

ccho-mongodb: Great, let me know if this makes sense:

https://github.com/mongodb/docs-ecosystem/pull/535/commits/58163994ae3c0dc20ee19ba73fbfe17a960b90db

source/includes/steps-create-data-encryption-key.yaml
```diff
@@ -0,0 +1,127 @@
...
+  .. tabs::
+
+     tabs:
+
+       - id: use-cases-fle-guide-java
+         name: "Java"
+         content: |
+
+           .. code-block:: java
+              :emphasize-lines: 8,9,12
```
ccho-mongodb: Do you think I might be overusing the `emphasize-lines` directive?

source/includes/steps-create-data-encryption-key.yaml
nlarew: AFAIK non-restrictive clause just means that the clause does not increase specificity. I think this might qualify under that definition but I'm not certain.

I think the main way we use it is in appositives, but I don't know if we use verb phrase appositives anywhere. I personally don't prefer this style because it feels like passive voice, i.e.:

"read the file containing the local master key (that was) generated in the Create a Master Key section (by whom?)."

source/includes/steps-create-data-encryption-key.yaml
nlarew: Don't think there's any settings we don't have in this or the final step. I'm thinking more of "the client uses this settings map to access the master key"

source/includes/steps-create-data-encryption-key.yaml
nlarew: Nope! Only 3 lines is fine and the highlighted lines here seem relevant.

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -124,7 +126,44 @@ In this step, we generate a local master key to expedite setup of our developmen
...
+#.  Inserts a new document into the key vault collection.
+#.  Returns the ``_id`` field of the inserted document as a `UUID <https://en.wikipedia.org/wiki/Universally_unique_identifier>`_.
+
+Follow the steps below to generate a single data encryption key from the
+local master key. You can also download the `complete code example on
+GitHub <https://raw.githubusercontent.com/mongodb/docs-assets/DOCSP-csfle-data-encryption-key/DataEncryptionKeyGenerator.java>`_.
+
+.. note::
+
+  If using AWS KMS to provide the master key, refer to the :ref:`Convert
```
ccho-mongodb: Remove this

source/use-cases/sensitive-data-encryption.txt
ccho-mongodb: Delete this list

source/includes/steps-create-data-encryption-key.yaml
ccho-mongodb: Thanks for the detailed explanation!

source/includes/steps-create-data-encryption-key.yaml
ccho-mongodb: Makes sense, will add detail.

PR #544, opened 2019-09-23
fixes to CSFLE use-case guide

ccho-mongodb:
JIRA: 
https://jira.mongodb.org/browse/DOCSP-6971

STAGE:
https://docs-mongodbcom-staging.corp.mongodb.com/ecosystem/docsworker/DOCSP-6971-csfle-improvements/use-cases/sensitive-data-encryption.html

### DIFF
--- a/source/includes/steps-fle-configure-the-mongodb-client.yaml
+++ b/source/includes/steps-fle-configure-the-mongodb-client.yaml
@@ -5,10 +5,10 @@ content: |
   encrypt and decrypt fields. MedcoMD uses the collection
   ``encryption.__keyVault`` as the key vault.
 
-  .. tabs::
-     
-     .. tab:: Java
-        :tabid: use-cases-fle-guide-java
+  .. tabs-drivers::
+   
+     .. tab::
+        :tabid: java-sync
         
         .. code-block:: java
            
@@ -22,10 +22,10 @@ content: |
   master key, so they use the ``local`` KMS provider and specify the key
   inline.
 
-  .. tabs::
-     
-     .. tab:: Java
-        :tabid: use-cases-fle-guide-java
+  .. tabs-drivers::
+   
+     .. tab::
+        :tabid: java-sync
         
         .. code-block:: java
            :emphasize-lines: 2,3
@@ -43,10 +43,10 @@ content: |
   collection to which it applies. To assign the schema, they map it to
   the ``medicalRecords.patients`` collection namespace:
 
-  .. tabs::
-     
-     .. tab:: Java
-        :tabid: use-cases-fle-guide-java
+  .. tabs-drivers::
+   
+     .. tab::
+        :tabid: java-sync
         
         .. code-block:: java
            :emphasize-lines: 2
@@ -63,10 +63,10 @@ content: |
   ``mongocryptd`` process. To configure the client to find the binary,
   MedcoMD specifies the binary's path as an extra option:
   
-  .. tabs::
-     
-     .. tab:: Java
-        :tabid: use-cases-fle-guide-java
+  .. tabs-drivers::
+   
+     .. tab::
+        :tabid: java-sync
         
         .. code-block:: java
            :emphasize-lines: 2
@@ -96,10 +96,10 @@ content: |
   MongoDB client object with the additional automatic encryption
   settings:
 
-  .. tabs::
-     
-     .. tab:: Java
-        :tabid: use-cases-fle-guide-java
+  .. tabs-drivers::
+   
+     .. tab::
+        :tabid: java-sync
         
         .. code-block:: java
            :emphasize-lines: 3-8
--- a/source/includes/steps-fle-configure-the-mongodb-client.yaml
+++ b/source/includes/steps-fle-configure-the-mongodb-client.yaml
@@ -168,7 +168,7 @@ content: |
         :tabid: nodejs
 
         .. code-block:: javascript
-
+           :emphasize-lines: 5-9
            const secureClient = new MongoClient(connectionString, {
                useNewUrlParser: true,
                useUnifiedTopology: true,
--- a/source/includes/steps-fle-configure-the-mongodb-client.yaml
+++ b/source/includes/steps-fle-configure-the-mongodb-client.yaml
@@ -78,7 +78,7 @@ content: |
            :emphasize-lines: 2
 
            const patientSchema = {
-             ["medicalRecords.patients"]: jsonSchema,
+             "medicalRecords.patients": jsonSchema,
            }
 ---
 title: Specify the Location of the Encryption Binary
@@ -121,7 +121,7 @@ content: |
            :emphasize-lines: 2
 
            const extraOptions = {
-             mongocryptdPath: "/usr/local/bin/mongocryptd",
+             mongocryptdSpawnPath: "/usr/local/bin/mongocryptd",
            }
 
         .. admonition:: Encryption Binary Daemon
--- a/source/includes/steps-fle-configure-the-mongodb-client.yaml
+++ b/source/includes/steps-fle-configure-the-mongodb-client.yaml
@@ -45,13 +45,13 @@ content: |
         :tabid: nodejs
 
         .. code-block:: javascript
+           :emphasize-lines: 2,3
 
            const kmsProviders = {
              local: {
-               key: localMasterKey
+               key: localMasterKey,
              }
            }
-
 ---
 title: Map the JSON Schema to the Patients Collection
 ref: map-the-json-schema-to-the-patients-collection
@@ -75,10 +75,10 @@ content: |
         :tabid: nodejs
 
         .. code-block:: javascript
+           :emphasize-lines: 2
 
-           const dataNamespace = "medicalRecords.patients";
            const patientSchema = {
-             [dataNamespace]: jsonSchema
+             ["medicalRecords.patients"]: jsonSchema,
            }
 ---
 title: Specify the Location of the Encryption Binary
@@ -121,7 +121,7 @@ content: |
            :emphasize-lines: 2
 
            const extraOptions = {
-            mongocryptdPath: "/usr/local/bin/mongocryptd";
+             mongocryptdPath: "/usr/local/bin/mongocryptd",
            }
 
         .. admonition:: Encryption Binary Daemon
@@ -135,7 +135,7 @@ content: |
               :emphasize-lines: 2
 
                const extraOptions = {
-                 mongocryptdBypassSpawn: true;
+                 mongocryptdBypassSpawn: true,
                }
 ---
 title: Create the MongoClient
@@ -169,6 +169,7 @@ content: |
 
         .. code-block:: javascript
            :emphasize-lines: 5-9
+
            const secureClient = new MongoClient(connectionString, {
                useNewUrlParser: true,
                useUnifiedTopology: true,
@@ -177,7 +178,7 @@ content: |
                  keyVaultNamespace,
                  kmsProviders,
                  schemaMap: patientSchema,
-                 extraOptions: extraOptions
+                 extraOptions: extraOptions,
                }
            });
 ...
--- a/source/includes/steps-fle-create-data-encryption-key.yaml
+++ b/source/includes/steps-fle-create-data-encryption-key.yaml
@@ -120,7 +120,7 @@ content: |
         :tabid: nodejs
 
         .. code-block:: javascript
-           :emphasize-lines: 10,11,13
+           :emphasize-lines: 10,11,13,16
 
            const base64 = require("uuid-base64");
  
--- a/source/includes/steps-fle-create-data-encryption-key.yaml
+++ b/source/includes/steps-fle-create-data-encryption-key.yaml
@@ -6,23 +6,23 @@ content: |
   in the :ref:`Create a Master Key <fle-create-a-master-key>` section:
 
   .. tabs-drivers::
-   
+
      .. tab::
         :tabid: java-sync
 
         .. code-block:: java
            :emphasize-lines: 11
-        
+
            String path = "master-key.txt";
-        
+
            byte[] fileBytes = new byte[96];
-        
+
            try (FileInputStream fis = new FileInputStream(path)) {
              fileBytes = fis.readAllBytes();
            } catch (IOException e) {
              e.printStackTrace();
            }
-        
+
            final byte[] localMasterKey = Arrays.copyOf(fileBytes, 96);
 ---
 title: Specify KMS Provider Settings
@@ -34,15 +34,20 @@ content: |
   master key:
 
   .. tabs-drivers::
-     
+
      .. tab::
         :tabid: java-sync
 
+        The KMS provider settings are stored in a Map in order to use the
+        :java-docs-latest:`kmsProviders helper
+        method <javadoc/com/mongodb/ClientEncryptionSettings.Builder.html#kmsProviders(java.util.Map)>
+        for the ClientEncryptionSettings Builder.
+
         .. code-block:: java
            :emphasize-lines: 4,5
-        
+
            String kmsProvider = "local";
-        
+
            Map<String, Map<String, Object>> kmsProviders = new HashMap<String, Map<String, Object>>() {{
              put(kmsProvider, new HashMap<String, Object>() {{
                put("key", localMasterKey);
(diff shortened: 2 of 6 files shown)


source/includes/steps-fle-create-data-encryption-key.yaml
```diff
@@ -1,4 +1,4 @@
-title: Read the Local Master Key from a File
+title: Read the Locally-managed Master Key from a File
```
nlarew: nit: Locally-Managed

source/includes/steps-fle-create-data-encryption-key.yaml
```diff
@@ -88,8 +88,37 @@ title: Verify the data encryption key was created
 ref: verify-data-key-created
```
nlarew: nit: title case on the step title (Missed this in the unification pass)

**Verify that the Data Encryption Key was Created**

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -300,6 +289,15 @@ information:
...
   encrypted fields)
 
+.. note::
+   
+   While a MongoDB client can use JSON Schema for validating types and values
+   of a document, the encrypted client can only process instructions for
+   automatic client-side encryption and decryption. In this example, we define
+   a JSON Schema for the encrypted client only. For information on type 
+   and value validation with JSON Schema, see 
+   :manual:`Schema Validation </core/schema-validation/>`.
```
nlarew: AFAIK clients cannot use JSON Schema for validation at all. It's a feature that exclusively lives on the server.

```rst
.. admonition:: JSON Schema Validation
   :important:
   
   MongoDB drivers only use JSON schema to configure automatic
   client-side field level encryption. If you specify a document
   validation keyword in your CSFLE JSON Schema then the client will
   throw an error (when?).
   
   For information on how to use JSON Schema on the server to validate
   data, see :manual:`Schema Validation </core/schema-validation/>`.
```

- When does the client error (on creation? on query? etc.)

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -649,9 +647,9 @@ researching the available options, MedcoMD determined that MongoDB
 Client-Side Field Level Encryption satisfies their requirements and
 decided to implement it in their application. To implement CSFLE they:
 
-**1. Created a Local Master Encryption Key**
+**1. Created a Locally-managed Master Encryption Key**
```
nlarew: nit: Locally-Managed

source/includes/steps-fle-create-data-encryption-key.yaml
```diff
@@ -88,8 +88,37 @@ title: Verify the data encryption key was created
...
-  key vault collection. This document contains the following data:
+  First, verify that the command to create the data encryption key returns the
+  ID of the newly-created key:
+
+  .. tabs-drivers::
+
+     .. tab::
+        :tabid: java-sync
+
+        The ``createDataKey()`` method returns a :java-docs-latest:`BsonBinary
```
nlarew: It seems like we should mention `createDataKey()` in the previous step (**B.3 Create a Data Encryption Key**) but we don't currently.

- We should add `createDataKey()` to the end of the previous step's code block
- Seems like this new block would make sense as part of the previous step as well. Thoughts?

source/includes/steps-fle-create-data-encryption-key.yaml
ccho-mongodb: Thanks, I was wondering whether hyphenated words would be uppercase in title case.

source/includes/steps-fle-create-data-encryption-key.yaml
ccho-mongodb: Thanks, will fix.

source/includes/steps-fle-create-data-encryption-key.yaml
ccho-mongodb: Thanks, I was wondering whether hyphenated words would be uppercase in title case.

source/use-cases/sensitive-data-encryption.txt
ccho-mongodb: Thanks. Do you think we may need a more detailed explanation of "document validation keyword" for users that may not be familiar with the server-based validation? Will investigate the Java driver to see when the error is thrown and check if it exists in the driver specification.

source/includes/steps-fle-create-data-encryption-key.yaml
ccho-mongodb: Makes sense to add the call `createDataKey`, looks like we just missed that part. I think the "verify" step could stay separate since it already has 3 sub-sections: print the returned key, what the output should look like, and what the key vault collection should now contain.
We should probably show the user the commands to check the `__keyVault` collection as well. What do you think?

source/use-cases/sensitive-data-encryption.txt
ccho-mongodb: According to the driver specification it should error, so the Java driver is likely not up to spec, or the spec is outdated:

From: https://github.com/mongodb/specifications/blob/master/source/client-side-encryption/client-side-encryption.rst#schemamap

> Schemas supplied in the schemaMap only apply to configuring automatic encryption for client side encryption. Other validation rules in the JSON schema will not be enforced by the driver and will result in an error.

nlarew: Hmm the spec is also pretty unclear ("will result in an error _when????_") so maybe we just leave it out for now. Up to you - thanks for checking!

source/includes/steps-fle-create-data-encryption-key.yaml
nlarew: 👍 Fine with separate if you think it works!

RE: the keyvault commands, would that just be a `mongoClient.db("encryption").collection("__keyVault").find()` or are you thinking of something else? I don't think we _need_ to have it but if you think of a nice way to slot it in go for it.

source/use-cases/sensitive-data-encryption.txt
rkumar-mongo: My reading is that specifying validation rules in the JSON schema should *always* error. This is also my understanding of how mongocryptd is meant to work. 

IMO we should document the truth by the spec document and open a JAVA ticket w/ reproduction steps and requesting that they fix and always return an error. I might be wrong but I'm almost certain that mongocryptd throws an error if any document validation keywords are included in the schema.

source/includes/steps-fle-create-data-encryption-key.yaml
```diff
@@ -77,19 +77,48 @@ content: |
...
+        <javadoc/org/bson/BsonBinary.html>` object from which we can extract
+        the UUID and Base64 representations of the key ID.
```
rkumar-mongo: I think we should let the user know here why this matters.

UUID is how the shell presents data key IDs and represents the container of the data.

the base64 representation is required for constructing the extended JSON Binary Data representation, and is required for automatic json schema.

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -300,6 +289,17 @@ information:
 - The BSON Type of each field (only required for deterministically
   encrypted fields)
 
+.. admonition::JSON Schema Validation
+   :important:
+   
+   MongoDB drivers only use JSON Schema to configure automatic client-side
```
rkumar-mongo: The drivers use the JSON Schema *syntax* to configure automatic client-side field level encryption and support *only* field level encryption-specific keywords.

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -300,6 +289,17 @@ information:
...
+   field level encryption. If you specify a document validation instance 
+   without an ``encrypt`` property, the client will either ignore it or throw
```
rkumar-mongo: Maybe link to https://docs.mongodb.com/manual/reference/security-client-side-automatic-json-schema/ for supported keywords and explicitly state that any validation-specific keywords result in errors?

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -300,6 +289,17 @@ information:
...
+   an error.
+
+   For information on how to use JSON Schema on the server to validate
```
rkumar-mongo: I think this may confuse users. Especially since they *can* specify encryption rules for server-side document validation. I would opt to remove this entirely, or re-frame as.

"Users *may* set server-side document validation using encryption keywords to enforce client-side field level encryption."

That said ,I've not yet completed the docs for that feature so it would be a dead link for now.

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -300,6 +289,17 @@ information:
 - The BSON Type of each field (only required for deterministically
   encrypted fields)
 
+.. admonition::JSON Schema Validation
```
rkumar-mongo: "Automatic Client-Side Field Level Encryption JSON Schema Does Not Support Document Validation"

source/use-cases/sensitive-data-encryption.txt
ccho-mongodb: Thanks, this detail really helps clarify the point. Will update.

source/use-cases/sensitive-data-encryption.txt
ccho-mongodb: Thanks for the link. I'll incorporate this into this admonition.

source/use-cases/sensitive-data-encryption.txt
ccho-mongodb: Ok, I'll create a JIRA ticket to add the server-side document validation statement once we have a link. For now, i will remove mention of general schema validation.

ccho-mongodb: https://jira.mongodb.org/browse/DOCSP-7006

source/includes/steps-fle-create-data-encryption-key.yaml
ccho-mongodb: Good idea. I will add information on when each representation is used.

source/use-cases/sensitive-data-encryption.txt
ccho-mongodb: I opened an issue for the Java driver yesterday:
https://jira.mongodb.org/browse/JAVA-3443

Hopefully, this is enough info. If not, I can work on a gist that steps them through end-to-end.

I think Moe will get the opportunity to test the behavior with Python and Node drivers this week, so it is wording we will revisit based on those results.

Side note: just noticed another field that was getting added into the schema that doesn't include any encrypt property that I need to remove.

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -188,8 +187,8 @@ File System Permissions
 A. Create a Master Key
 ~~~~~~~~~~~~~~~~~~~~~~
 
-MongoDB `Client-Side Field Level Encryption (CSFLE)
-<https://docs.mongodb.com/manual/core/security-client-side-encryption/>`_
+MongoDB :manual:`Client-Side Field Level Encryption (CSFLE)
```
rkumar-mongo: Bah. stupid github. Line 149, extra apostrophe

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -257,25 +246,25 @@ access the key for automatic encryption and decryption.
...
    .. tab::
       :tabid: java-sync
        
-      You can also download the `complete code example on GitHub
-      <https://raw.githubusercontent.com/mongodb/docs-assets/DOCSP-csfle-data-encryption-key/DataEncryptionKeyGenerator.java>`_.
+      You can also download the `complete code example
+      <https://raw.githubusercontent.com/mongodb/docs-assets/DOCSP-json-schema-helper-and-json/DataEncryptionKeyGenerator.java>`_.
 
-.. include:: /includes/steps/create-data-encryption-key.rst
+.. include:: /includes/steps/fle-create-data-encryption-key.rst
```
rkumar-mongo: A few thoughts:

- For step 2, can we point to https://mongodb.github.io/mongo-java-driver/3.11/javadoc/com/mongodb/AutoEncryptionSettings.Builder.html#kmsProviders(java.util.Map) so users have a reference for why we're doing a map of maps of maps?
- For step 3, note that the base64 encoded key ID is required for automatic field level encryption
- For step 4, we're using the built in Document to string. I think Jeff Yemin provided a better example that returns the proper JSON - can we use that instead?

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -465,7 +464,7 @@ full `JSON Schema for the Medco Medical Management System
       :tabid: java-sync
        
       View the `helper code in Java
-      <https://gist.github.com/ccho-mongodb/088176b1bed3b9e54cdc0c2c3c537d1b>`_.
+      <https://raw.githubusercontent.com/mongodb/docs-assets/DOCSP-json-schema-helper-and-json/JSONSchemaCreator.java>`_.
```
rkumar-mongo: Line 352, we have not told users that keyID must use the extended json v2 format for a $binary key. I think most users will paste in the base64 key and get confused. 

Strongly recommend providing a full skeleton of the $binary format as you have in the [example schema](https://raw.githubusercontent.com/mongodb/docs-assets/DOCSP-json-schema-helper-and-json/MedcoMDSchema.json)

source/use-cases/sensitive-data-encryption.txt
ccho-mongodb: Good catch, will fix.

source/use-cases/sensitive-data-encryption.txt
ccho-mongodb: Step 2: sure, I'll add a link. Other than it being the documented interface, is there any other info that explains why it's a map of maps?

Step 3: sure; the code snippet also contains a comment and variable name that reflects that they need to use the base64 key id, but I can see how it would be helpful to mention earlier. Will update.

Step 4: I'm on the fence about this proposed addition. My understanding of the extended json section you're asking for would explain both extended json and each parameter. I'd like to separate the document from function or feature documentation and instead maintain its focus as a tutorial. I can definitely make a note and include a link to mention the ejson representation. Let me know what you think @rkumar-mongo

source/use-cases/sensitive-data-encryption.txt
ccho-mongodb: The code snippet we provided on this line contains a helper `buildEncryptMetadata` that constructs the contents of the encryptMetadata field in the JSON Schema for them which allows the reader to paste the base64 format key id in the `<paste_your_key_id_here>` line.

Is the request to add a note that mentions that the helper code takes care of this for you? @rkumar-mongo

source/includes/steps-fle-create-data-encryption-key.yaml
```diff
@@ -0,0 +1,162 @@
...
+  master key:
+
+  .. tabs-drivers::
+
+     .. tab::
+        :tabid: java-sync
+
+        The KMS provider settings are stored in a Map in order to use the
+        :java-docs-latest:`kmsProviders helper
+        method <javadoc/com/mongodb/ClientEncryptionSettings.Builder.html#kmsProviders(java.util.Map)>
```
rkumar-mongo: Missing terminating ` on this line reslts broken link.

ccho-mongodb: Good catch, thanks.

source/includes/steps-fle-configure-the-mongodb-client.yaml
```diff
@@ -61,24 +61,24 @@ content: |
...
-        
+
         .. code-block:: javascript
-        
-           const dataNamespace = "medicalRecords.patients";
+           :emphasize-lines: 2
+
            const patientSchema = {
-             [dataNamespace]: jsonSchema
+             ["medicalRecords.patients"]: jsonSchema,
```
mongomoe: If we're removing the use of the variable 'dataNamespace' there's no need for the array brackets around  "medicalRecords.patients". The array brackets around a keyname are only useful if you want the keyname comes from a bracket. Please change this to:

   ```
        const patientSchema = {
"medicalRecords.patients": jsonSchema,
}
```

File: source/includes/steps-fle-configure-the-mongodb-client.yaml [hunk 16]
```diff
@@ -87,55 +87,55 @@ content: |
...
         :tabid: nodejs
-        
+
         .. code-block:: javascript
            :emphasize-lines: 2
-           
+
            const extraOptions = {
-            mongocryptdPath: "/usr/local/bin/mongocryptd";
+             mongocryptdPath: "/usr/local/bin/mongocryptd",
```
mongomoe: I'm sorry this one is probably my error, but the extraOptions takes in a key called "mongocryptdSpawnPath" not "mongocryptdPath". See: https://www.npmjs.com/package/mongodb-client-encryption#AutoEncrypter..AutoEncryptionExtraOptions

ccho-mongodb: Will update

source/includes/steps-fle-configure-the-mongodb-client.yaml
ccho-mongodb: Will fix

PR #545, opened 2019-09-23
Kafka connector initial scaffold

ccho-mongodb:
JIRA: https://jira.mongodb.org/browse/DOCSP-6631

STAGE: https://docs-mongodbcom-staging.corp.mongodb.com/ecosystem/docsworker/DOCSP-6630-kafka-connector/tools.html

### DIFF
--- a/source/connectors/kafka.txt
+++ b/source/connectors/kafka.txt
@@ -1,0 +1,34 @@
+.. _kafka:
+
+.. _kafka-connector-landing:
+
+=======================
+MongoDB Kafka Connector
+=======================
+
+.. default-domain:: mongodb
+
+.. contents:: On this page
+   :local:
+   :backlinks: none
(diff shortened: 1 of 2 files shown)


ccho-mongodb:
>  it's gonna turn out to be a somewhat long chunky page
It will definitely be a long page. Is there a common pattern to organize it into a more hierarchical structure? Now that I searched through docs-ecosystem a little more thoroughly, I see "Python Drivers" is split into child pages, so I can emulate that rather than the model we used for CSFLE.

> It might be more effective to name the guides using action phrases, e.g.
Good idea, I will update them in the context of the aforementioned child docs.

> Migration Guide might be a bit more descriptive. Migrate from what to what?
`hpgrahsl/kafka-connect-mongodb` => `mongodb/mongo-kafka`

PR #552, opened 2019-10-07
(DOCS-12995): add admonition for version in dev

terakilobyte:
[Staging](https://docs-mongodbcom-staging.corp.mongodb.com/ecosystem/docsworker/DOCS-12995/drivers/cxx.html#mongodb-compatibility)

This adds an admonition to the cxx page and marks the 3.5 version as in development.

### DIFF
--- a/source/drivers/cxx.txt
+++ b/source/drivers/cxx.txt
@@ -65,6 +65,11 @@ Compatibility
 MongoDB Compatibility
 ~~~~~~~~~~~~~~~~~~~~~
 
+.. admonition:: Development
+   :class: important
+
+   `*` denotes a version in development.
+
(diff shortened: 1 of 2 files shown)


source/drivers/cxx.txt
```diff
@@ -65,6 +65,11 @@ Compatibility
 MongoDB Compatibility
 ~~~~~~~~~~~~~~~~~~~~~
 
+.. admonition:: Development
+   :class: important
+
+   `*` denotes a version in development.
```
ccho-mongodb: Quick question:
Is this version available as a snapshot? I was planning to investigate that as part of the ticket prior to reassignment.

terakilobyte: I don't have an answer for you at the moment. I'll look into it.

PR #554, opened 2019-10-08
Python CSFLE implementation

mongomoe:
jira: https://jira.mongodb.org/browse/DOCSP-6987
stage: https://docs-mongodbcom-staging.corp.mongodb.com/ecosystem/docsworker/DOCSP-6987-Python-CSFLE/use-cases/sensitive-data-encryption.html


source/use-cases/sensitive-data-encryption.txt
```diff
@@ -203,6 +203,18 @@ Additional Dependencies
...
+      :tabid: python
+
+      .. list-table::
+         :header-rows: 1
+
+         * - Dependency Name
+           - Description
+
+         * - `pymongocrypt
+             <hhttps://pypi.org/project/pymongocrypt/>`_
```
terakilobyte: s/hhttps/https

source/includes/steps-fle-create-data-encryption-key.yaml
```diff
@@ -36,6 +36,19 @@ content: |
...
+     .. tab::
+        :tabid: python
+
+        .. code-block:: python
+           :emphasize-lines: 4
+
+           from bson import binary
+           
+           const path = "./master-key.txt";
+           local_master_key =  binary.Binary(open(path, "rb").read (96))
```
ccho-mongodb: Would it make sense to be consistent with:
-using a single space after `=`
-removing the space between functions (in this case, `read`) and the parameter?

source/includes/steps-fle-create-data-encryption-key.yaml
```diff
@@ -36,6 +36,19 @@ content: |
...
            });
+     .. tab::
+        :tabid: python
+
+        .. code-block:: python
+           :emphasize-lines: 4
+
+           from bson import binary
+           
+           const path = "./master-key.txt";
```
ccho-mongodb: Does `const` have significance in Python? I think maybe you meant:
```
path = "./master-key.txt";
```

source/includes/steps-fle-create-data-encryption-key.yaml
```diff
@@ -157,6 +181,39 @@ content: |
...
+               key_vault_namespace,
+               mongocryptd_bypass_spawn = True 
+           )
+ 
+           client = MongoClient(
+               connection_string,
+               auto_encryption_opts = fle_opts 
+           )
+ 
+           client_encryption = pymongo.encryption.ClientEncryption(
```
ccho-mongodb: Would it make sense to also add an import:
```
from pymongo.encryption import ClientEncryption
```
and use that here to be consistent with `MongoClient`?

source/includes/steps-fle-create-data-encryption-key.yaml
```diff
@@ -157,6 +181,39 @@ content: |
...
+               kms_providers,
+               key_vault_namespace,
+               client, 
+               OPTS
```
ccho-mongodb: I did not see `OPTS` defined anywhere, are you certain this belongs here?

source/includes/steps-fle-create-data-encryption-key.yaml
```diff
@@ -157,6 +181,39 @@ content: |
            npm package. See the `npmjs documentation on the uuid-base64
            package <https://www.npmjs.com/package/uuid-base64>`_ for
            installation instructions.
+     .. tab::
+        :tabid: python
+
+        .. code-block:: python
```
ccho-mongodb: Could this also highlight important lines similar to the other two driver examples?

source/includes/steps-fle-create-data-encryption-key.yaml
```diff
@@ -157,6 +181,39 @@ content: |
...
+ 
+           client_encryption = pymongo.encryption.ClientEncryption(
+               kms_providers,
+               key_vault_namespace,
+               client, 
+               OPTS 
+           )
+           data_key = client_encryption.create_data_key("local")
+           base64DataKeyId = base64.b64encode(data_key)
+           print("DataKeyId [uuid]: " + data_key)
```
ccho-mongodb: Nitpick: could we make this substitution `uuid` => `UUID` to be consistent with the other drivers' output?

source/includes/steps-fle-create-data-encryption-key.yaml
```diff
@@ -157,6 +181,39 @@ content: |
...
+        :tabid: python
+
+        .. code-block:: python
+           
+           from pymongo import MongoClient
+           from pymongo.encryption_options import AutoEncryptionOpts
+           import base64
+
+           connectionString = "mongodb://localhost:27017"
+           keyVaultNamespace = "encryption.keyvault"
```
terakilobyte: you refer to this as key_vault_namespace later which feels more pythonic

source/includes/steps-fle-create-data-encryption-key.yaml
```diff
@@ -291,8 +348,32 @@ content: |
...
            }
+     .. tab::
+        :tabid: python
 
+        .. code-block:: python
+           :emphasize-lines: 4
 
+           key = client.encryption.keystore.find_one({ "_id": data_key})
+           client_encryption.close()
+           client.close()
```
ccho-mongodb: Could this snippet explicitly print out the value of `key` similar to the other driver examples?

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -603,7 +637,24 @@ MedcoMD engineers write a function to create a new patient record:
                  console.log("writeError occurred:", writeError);
              })
          }
-
+   .. tab::
+      :tabid: python
+
+      .. code-block:: python
+
+         def insertPatient(collection, name, ssn, bloodType,  medicalRecords,  policyNumber, provider):
```
ccho-mongodb: Could you keep the spaces between parameter names consistent?

source/includes/steps-fle-configure-the-mongodb-client.yaml
```diff
@@ -80,6 +97,15 @@ content: |
            const patientSchema = {
              "medicalRecords.patients": jsonSchema,
            }
+     .. tab::
+        :tabid: python
+
+        .. code-block:: python
+           :emphasize-lines: 2
+
+           patientSchema = {
```
ccho-mongodb: Would it be better to use snake case to stay consistent with other variables like `local_master_key`?

source/includes/steps-fle-configure-the-mongodb-client.yaml
```diff
@@ -80,6 +97,15 @@ content: |
...
+             "medicalRecords.patients": jsonSchema,
```
ccho-mongodb: Would it be better to use snake case to stay consistent with other variables like `local_master_key`?

source/includes/steps-fle-configure-the-mongodb-client.yaml
```diff
@@ -137,6 +163,25 @@ content: |
...
+           :class: note
+
+           If the ``mongocryptd`` daemon is already running, you can
+           configure the client to skip starting it by passing the
+           following option:
+
+           .. code-block:: python
+              :emphasize-lines: 1
+
+               mongocryptd_bypass_spawn = True
```
ccho-mongodb: The other examples show where this variable gets passed in to construct `extraOptions` that gets passed to the MongoClientSettings. Rather than creating a variable, it could be better to show it directly as an optional parameter of `AutoEncryptionOpts`. The reason for this is that the code snippet that constructs the `AutoEncryptionOpts` in the next step uses this as a variable, but the snippet will likely **not work** if the value of the variable is not set. This would be common to any user that does not explicitly define the variable.

ccho-mongodb: For example, something like this may suffice (as well as a short explanation that the default value for the optional parameter is False):
```
fle_opts = AutoEncryptionOpts(
    ...
    mongocryptd_bypass_spawn = True
)
```

source/includes/steps-fle-configure-the-mongodb-client.yaml
```diff
@@ -181,4 +226,18 @@ content: |
...
+
+           const secureClient = new MongoClient(connectionString, {
+           fle_opts = AutoEncryptionOpts( 
+              kms_providers, 
+              key_vault_namespace,
+              schema_map = patientSchema,
+              mongocryptd_bypass_spawn = mongocryptd_bypass_spawn,
+              mongocryptd_spawn_path=mongocryptd_spawn_path 
+           )
+           client = MongoClient( connectionString, auto_encryption_opts = fle_opts )
```
ccho-mongodb: Would it be better to stay consistent with spacing for function calls/definitions? Here, there are spaces around the first and last parameters, but in other function calls/definitions, there are no spaces.

source/includes/steps-fle-convert-to-a-remote-master-key.yaml
```diff
@@ -69,6 +69,17 @@ content: |
...
                }
            }
+     .. tab::
+        :tabid: python
+
+        .. code-block:: python
+
+           "kmsProviders" = {
+               "aws": {
+                   "access_key_id":  "<IAM User Access Key ID>",
```
ccho-mongodb: I think there is an extra space here before the value.

source/includes/steps-fle-convert-to-a-remote-master-key.yaml
```diff
@@ -128,6 +139,17 @@ content: |
...
             console.log("DataKeyId [base64]: ", base64DataKeyId);
            });
+     .. tab::
+        :tabid: python
+
+        .. code-block:: python
+
+           connectionString = "mongodb://localhost:27017";
+           keyVaultNamespace = "encryption.__keyVault";
+           fle_opts = AutoEncryptionOpts( kms_providers , keyVaultNamespace, mongocryptd_bypass_spawn = True )
```
ccho-mongodb: I think there is an extra space after `kms_providers`. 
Also, per a prior comment:

> Would it be better to stay consistent with spacing for function calls/definitions? Here, there are spaces around the first and last parameters, but in other function calls/definitions, there are no spaces.

source/includes/steps-fle-convert-to-a-remote-master-key.yaml
```diff
@@ -128,6 +139,17 @@ content: |
...
+           client = MongoClient( connectionString, auto_encryption_opts = fle_opts )  
+           client_encryption = pymongo.encryption.ClientEncryption(kms_providers, keyVaultNamespace, client, OPTS )
```
ccho-mongodb: Would it be better to stay consistent with spacing for function calls/definitions? Here, there are spaces around the first and last parameters, but in other function calls/definitions, there are no spaces.

source/includes/steps-fle-convert-to-a-remote-master-key.yaml
```diff
@@ -128,6 +139,17 @@ content: |
...
            });
+     .. tab::
+        :tabid: python
+
+        .. code-block:: python
+
+           connectionString = "mongodb://localhost:27017";
+           keyVaultNamespace = "encryption.__keyVault";
+           fle_opts = AutoEncryptionOpts( kms_providers , keyVaultNamespace, mongocryptd_bypass_spawn = True )
+           client = MongoClient( connectionString, auto_encryption_opts = fle_opts )
```
ccho-mongodb: Would it be better to stay consistent with spacing for function calls/definitions? Here, there are spaces around the first and last parameters, but in other function calls/definitions, there are no spaces.

source/includes/steps-fle-configure-the-mongodb-client.yaml
ccho-mongodb: Another option is to mention where you would pass the following option around line 178 and then omit it from the next step.

source/includes/steps-fle-create-data-encryption-key.yaml
```diff
@@ -157,6 +181,39 @@ content: |
...
+           )
+ 
+           client_encryption = pymongo.encryption.ClientEncryption(
+               kms_providers,
+               key_vault_namespace,
+               client, 
+               OPTS 
+           )
+           data_key = client_encryption.create_data_key("local")
+           base64DataKeyId = base64.b64encode(data_key)
```
ccho-mongodb: Would it make sense to use snake case here to be consistent with other variable names?

source/includes/steps-fle-create-data-encryption-key.yaml
```diff
@@ -291,8 +348,32 @@ content: |
              status: 0,
              masterKey: { provider: 'local' }
            }
+     .. tab::
+        :tabid: python
 
+        .. code-block:: python
+           :emphasize-lines: 4
 
+           key = client.encryption.keystore.find_one({ "_id": data_key})
```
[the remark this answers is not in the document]
ccho-mongodb: Also, I think there is an extra space after the opening curly brace, before `"_id"`

source/includes/steps-fle-convert-to-a-remote-master-key.yaml
```diff
@@ -128,6 +139,17 @@ content: |
             const base64DataKeyId = key.toString('base64');
             console.log("DataKeyId [base64]: ", base64DataKeyId);
            });
+     .. tab::
+        :tabid: python
+
+        .. code-block:: python
```
ccho-mongodb: Could you make sure to emphasize important lines similar to the other driver examples?

source/includes/steps-fle-convert-to-a-remote-master-key.yaml
```diff
@@ -128,6 +139,17 @@ content: |
...
+
+           connectionString = "mongodb://localhost:27017";
```
ccho-mongodb: Could you use snake case here to stay consistent with the rest of the Python variable names?

source/includes/steps-fle-convert-to-a-remote-master-key.yaml
```diff
@@ -128,6 +139,17 @@ content: |
...
+           keyVaultNamespace = "encryption.__keyVault";
```
ccho-mongodb: Could you use snake case here to stay consistent with the rest of the Python variable names?

source/includes/steps-fle-configure-the-mongodb-client.yaml
```diff
@@ -181,4 +226,17 @@ content: |
                  extraOptions: extraOptions,
                }
            });
+     .. tab::
+        :tabid: python
+
+        .. code-block:: python
```
ccho-mongodb: emphasize

ccho-mongodb: Could you make sure to emphasize important lines similar to the other driver examples?

source/includes/steps-fle-configure-the-mongodb-client.yaml
```diff
@@ -52,6 +58,17 @@ content: |
                key: localMasterKey,
              }
            }
+     .. tab::
+        :tabid: python
+
+        .. code-block:: python
+           :emphasize-lines: 2,3
+
+           kmsProviders = {
```
ccho-mongodb: Would it be better to use snake case to stay consistent with other variables like `local_master_key`?

source/includes/steps-fle-configure-the-mongodb-client.yaml
```diff
@@ -52,6 +58,17 @@ content: |
...
+             "local": {
+               "key": localMasterKey,
```
ccho-mongodb: Would it be better to use snake case to stay consistent with other variables like the first mention of `local_master_key`?

source/includes/steps-fle-convert-to-a-remote-master-key.yaml
```diff
@@ -69,6 +69,17 @@ content: |
                    secretAccessKey: "<IAM User Secret Access Key>",
                }
            }
+     .. tab::
+        :tabid: python
+
+        .. code-block:: python
+
+           "kmsProviders" = {
```
ccho-mongodb: Did you mean to assign a variable here rather than a string?

source/includes/steps-fle-convert-to-a-remote-master-key.yaml
```diff
@@ -69,6 +69,17 @@ content: |
...
+               "aws": {
+                   "access_key_id":  "<IAM User Access Key ID>",
+                   "secretAccessKey": "<IAM User Secret Access Key>",
```
ccho-mongodb: I did not check the Python API, but am wondering whether `secretAccessKey` is correct or should it be snake case?

source/includes/steps-fle-convert-to-a-remote-master-key.yaml
mongomoe: Fixed:
- Switched to camel case

source/includes/steps-fle-create-data-encryption-key.yaml
mongomoe: Fixed:

- Switched to snake case

source/includes/steps-fle-convert-to-a-remote-master-key.yaml
mongomoe: Fixed:

- Removed extra space

source/includes/steps-fle-convert-to-a-remote-master-key.yaml
mongomoe: Fixed:

- Removed extra space

source/includes/steps-fle-convert-to-a-remote-master-key.yaml
mongomoe: Fixed:

- Removed extra space

Sorry about this spacing issue. Not sure why it keeps coming up, maybe something wrong with my vscode settings

source/includes/steps-fle-configure-the-mongodb-client.yaml
mongomoe: Fixed:

- Made the spacing similar to other function calls/definition

source/includes/steps-fle-configure-the-mongodb-client.yaml
mongomoe: I'd actually prefer to keep it the way that it is. The structure that the Node/Java examples use is:

In step four, Specify the Location of the Encryption Binary: they define extraOptions with a mongocryptdBypassSpawn variable
In step five, Create the MongoClient: they use the extraOptions defined in the last step (but do not redefine it) and pass it into the auto encryption settings.
The python driver unfortunately doesn't use the object/dict "extraOptions". Instead we pass it directly to the auto encryption settings/options. Thus I am following the same general pattern for python:

In step four, Specify the Location of the Encryption Binary: python snippet defines a mongocryptd_bypass_spawn variable
In step five, Create the MongoClient: python snippet uses the mongocryptd_bypass_spawn defined in the last step (but does not redefine it) and passes it into the auto encryption settings.
I think if it is safe to assume that the readers will either copy and paste the extraOptions from the last section or not (and receive an error), then it is also same to assume the same for mongocryptd_bypass_spawn

source/includes/steps-fle-create-data-encryption-key.yaml
mongomoe: Fixed:
- printed the key

source/includes/steps-fle-create-data-encryption-key.yaml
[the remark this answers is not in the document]
mongomoe: Fixed:
- Added the creation of the client
- Removed the extra space before _id

Let me know what you think of the updated snippet? Note, I didn't include the definition of kms_providers because if I included the definition of kms_providers I would need to include the definition of local_master_key

source/includes/steps-fle-create-data-encryption-key.yaml
mongomoe: Fixed:
- Added highlighting similar to other drivers

source/includes/steps-fle-create-data-encryption-key.yaml
mongomoe: Sorry, missing import. Fixed:

- added import: 
```
           from bson.codec_options import CodecOptions
           OPTS = CodecOptions(uuid_representation=STANDARD)
```

source/includes/steps-fle-create-data-encryption-key.yaml
mongomoe: It doesn't have significance in py. Fixed:

- Removed the 'const'

source/includes/steps-fle-create-data-encryption-key.yaml
mongomoe: It would make a lot of sense (actually not even sure how this ended up with the extra space). I've altered the following:

- Used a single space after = 
- Removed the extra space between functions

source/includes/steps-fle-create-data-encryption-key.yaml
```diff
@@ -291,8 +357,46 @@ content: |
              status: 0,
              masterKey: { provider: 'local' }
            }
+     .. tab::
+        :tabid: python
+
+        .. code-block:: python
+
```
terakilobyte: should this be a new file or just added into the script previous?

source/includes/steps-fle-create-data-encryption-key.yaml
```diff
@@ -36,6 +36,19 @@ content: |
...
+        :tabid: python
+
+        .. code-block:: python
+           :emphasize-lines: 4
+
+           from bson import binary
+           
+           path = "./master-key.txt";
+           local_master_key = binary.Binary(open(path, "rb").read(96))
+
```
ccho-mongodb: Nitpick: Although the additional line breaks are not rendered, could we limit them to at most, a single blank line?

source/includes/steps-fle-create-data-encryption-key.yaml
```diff
@@ -291,8 +357,46 @@ content: |
...
+           client_encryption.close()
+           client.close()
+
+
+        This code example should print a retrieved document that resembles the
+        following:
 
+        .. code-block:: none
 
+           {
```
terakilobyte: might want to use pprint to print it out prettily like this

source/includes/steps-fle-create-data-encryption-key.yaml
```diff
@@ -291,8 +357,46 @@ content: |
...
+             mongocryptd_bypass_spawn = True
+           )
+           client = MongoClient(
+             connection_string,
+             auto_encryption_opts = fle_opts
+           )
+           db = client["encryption"]
+           collection = db["__keyVault"]
+           key = collection.find_one({"_id": data_key})
+           print(key)
```
terakilobyte: see below comment, may want to use pprint

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -530,6 +565,11 @@ full `JSON Schema for the Medco Medical Management System
 
       View the `helper code in Javascript
       <https://raw.githubusercontent.com/mongodb/docs-assets/DOCSP-json-schema-helper-and-json/JSONSchemaCreator.js>`_.
+   .. tab::
+      :tabid: python
+
+      View the `helper code in Python
+      <https://raw.githubusercontent.com/mongodb/docs-assets/DOCSP-json-schema-helper-and-json/json_schema_creator.py>`_.
```
ccho-mongodb: Could we use camel case for the `keyId` function parameter in this snippet to be consistent?

source/includes/steps-fle-convert-to-a-remote-master-key.yaml
```diff
@@ -128,6 +139,21 @@ content: |
...
+           from bson.codec_options import CodecOptions
+           from bson.binary import (STANDARD)
+           OPTS = CodecOptions(uuid_representation=STANDARD)
+           
+           connection_string = "mongodb://localhost:27017";
+           key_vault_namespace = "encryption.__keyVault";
+           fle_opts = AutoEncryptionOpts(kms_providers, key_vault_namespace, mongocryptd_bypass_spawn = True)
+           client = MongoClient(connection_string, auto_encryption_opts = fle_opts)  
+           client_encryption = pymongo.encryption.ClientEncryption(kms_providers, key_vault_namespace, client, OPTS)
+           data_key = client_encryption.create_data_key("local")
```
terakilobyte: is this supposed to be "local" when using aws?

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -203,6 +203,18 @@ Additional Dependencies
...
            - Convert between Base64 and hexadecimal UUIDs.
+   .. tab::
+      :tabid: python
+
+      .. list-table::
+         :header-rows: 1
+
+         * - Dependency Name
+           - Description
+
```
ccho-mongodb: This is more of a question for @terakilobyte and @nlarew -- since it appears from the ecosystem pages that we support two Python Drivers (PyMongo and Motor), should we specify that this uses pymongo only, and if so, where?

source/includes/steps-fle-create-data-encryption-key.yaml
```diff
@@ -157,6 +181,48 @@ content: |
...
+ 
+           client_encryption = ClientEncryption(
+               kms_providers,
+               key_vault_namespace,
+               client, 
+               OPTS 
+           )
+           data_key = client_encryption.create_data_key("local")
+ 
+           uuid_data_key_id = uuid.UUID(bytes=data_key)
```
ccho-mongodb: I noticed for optional params in function calls, sometimes you're using a space between the "=" and sometimes no space. Could you keep it consistent whether it's part of PEP8? I would recommend a single space on each side of the "=". There are other instances, so make sure that you check all the python snippets.

source/includes/steps-fle-configure-the-mongodb-client.yaml
ccho-mongodb: I would recommend mentioning it in "Create the MongoDB Client Step 3:
Map the JSON Schema to the Patients Collection" since that is where it is used. Instead of adding any code, you can just mention that they would just assign the value they built in the prior section.

source/includes/steps-fle-configure-the-mongodb-client.yaml
ccho-mongodb: I think I see the misunderstanding. This comment is going to be a little lengthy, so feel free to ask me in person to go through it.

To summarize my request:
-Since setting mongocryptd_bypass_spawn is mentioned as optional, we should not assume that it has been defined. It has a default value, but if we pass `mongocryptd_bypass_spawn = mongocryptd_bypass_spawn` and the variable has not been defined, it will error.

To summarize what I read from your response:
1) Python (or pymongo) does not have the same call syntax as Java to allow us to make it optionally included.
2) The reader will just copy and paste extraOptions from the prior section (assuming you meant from the same step

For 2, are you referring to "B. Create a Data Encryption Key Step 4"? This is two sections prior to "D. Create the MongoDB Client". I think for effective and maintainable guides, we should try to keep sections self-enclosed.

For 1, I was not concerned there is a feature of python2/3 called "unpacking" that allows you to create something very similar to the Java example.
Here's a link to the python doc: https://docs.python.org/3/tutorial/controlflow.html#tut-unpacking-arguments
Here's a sample:

```
def my_func(a, b = True, c = "mongodb://localhost:27017"):
   print("a:", a)
   print("b:", b)
   print("c:", c)

>>> opts = { "hello": "false", "c": "mongodb://localhost:27018" }
>>> my_func("foobar", **opts)
```

From looking at the Node example, I'd like to request a change to that snippet too. When showing`mongocryptd_bypass_spawn`, instead of showing it as the only parameter of `extraOptions` (const), append it to extraOptions (var) in the admonition.

source/includes/steps-fle-create-data-encryption-key.yaml
```diff
@@ -291,8 +357,46 @@ content: |
...
+             kms_providers,
+             key_vault_namespace,
+             mongocryptd_bypass_spawn = True
+           )
+           client = MongoClient(
+             connection_string,
+             auto_encryption_opts = fle_opts
+           )
+           db = client["encryption"]
+           collection = db["__keyVault"]
```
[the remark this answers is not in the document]
mongomoe: Fixed:
* Switched "keyvault" -> "keyVault"

source/includes/steps-fle-create-data-encryption-key.yaml
```diff
@@ -157,6 +181,48 @@ content: |
            npm package. See the `npmjs documentation on the uuid-base64
            package <https://www.npmjs.com/package/uuid-base64>`_ for
            installation instructions.
+     .. tab::
+        :tabid: python
+
+        .. code-block:: python
+           :emphasize-lines: 14,15,24,30
+
+           from pymongo import MongoClient
```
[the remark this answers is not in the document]
mongomoe: Fixed, but I've changed the snippet a bit, so the imports are different. All imports are needed, and there aren't missing imports now:

```
           from pymongo import MongoClient
           from pymongo.encryption_options import AutoEncryptionOpts
           from pymongo.encryption import ClientEncryption
           import base64
           from bson.codec_options import CodecOptions
           from bson.binary import STANDARD
           import uuid
```

source/includes/steps-fle-create-data-encryption-key.yaml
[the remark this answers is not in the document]
mongomoe: Fixed:

* Added import for `uuid`

source/includes/steps-fle-create-data-encryption-key.yaml
mongomoe: * Switched this to a function snippet so this can be run as a standalone file or part of a monolithic script for this whole page

source/includes/steps-fle-create-data-encryption-key.yaml
mongomoe: Fixed:

* added pprint import & usage

source/includes/steps-fle-create-data-encryption-key.yaml
mongomoe: Fixed:

* added pprint import & usage

source/includes/steps-fle-convert-to-a-remote-master-key.yaml
mongomoe: Fixed:
* Switched "local" to "aws"

source/includes/steps-fle-create-data-encryption-key.yaml
```diff
@@ -291,8 +357,46 @@ content: |
...
+             key_vault_namespace,
+             mongocryptd_bypass_spawn = True
+           )
+           client = MongoClient(
+             connection_string,
+             auto_encryption_opts = fle_opts
+           )
+           db = client["encryption"]
+           collection = db["__keyVault"]
+           key = collection.find_one({"_id": data_key})
```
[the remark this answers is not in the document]
mongomoe: Nathan & I talked about making this guide a bit more functional, in terms of having the snippets work as their own standalone functions, so that people can copy and paste the code and not be lost in the steps. Given that discussion, I've altered the snippet to make it a function. 

With the new snippet, you can either pass in the data_key as the explicit string or as the actual data_key variable. However, I think having the explicit string doesn't particularly make sense even for java or node. I can't see people explicitly querying for a specific key in their application in a real-world setting. Potentially we could fix the Java & Node examples retroactively in a new ticket.

What do you think?

source/includes/steps-fle-configure-the-mongodb-client.yaml
```diff
@@ -181,4 +226,19 @@ content: |
...
+
+        .. code-block:: python
+           :emphasize-lines: 2-6,8
+
+           fle_opts = AutoEncryptionOpts( 
+              kms_providers, 
+              key_vault_namespace,
+              schema_map = patient_schema,
+              mongocryptd_bypass_spawn = mongocryptd_bypass_spawn,
+              mongocryptd_spawn_path=mongocryptd_spawn_path
```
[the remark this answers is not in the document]
mongomoe: Thanks for the suggestion. Ran all my code through python-autopep now. Will let you know about any tools for the list

source/includes/steps-fle-create-data-encryption-key.yaml
```diff
@@ -157,6 +181,48 @@ content: |
...
+           from pymongo import MongoClient
+           from pymongo.encryption_options import AutoEncryptionOpts
+           from pymongo.encryption import ClientEncryption
+           import base64
+           from bson.codec_options import CodecOptions
+           from bson.binary import (Binary,
+                         JAVA_LEGACY,
+                         STANDARD,
+                         UUID_SUBTYPE)
+           OPTS = CodecOptions(uuid_representation=STANDARD)
```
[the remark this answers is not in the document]
mongomoe: Fixed:
* Ran all the py code through a vscode linter

source/includes/steps-fle-create-data-encryption-key.yaml
mongomoe: Fixed:

* Ran all the py code through a vscode linter

source/use-cases/sensitive-data-encryption.txt
* Switched `keyId` -> `key_id` for json_schema_creator.py

source/includes/steps-fle-create-data-encryption-key.yaml
mongomoe: Fixed:
* Ran code through python linter

source/includes/steps-fle-configure-the-mongodb-client.yaml
mongomoe: Added it to the Create the MongoDB Client Section, under "Map the JSON Schema to the Patients Collection". Here's the updated verbiage:

The MedcoMD engineers assign their schema to a variable. The JSON Schema that MedcoMD defined doesn’t explicitly specify the collection to which it applies. To assign the schema, they map it to the medicalRecords.patients collection namespace:

source/includes/steps-fle-configure-the-mongodb-client.yaml
mongomoe: Fixed, switched snippets roughly to:

```
           extra_options = {
              'mongocryptd_spawn_path': '/usr/local/bin/mongocryptd'
           }
           extra_options['mongocryptd_bypass_spawn'] = True
           fle_opts = AutoEncryptionOpts( 
              kms_providers, 
              key_vault_namespace,
              schema_map = patient_schema,
              **extra_options
           )
           client = MongoClient(connection_string, auto_encryption_opts = fle_opts)
```

source/includes/steps-fle-create-data-encryption-key.yaml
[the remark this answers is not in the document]
ccho-mongodb: I think we should be consistent in usage. If we use functions/methods only sometimes, what are the specific conditions that warrant them? In the Java code examples, I think the conditions match at least one of the following (open for debate or other ideas):

1) The snippet is a complete, self-contained program, structured idiomatically for that driver language.
2) It is a snippet from a complete, self-contained program taken verbatim.
3) We preface the snippet with "create a function/method".

I think it would be great if we just structured snippets in this step file to follow 2 in which we omit imports and instead depend on them downloading the complete example.

source/includes/steps-fle-configure-the-mongodb-client.yaml
```diff
@@ -137,6 +164,28 @@ content: |
...
+        :tabid: python
+
+        .. code-block:: python
+           :emphasize-lines: 2
+      
+           extra_options = {
+              'mongocryptd_spawn_path': '/usr/local/bin/mongocryptd'
+           }
+
+        .. admonition:: Encryption Binary Daemon
```
ccho-mongodb: I like this, but can you update the Node and Java examples so that they do not re-declare `extraOptions`?

source/includes/steps-fle-convert-to-a-remote-master-key.yaml
```diff
@@ -128,6 +139,42 @@ content: |
...
+           key_vault_namespace = "encryption.__keyVault"
+           fle_opts = AutoEncryptionOpts(
+              {
+                 "aws": {
+                       "accessKeyId": "<IAM User Access Key ID>",
+                       "secretAccessKey": "<IAM User Secret Access Key>"
+                 }
+              },
+              key_vault_namespace,
+              mongocryptd_bypass_spawn=False
```
ccho-mongodb: Question: is this necessary, or is the default value `True`?

source/includes/steps-fle-create-data-encryption-key.yaml
```diff
@@ -157,6 +179,52 @@ content: |
...
+               data_key = client_encryption.create_data_key("local")
+               uuid_data_key_id = uuid.UUID(bytes=data_key)
+               base_64_data_key_id = base64.b64encode(data_key)
+
+               print("DataKeyId [UUID]: ", str(uuid_data_key_id))
+               print("DataKeyId [base64]: ", base_64_data_key_id)
+
+               return data_key
+
+           data_key = create_data_encryption_key()
```
ccho-mongodb: Since anything after this line is out of scope, would it make more sense to call ```create_data_encryption_key()``` without a variable assignment?

source/includes/steps-fle-create-data-encryption-key.yaml
```diff
@@ -274,7 +342,7 @@ content: |
         .. code-block:: none
 
            {
-             _id: Binary {
+             _id: Binary {o
```
ccho-mongodb: Is this a type-o?

source/use-cases/sensitive-data-encryption.txt
ccho-mongodb: Could you run the linter on this file?

source/includes/steps-fle-convert-to-a-remote-master-key.yaml
```diff
@@ -128,6 +139,42 @@ content: |
...
+        :tabid: python
+
+        .. code-block:: python
+
+           import pymongo
+           from pymongo import MongoClient
+           from pymongo.encryption_options import AutoEncryptionOpts
+           from bson.codec_options import CodecOptions
+           from bson.binary import (STANDARD)  
+           connection_string = "mongodb://localhost:27017"
```
ccho-mongodb: This could be easier to read if there was a line break between the imports and logic.

source/includes/steps-fle-create-data-encryption-key.yaml
```diff
@@ -291,8 +359,44 @@ content: |
...
+
+        This code example should print a retrieved document that resembles the
+        following:
+
+        .. code-block:: none
+
+           {
+             "_id": UUID('1e83d013-d873-47df-abb1-e57898a72d4c'),
+             "keyMaterial":  b'\x96#\xeb\xa1xKA\xa7GM\xef\x08\xc04\'gD\x96\x9c\xa4\xd3?\xe3Db0H\xbb\x86\xa5\xc2\x1f\x14\x0f\xb8\xb8\x8a\x9d\xc9\xae\xa1g\xaf\xeb\x8b\x99\xb4b"\xc0\xe8e\x07\x1b.\xeet\xf5<%\xfd\x06Y\x15o=3Yk\x9fue\xd8V#X\xc1IB\xf5\xc9+\x95\xb1\x9c\xc0\x08U?\xaf\xb1U\xc6\x84\x89\x9b\xdc\x98\xc9~\xb2\xbd\xf6\\\xa2y\x08\xdf\x8f\xa1\x03\t9\xe7_+J>_H\xb4\x97up\x93Sc\x88\x0fG-+\x86\x95\x9e\xc2\x8es\x9e\xcb%%lVQ\xa2\xf1\xe3W\x83\x10]\xc9\x1fm\x7f\xbc\xbf\xd2d',
+             "creationDate":  datetime.datetime(2019, 9, 30, 20, 43, 10, 951000),
```
ccho-mongodb: Nitpick: looks like there's an extra space here and in the line above. Is this verbatim what gets printed?

source/use-cases/sensitive-data-encryption.txt
```diff
@@ -603,7 +643,24 @@ MedcoMD engineers write a function to create a new patient record:
                  console.log("writeError occurred:", writeError);
              })
          }
-
+   .. tab::
+      :tabid: python
+
+      .. code-block:: python
+
+         def insert_patient(collection, name, ssn, blood_type, medical_records, policy_number, provider):
```
ccho-mongodb: Nitpick: no action needed here, just a coding tip. Long parameter lists are considered a "code smell" that indicates it should be refactored. One simple way to reduce the number of parameters is to store them in a struct/dict/map and pass that as a parameter (similar to what you did with unpacking `extra_options`).

source/includes/steps-fle-create-data-encryption-key.yaml
```diff
@@ -291,8 +359,44 @@ content: |
...
+        :tabid: python
+
+        .. code-block:: python
+
+           from pprint import pprint
+           connection_string = "mongodb://localhost:27017"
+           key_vault_namespace = "encryption.__keyVault"
+
+
+           def verify_data_key_created(fle_opts, client, client_encryption,  data_key):
```
ccho-mongodb: Looks like there's an extra space in here. Could you run the linter on this snippet, just in case?

ccho-mongodb: Two more nitpicks: 
1) The parameter order here would make more sense with the important arguments first. Since you are performing a "verify" on a "data key", perhaps `data_key` could be first, then `client` and then `client_encryption` & `fle_opts` (next to each other since they are related).
2) The side-effect of closing clients is unexpected by a helper function that does not create the new client. Usually, you would expect that in the scope of the function that constructed the client. Since no clients are being created in this snippet, can we omit lines 377-378?

source/includes/steps-fle-create-data-encryption-key.yaml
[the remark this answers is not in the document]
terakilobyte: I think it's easier to just import UUID from `bson.binary` since you are already importing something from it. Is there a reason to prefer adding it via import `uuid`?

source/includes/steps-fle-create-data-encryption-key.yaml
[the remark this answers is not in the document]
terakilobyte: We talked about this approach, and I'm ok with it. It just has to be a working example that people can follow along to.

source/use-cases/sensitive-data-encryption.txt
mongomoe: Done :)

source/includes/steps-fle-configure-the-mongodb-client.yaml
mongomoe: I'd love to do that, I just don't want to hold python for this. I created a ticket for doing this: https://jira.mongodb.org/browse/DOCSP-7332 for next sprint

source/includes/steps-fle-convert-to-a-remote-master-key.yaml
mongomoe: Default is false. I've removed "mongocryptd_bypass_spawn=False" since I see the other driver's adjacent snippet doesn't use this variable here.

source/includes/steps-fle-create-data-encryption-key.yaml
mongomoe: `data_key` is outside of the function and is accessible if the reader copies and pastes this, it's also required for the next snippet (which mentions using data_key)

source/includes/steps-fle-create-data-encryption-key.yaml
mongomoe: Haha yup & removed now

source/includes/steps-fle-convert-to-a-remote-master-key.yaml
mongomoe: Fixed!

source/includes/steps-fle-create-data-encryption-key.yaml
mongomoe: Not verbatim what gets printed, the actual log is quite a bit uglier even w/ pprint (pretty print). The real thing looks more like this:

```
{u'_id': UUID('46ade0a2-fe52-4dec-af7d-ef3963a66414'),
 u'creationDate': datetime.datetime(2019, 10, 15, 13, 46, 26, 126000),
 u'keyMaterial': Binary('\xb8\x1c\xf4_\x18\t\xf3\xa8\xe1\x96p\x87|\xdb\x0e4G:\xf3/\x03\x04c\xfb\xb0\x19\xe6\xadPg\xd2G\xcaH\x8f\xe0\x0fsAG\x06\x9c\xe8\xc0\xbf}\x8f{\xa2\xcd\xce\xb2Z<\xc9mR\xda\xd1&\x8a\xc5aeTzx\xe9\xcf\xad&\xb5y\x89\xa6Cn\xfc6\x87\xc6\x1c\x8b\xa5\x16\xfb\xa1\xf97xx\x91\x0b1\x03\x82m\xbc\xe0\x85\xb1<3eO\r\xba\xc6\x8c\x11\x8b\x0f0lK\xa3\xae\x91\x94^\xdb\x9d\x89T\xee\x898}:^X\x82+\xbaK\x9f\x96\x18\x8f\xb5\xea\x18\xf5\x15 ~\x96H\x99\xab\x06D\xc9\x19\xa6\xbb\xe4\x8e\x7f\xa2', 0),
 u'masterKey': {u'provider': u'local'},
 u'status': 0,
 u'updateDate': datetime.datetime(2019, 10, 15, 13, 46, 26, 126000)}
```


---- On the bright side, I just pushed a fix for the extra whitespace

source/use-cases/sensitive-data-encryption.txt
mongomoe: Cool thanks for the tip, will keep that in mind for the future

source/includes/steps-fle-create-data-encryption-key.yaml
mongomoe: Fixed:
* Ran Linter
* Fixed parameter order per your spec ^
* Removed closing of the client

source/includes/steps-fle-create-data-encryption-key.yaml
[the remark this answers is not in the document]
mongomoe: For Python, largely everything is wrapped in a function unless it is a snippet to create variables that should be accessible to other snippets (for instance local_master_key is defined outside of any function because we require that later). 

I think this fits under Nathan's requirement of it being a working example that people can follow along with.

source/includes/steps-fle-create-data-encryption-key.yaml
ccho-mongodb: Did you check in the fixes mentioned above? For some reason, I do not see the changes yet.

source/includes/steps-fle-create-data-encryption-key.yaml
ccho-mongodb: Ah ok, I just meant the extra space as I would not have expected pretty print to add one.

source/includes/steps-fle-create-data-encryption-key.yaml
mongomoe: Re-pushed

mongomoe:
"Don't forget line 382 where you call the function!" @ccho-mongodb 
Fixed!

source/includes/steps-fle-create-data-encryption-key.yaml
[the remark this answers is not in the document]
mongomoe: Fixed

PR #555, opened 2019-10-10
DOCSP-7164 - Kafka Sink Connector - Post-processor

ccho-mongodb:
JIRA: https://jira.mongodb.org/browse/DOCSP-7164

Staging: https://docs-mongodbcom-staging.corp.mongodb.com/ecosystem/docsworker/DOCSP-7164-sink-connector-3/connectors/kafka-sink-postprocessors.html


ccho-mongodb:
Please review only "source/connectors/kafka-sink-postprocessors.txt"

source/connectors/kafka-sink-postprocessors.txt
```diff
@@ -9,10 +9,821 @@ Kafka Sink Connector Post-Processors
...
+   [key|value].projection.type=blacklist
+   [key|value].projection.list=population,forecast.*.*
+
+The record contains the following data after applying the projection:
+
+.. code-block:: json
+
+   {
+     "forecast": [
+       { "day": "Monday",
```
terakilobyte: formatting, move key/value to new line

source/connectors/kafka-sink-postprocessors.txt
```diff
@@ -9,10 +9,821 @@ Kafka Sink Connector Post-Processors
...
+       | Default value for ``writemodel.strategy`` configuration setting.
+
+   * - ReplaceOneBusinessKeyStrategy
+     - | Replaces at most one document that matches filters provided by the ``document.id.strategy`` setting.
+       | Set the following configuration: ``writemodel.strategy=com.mongodb.kafka.connect.sink.writemodel.strategy.ReplaceOneBusinessKeyStrategy``
+
+       .. seealso:: Example of usage in :ref:`writemodel-strategy-business-key`.
+
+   * - DeleteOneDefaultStrategy
+     - | Deletes at most one document that contain null value sink records that matches the id specified by the ``document.id.strategy`` setting.
```
terakilobyte: s/matches/match ?

source/connectors/kafka-sink-postprocessors.txt
```diff
@@ -9,10 +9,821 @@ Kafka Sink Connector Post-Processors
...
+   specify a user-created, custom strategy in the configuration settings.
+
+.. _writemodel-strategy-business-key:
+
+WriteModel Strategy: Business Keys
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+A business key is a value comprised of data within the sink record that
+identifies it as a unique document. This example shows how to define
+a business key using the data contained by multiple fields in the record as
```
terakilobyte: I think this should read `contained within`, thoughts?

source/connectors/kafka-sink-postprocessors.txt
```diff
@@ -9,10 +9,821 @@ Kafka Sink Connector Post-Processors
...
+well as instruct the post processor to generate *BSON ObjectIds* for
```
terakilobyte: s/instruct/instructs ?

source/connectors/kafka-sink-postprocessors.txt
```diff
@@ -9,10 +9,821 @@ Kafka Sink Connector Post-Processors
...
+inserts, but not for updates.
+
+The following steps are required to configure this strategy:
+
+#. Create a **unique index** that corresponds to your business key in your
```
terakilobyte: should we link to unique indexes here?

source/connectors/kafka-sink-postprocessors.txt
```diff
@@ -9,10 +9,821 @@ Kafka Sink Connector Post-Processors
...
+   target MongoDB collection.
+#. In the connector configuration, specify the ``PartialValueStrategy``
+   as the id strategy to identify the fields that belong to the business
+   key.
+#. In the connector configuration, specify the
+   ``ReplaceOneBusinessKeyStrategy`` writemodel strategy.
+
+Suppose the parking usage data was being tracked for city blocks, defined
```
terakilobyte: This feels abrupt, I think because of `Suppose the` instead of `Suppose that`. I scrolled back up the page to make sure I hadn't missed parking usage data mentioned elsewhere

source/connectors/kafka-sink-postprocessors.txt
```diff
@@ -9,10 +9,821 @@ Kafka Sink Connector Post-Processors
...
+by distinct values for ``street`` and ``avenue``. An example message
```
terakilobyte: Street and Avenue feel strange. Perhaps houseNumber and street? http://desktop.arcgis.com/en/arcmap/10.3/guide-books/geocoding/what-is-an-address.htm

source/connectors/kafka-sink-postprocessors.txt
```diff
@@ -9,10 +9,821 @@ Kafka Sink Connector Post-Processors
...
+The sample data inserted into the collection contains the following:
+
+.. code-block:: json
+   :emphasize-lines: 2
+
+   {
+     "_id": ObjectId('5db0b81199b7fbcc89631d06'),
+     "street": "50th Street",
+     "avenue": "Broadway",
+     "parking_spaces": {
```
terakilobyte: just a note that I was surprised to see parking_spaces still included with the whitelist of street,avenue

source/connectors/kafka-sink-postprocessors.txt
```diff
@@ -9,10 +9,821 @@ Kafka Sink Connector Post-Processors
...
+     "_id": "MN-1234",
+     "_insertedTS": ISODate('2019-10-23T15:08:000Z"),
+     "_modifiedTS": ISODate("2019-10-23T15:08:000Z"),
+     "start": "Beacon",
+     "destination": "Grand Central"
+     "position": [ 40.805, -73.939 ]
+   }
+
+After an hour, the train reports its current location along its route. The
+``position`` and ``_insertedTS`` fields are updated:
```
terakilobyte: shouldn't this be `_modifiedTS`?

source/connectors/kafka-sink-postprocessors.txt
```diff
@@ -9,10 +9,821 @@ Kafka Sink Connector Post-Processors
...
+.. code-block:: json
+   :emphasize-lines: 4,7
+
+   {
+     "_id": "MN-1234",
+     "_insertedTS": ISODate('2019-10-23T15:08:000Z"),
+     "_modifiedTS": ISODate("2019-10-23T16:08:000Z"),
+     "start": "Beacon",
+     "destination": "Grand Central"
+     "position": [ 41.156, -73.870 ]
```
terakilobyte: I think we should prefer GeoJson here, but not a hard nit. Thoughts?

source/connectors/kafka-sink-postprocessors.txt
```diff
@@ -9,10 +9,821 @@ Kafka Sink Connector Post-Processors
 .. contents:: On this page
    :local:
    :backlinks: none
-   :depth: 1
+   :depth: 2
```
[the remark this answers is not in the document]
ccho-mongodb: Thanks for spotting, will fix.

source/connectors/kafka-sink-postprocessors.txt
```diff
@@ -9,10 +9,821 @@ Kafka Sink Connector Post-Processors
...
+
+       .. seealso:: :ref:`Strategy options and configuration <config-document-id-adder>`.
+
+   * - BlacklistKeyProjector
+     - | Full Path: ``com.mongodb.kafka.connect.sink.processor.BlacklistKeyProjector``
+       | Removes matching key fields from the sink record.
+
+       .. seealso:: :ref:`Configuration <config-blacklist-whitelist>` and :ref:`Example <blacklist-example>`.
+
+   * - BlacklistValueProjector
```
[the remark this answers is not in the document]
ccho-mongodb: Was I missing a value projection example? Both key and value projection descriptions link to the same example which describes how to configure it for either field. In Kafka sink documents, there are two fields, "key" and "value". The key and value can be JSON or a string. While the "key" might be more commonly used to store metadata, they are two separate "documents" with different names with the same set of restrictions.

source/connectors/kafka-sink-postprocessors.txt
ccho-mongodb: Good catch, will update.

source/connectors/kafka-sink-postprocessors.txt
ccho-mongodb: I think this is a confusing sentence, so I will rearrange the phrasing to more closely match the other items. Since "matches" is referring to the "at most one document" part, I think the singular form is appropriate. However, the phrase inserted to describe the null value record makes it more confusing what the verb references.

source/connectors/kafka-sink-postprocessors.txt
ccho-mongodb: Sounds good, thanks for spotting.

source/connectors/kafka-sink-postprocessors.txt
ccho-mongodb: Sure, is the server manual link one ok (https://docs.mongodb.com/manual/core/index-unique/)? Since they're working with a connector, this is a shift of persona responsibilities so I wonder if we should start to note this is a task for a DBA/server administrator.

source/connectors/kafka-sink-postprocessors.txt
ccho-mongodb: Do you think it might over-complicate a Kafka Connector guide to bring in GeoJSON which has no differences in behavior from any other field?

source/connectors/kafka-sink-postprocessors.txt
ccho-mongodb: Good catch, will fix.

source/connectors/kafka-sink-postprocessors.txt
ccho-mongodb: I am fine with any modification, but first would like to know whether you're requesting something different. I would like to adhere to the following goals regardless of the modification:

1) Connect the reader to an example they can conceptualize (rather than fieldA, fieldB which is in the currently published connector docs).
2) Keep the example as concise as possible. The minimum number of fields required to demonstrate a compound business key is two fields, and the minimum number of fields outside the business key is one.

That said, as mentioned in the example I am describing city blocks, so a simple substitution of houseNumber and street may not make sense for parking_spaces. Was there some other transient data that you were thinking of that I could use instead?

terakilobyte: I took this to mean "location".  The example immediately led me to think "what about parking spaces that aren't at an intersection".

source/connectors/kafka-sink-postprocessors.txt
terakilobyte: I suppose that makes me wonder why there was effort put into [the rfc](https://tools.ietf.org/html/rfc7946). It does add overhead though, so not a hard nit.

source/connectors/kafka-sink-postprocessors.txt
terakilobyte: Still needs changed.

source/connectors/kafka-sink-postprocessors.txt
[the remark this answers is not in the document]
terakilobyte: Reading through it, it looked like it. I only saw keys being whitelisted/blacklisted, not values. Am I missing something?

source/connectors/kafka-sink-postprocessors.txt
ccho-mongodb: Fixed in latest commit 11267f48f8ae578399a2902334b14db2bbb36a8d

source/connectors/kafka-sink-postprocessors.txt
ccho-mongodb: Ok, I will change the example.

source/connectors/kafka-sink-postprocessors.txt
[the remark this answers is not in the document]
ccho-mongodb: Discussed in person. Let me know if the `[key|value]` notation does not make sense in the examples.

PR #558, opened 2019-10-14
DOCSP-7163: Sink Connector - Data Formats

ccho-mongodb:
JIRA:
https://jira.mongodb.org/browse/DOCSP-7163

Staging:
https://docs-mongodbcom-staging.corp.mongodb.com/ecosystem/docsworker/DOCSP-7163-sink-connector-2/connectors/kafka-sink-data-formats.html

Only `source/connectors/kafka-sink-data-formats.txt` is relevant to this PR.


source/connectors/kafka-sink-data-formats.txt
```diff
@@ -9,10 +9,338 @@ Kafka Sink Connector Data Formats
...
+
+.. list-table::
+   :header-rows: 1
+   :stub-columns: 1
+   :widths: 1 4
+
+   * - Format Name
+     - Description
+
+   * - Avro
```
terakilobyte: Should this be AVRO to match the other mentions of it?

source/connectors/kafka-sink-data-formats.txt
```diff
@@ -9,10 +9,338 @@ Kafka Sink Connector Data Formats
...
+     - Serialized as a String. The JSON structure is not managed by
+       Kafka Connect.
+
+       .. note::
+
+          This format requires that the fields contain valid JSON even
+          when specifying the `StringConverter
+          <https://kafka.apache.org/21/javadoc/index.html?org/apache/kafka/connect/storage/StringConverter.html>`_
+
+.. note::
```
terakilobyte: These two notes are strikingly similar. Can they be combined?

source/connectors/kafka-sink-data-formats.txt
```diff
@@ -9,10 +9,338 @@ Kafka Sink Connector Data Formats
...
+     ]
+   }
+
+
+.. _avro-logical-types-example:
+
+AVRO Logical Type Example
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The following AVRO schema demonstrates use of each supported logical types
```
terakilobyte: `s/types/type`

source/connectors/kafka-sink-data-formats.txt
```diff
@@ -9,10 +9,338 @@ Kafka Sink Connector Data Formats
...
+             "type": "long",
+             "logicalType": "timestamp-micros"
+           }
+         }
+       ]
+     }
+
+.. note::
+
+   If you use AVRO code generation to generate logical types for
```
terakilobyte: This feels overly wordy. Perhaps

The following type mappings are created when using AVRO code generation for a Java-based producer application. 

Thoughts?

source/connectors/kafka-sink-data-formats.txt
ccho-mongodb: Will fix, thanks.

source/connectors/kafka-sink-data-formats.txt
ccho-mongodb: Good catch, will fix.

source/connectors/kafka-sink-data-formats.txt
ccho-mongodb: In both my existing sentence and your recommended replacement, I spotted what I think is passive voice "Xs are created..."

I will reword this to eliminate passive voice and provide emphasis on the fact that the generated types are not native to Java's system libraries.

source/connectors/kafka-sink-data-formats.txt
```diff
@@ -0,0 +1,346 @@
...
+       .. note::
+
+          This format requires that the fields contain valid JSON even
+          when specifying the `StringConverter
+          <https://kafka.apache.org/21/javadoc/index.html?org/apache/kafka/connect/storage/StringConverter.html>`_
+
+.. note::
+
+   **RAW JSON** mode
+   requires the Strings to be valid JSON
```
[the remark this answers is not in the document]
ccho-mongodb: Thanks for spotting this. I've reworded this per one of Nathan's unrelated comments.

Since I'm mentioning a named setting, "RAW JSON", I think it is ok to then refer to JSON format in the same sentence. Were you concerned that I was defining a type of JSON using the term JSON in the explanation?

source/connectors/kafka-sink-data-formats.txt
```diff
@@ -0,0 +1,346 @@
...
+   key.converter.schemas.enable=false
+
+   value.converter=org.apache.kafka.connect.json.JsonConverter
+   value.converter.schemas.enable=false
+
+.. admonition:: Choose the appropriate format
+   :class: note
+
+   Top-level JSON objects named ``schema`` or ``payload`` are parsed
+   literally instead of as a validation schema.
```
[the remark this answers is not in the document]
ccho-mongodb: Yes, I think so. Is that the term commonly used for not treating it as schema metadata and instead creating objects?

Also, I think the admonition title isn't enough context so I updated the sentence to help clarify. Let me know what you think.

source/connectors/kafka-sink-data-formats.txt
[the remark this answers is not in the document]
mongomoe: I like the new phrasing! My concern, at the time, wasn't with defining a type of JSON using the word JSON in the explanation. It was actually with the repetitiveness of the sentence. But even though the new sentence is actually a bit more repetitive than the previous rendition, I actually think it reads well and is easy to understand.

source/connectors/kafka-sink-data-formats.txt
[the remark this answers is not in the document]
mongomoe: The reason I mentioned explicitly is that using the word explicit is to contrast against the word 'implicit'.  For instance, the same (original) sentence could be roughly rephrased as "Top-level JSON objects that are named schema or payload are not parsed implicitly as a validation schema."

Vs if you had gone with the word "literally", it would be contrasting against the word figuratively. The following rendition of the above sentence doesn't make sense:
"Top-level JSON objects that are named schema or payload are not parsed figuratively as a validation schema."

That being said, I like the new phrasing a lot!

source/connectors/kafka-sink-data-formats.txt
```diff
@@ -9,10 +9,331 @@ Kafka Sink Connector Data Formats
...
+             "type": "long",
+             "logicalType": "timestamp-micros"
+           }
+         }
+       ]
+     }
+
+.. note::
+
+   If you use compile your schema using AVRO, it maps your logical types
```
terakilobyte: Looks like a small grammar mistake got introduced. Also, maybe "AVRO compiles your schema types to the following non-standard Java classes" ? Not a hard nit except for the grammar at the beginning of the sentence.

source/connectors/kafka-sink-data-formats.txt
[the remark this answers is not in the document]
ccho-mongodb: I'm still not sure what part is repetitive. Was there mention of the note elsewhere? What part of it is more repetitive than before?

source/connectors/kafka-sink-data-formats.txt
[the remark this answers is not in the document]
ccho-mongodb: I don't really know the correct term, but chose the term "literally" to describe a "string literal" rather than to use an expected format to parse out any special instructions (like a schema and payload). I agree that "explicitly" communicates this better though.

PR #549, opened 2019-09-26
Kafka Connector Guide (Feature Branch)

ccho-mongodb:
Original JIRA: https://jira.mongodb.org/browse/DOCSP-6631
Current JIRA (for external review task): https://jira.mongodb.org/browse/DOCSP-7583

Stage: https://docs-mongodbcom-staging.corp.mongodb.com/ecosystem/docsworker/DOCSP-6631-kafka-connector-guide/connectors/kafka.html

PR #561, opened 2019-10-28
DOCSP-6635: Kafka Source Connector

ccho-mongodb:
JIRA: https://jira.mongodb.org/browse/DOCSP-6635
Staging: https://docs-mongodbcom-staging.corp.mongodb.com/ecosystem/docsworker/DOCSP-6635-kafka-connector-source/connectors/kafka-source.html


source/connectors/kafka-source.txt
```diff
@@ -0,0 +1,265 @@
...
+
+* The ``operationType`` field identifies the type of change represented in
+  the change stream document. Possible values include: "insert", "update",
+  "replace", "delete", "invalidate", "drop", "dropDatabase", and "rename".
+
+* The ``fullDocument`` field contents depend on the operation as follows:
+
+  - For insert and replace operations, it contains the new document being
+    inserted or replacing the existing document.
+  - For update operations, it contains the complete document that is being
```
terakilobyte: Interesting choice of wording. Where did you get this from? Not a nit.

source/connectors/kafka-source.txt
```diff
@@ -0,0 +1,265 @@
...
+     - boolean
+     - | Only publish the changed document instead of the full change stream document. Sets the ``change.stream.full.document=updateLookup`` automatically so updated documents will be included.
+
+       .. seealso:: :ref:`Publish full document example <full-document-example>`
+
+       | **Default**: false
+       | **Accepted Values**: ``true`` or ``false``
+
+   * - pipeline
+     - string
```
terakilobyte: shouldn't this be `[]string` or `[string]`?

source/connectors/kafka-source.txt
ccho-mongodb: If you're referring to "some point in time", it is from this:
https://github.com/mongodb/mongo-kafka/blob/master/docs/source.md
Screen Shot 2019-11-01 at 3 00 13 PM

source/connectors/kafka-source.txt
ccho-mongodb: Since these are config files, the convention I adopted was to use "string" for JSON document arrays since there are no explicit data structures. The "Accepted Values" field describes the contents of the field in more detail.

There was a comment a while back in which I was asking a reviewer whether including that field was necessary; based on the current state, the result of that conversation must have been that it was still valid to include.

E.g. 
https://github.com/mongodb/docs-ecosystem/pull/549/files#diff-774db6e64574a30a2870b1d1b33a9638R111
https://github.com/mongodb/docs-ecosystem/pull/549/files#diff-774db6e64574a30a2870b1d1b33a9638R123
...

This is also the convention that was used in the original Kafka Connector docs.
Does that make sense to use, or should we adopt a new convention for config files?

source/connectors/kafka-source.txt
ccho-mongodb: If there's a better way to phrase this or if you think we need to find an external source to provide additional detail, let me know.

source/connectors/kafka-source.txt
terakilobyte: Ok, makes sense. Just wanted to make sure there was intention.

source/connectors/kafka-source.txt
terakilobyte: Nope, those is the best way to phrase it.

PR #569, opened 2019-11-08
DOCSP-7589: Update NodeJS examples in CSFLE to be more idiomatic

ccho-mongodb:
JIRA: https://jira.mongodb.org/browse/DOCSP-7589

Staging: https://docs-mongodbcom-staging.corp.mongodb.com/ecosystem/docsworker/DOCSP-7589-csfle-nodejs-improvements/use-cases/client-side-field-level-encryption-guide.html

### DIFF
--- a/source/includes/steps-fle-create-data-encryption-key.yaml
+++ b/source/includes/steps-fle-create-data-encryption-key.yaml
@@ -30,7 +30,7 @@ content: |
         .. code-block:: javascript
            :emphasize-lines: 5
 
-           const path = "./master-key.txt";
+           const path = './master-key.txt';
            let localMasterKey;
            fs.readFile(path, (err, data) => {
              if (err) throw err;
@@ -141,36 +141,33 @@ content: |
         :tabid: nodejs
 
         .. code-block:: javascript
-           :emphasize-lines: 10,11,13,14,16
 
-           const base64 = require("uuid-base64");
+           const base64 = require('uuid-base64');
 
-           const connectionString = "mongodb://localhost:27017";
-           const keyVaultNamespace = "encryption.__keyVault";
+           const connectionString = 'mongodb://localhost:27017';
+           const keyVaultNamespace = 'encryption.__keyVault';
            const client = new MongoClient(connectionString, {
              useNewUrlParser: true,
              useUnifiedTopology: true,
            });
 
-           client.connect()
-             .then((clientConnection) => {
+           async function main() {
+             try {
+               await client.connect();
                const encryption = new ClientEncryption(client, {
                  keyVaultNamespace,
                  kmsProviders,
                });
-               encryption.createDataKey("local", (err, key) => {
-                 if (err) {
-                   console.error(err);
-                 } else {
-                   const base64DataKeyId = key.toString("base64");
-                   const uuidDataKeyId = base64.decode(base64DataKeyId);
-                   console.log("DataKeyId [UUID]: ", uuidDataKeyId);
-                   console.log("DataKeyId [base64]: ", base64DataKeyId);
-                 }
-               });
-               // ...
-             });
-
+               const key = await encryption.createDataKey('local');
+               const base64DataKeyId = key.toString('base64');
+               const uuidDataKeyId = base64.decode(base64DataKeyId);
+               console.log('DataKeyId [UUID]: ', uuidDataKeyId);
+               console.log('DataKeyId [base64]: ', base64DataKeyId);
+             } finally {
+               await client.close();
+             }
+           }
+           main();
 
         .. note::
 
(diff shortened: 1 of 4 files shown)


ccho-mongodb:
Hi @daprahamian, let me know what you think about the changes I made based on your feedback. Thanks!

source/use-cases/client-side-field-level-encryption-guide.txt
```diff
@@ -277,26 +273,19 @@ To begin development, MedcoMD engineers generate a locally-managed master key:
...
-        wstream.end();
 
-      .. note::
-
-         This code includes a dependency on the ``secure-random``
-         npm package. See the `npmjs documentation on the secure-random
-         package <https://www.npmjs.com/package/secure-random>`_ for
-         installation instructions.
+        try {
+          fs.writeFileSync('master-key.txt', crypto.randomBytes(96));
```
daprahamian: need to require in `crypto` here, under `fs`:

```js
const fs = require('fs');
const crypto = require('crypto');
```

source/includes/steps-fle-create-data-encryption-key.yaml
```diff
@@ -302,28 +300,31 @@ content: |
...
                const query = {
                  _id: base64KeyId,
                };
-               keyColl.findOne(query)
-                 .then((dataKey) => {
-                   console.log(dataKey);
-                 });
-             });
+               const dataKey = await keyColl.findOne(query);
+               constole.log(dataKey);
```
daprahamian: my bad, propagated a typo. `s/constole/console/`

source/includes/steps-fle-create-data-encryption-key.yaml
```diff
@@ -30,7 +30,7 @@ content: |
         .. code-block:: javascript
            :emphasize-lines: 5
 
-           const path = "./master-key.txt";
+           const path = './master-key.txt';
            let localMasterKey;
            fs.readFile(path, (err, data) => {
```
daprahamian: This still needs to be changed to a synchronous version:

```js
const path = './master-key.txt';
const localMasterKey = fs.readFileSync(path);
```

source/use-cases/client-side-field-level-encryption-guide.txt
```diff
@@ -623,23 +612,22 @@ MedcoMD engineers write a function to create a new patient record:
...
-             .then((writeResult) => {
-                 console.log("writeResult:", writeResult);
-             })
-             .catch((writeError) => {
-                 console.log("writeError occurred:", writeError);
-             })
+         async function insertPatient(collection, name, ssn, bloodType, medicalRecords, policyNumber, provider) {
+           try {
+             const writeResult = await collection.insertOne({
+               name: name,
```
daprahamian: nit: these can all be object shorthand:

```js
const writeResult = await collection.insertOne({
  name,
  ssn,
  bloodType,
  medicalRecords,
  insurance: {
    policyNumber,
    provider
  }
});
```

source/use-cases/client-side-field-level-encryption-guide.txt
```diff
@@ -623,23 +612,22 @@ MedcoMD engineers write a function to create a new patient record:
...
+               ssn: ssn,
+               bloodType: bloodType,
+               medicalRecords: medicalRecords,
+               insurance: {
+                 policyNumber: policyNumber,
+                 provider: provider
+               }
+             });
+             console.log('writeResult:', writeResult);
```
daprahamian: do you want the `writeResult` to be returned to the user at all? For example:

```js
const writeResult = await insertPatient(coll, 'dan', '123-456-7890', ...);
```

if so, it should be returned after this `console.log`

source/use-cases/client-side-field-level-encryption-guide.txt
```diff
@@ -623,23 +612,22 @@ MedcoMD engineers write a function to create a new patient record:
...
-                    provider: provider
-                }
-             })
-             .then((writeResult) => {
-                 console.log("writeResult:", writeResult);
-             })
-             .catch((writeError) => {
-                 console.log("writeError occurred:", writeError);
-             })
+         async function insertPatient(collection, name, ssn, bloodType, medicalRecords, policyNumber, provider) {
```
daprahamian: typically, if you have more than 4 parameters with a JS function, you want to switch to using object destructuring here. But that might make it more complicated.

source/includes/steps-fle-create-data-encryption-key.yaml
ccho-mongodb: Haha oops, will fix.

source/use-cases/client-side-field-level-encryption-guide.txt
ccho-mongodb: 👍 will fix

source/use-cases/client-side-field-level-encryption-guide.txt
ccho-mongodb: Agreed, and also identified when code style was linted with Airbnb rules. We had an internal discussion about this and some people preferred the signature that included each field explicitly rather than implicitly within a document. Since this function is specific to the fabricated example and not an exemplar on something specific to MongoDB (such as handling a connection lifecycle, I'll leave the parameters expanded if that's ok with you.

source/use-cases/client-side-field-level-encryption-guide.txt
ccho-mongodb: 👍 will fix

source/use-cases/client-side-field-level-encryption-guide.txt
ccho-mongodb: Whoops, thanks for spotting this. It does not look like we use the result in any other snippet. Will remove.

source/includes/steps-fle-create-data-encryption-key.yaml
ccho-mongodb: Missed this one, thanks for spotting it. Will fix.

ccho-mongodb:
Thanks for the PR review. I made changes and it's now ready for re-review. @daprahamian

PR #572, opened 2019-11-12
DOCSP-7658: Java CSFLE fixes

ccho-mongodb:
JIRA: https://jira.mongodb.org/browse/DOCSP-7658

Staging: https://docs-mongodborg-staging.corp.mongodb.com/ecosystem/docsworker/DOCSP-7658-java-csfle-fixes/use-cases/client-side-field-level-encryption-guide.html

### DIFF
--- a/source/includes/steps-fle-configure-the-mongodb-client.yaml
+++ b/source/includes/steps-fle-configure-the-mongodb-client.yaml
@@ -199,14 +199,14 @@ content: |
            :emphasize-lines: 3-8
 
            MongoClientSettings clientSettings = MongoClientSettings.builder()
-             .applyConnectionString(new ConnectionString("mongodb://localhost:27017"))
-             .autoEncryptionSettings(AutoEncryptionSettings.builder()
-               .keyVaultNamespace(keyVaultNamespace)
-               .kmsProviders(kmsProviders)
-               .schemaMap(schemaMap)
-               .extraOptions(extraOptions)
-               .build())
-             .build();
+               .applyConnectionString(new ConnectionString("mongodb://localhost:27017"))
+               .autoEncryptionSettings(AutoEncryptionSettings.builder()
+                   .keyVaultNamespace(keyVaultNamespace)
+                   .kmsProviders(kmsProviders)
+                   .schemaMap(schemaMap)
+                   .extraOptions(extraOptions)
+                   .build())
+               .build();
 
            MongoClient mongoClient = MongoClients.create(clientSettings);
      .. tab::
--- a/source/includes/steps-fle-configure-the-mongodb-client.yaml
+++ b/source/includes/steps-fle-configure-the-mongodb-client.yaml
@@ -40,7 +40,7 @@ content: |
         :tabid: java-sync
 
         .. code-block:: java
-           :emphasize-lines: 6 
+           :emphasize-lines: 2,5
 
            Map<String, Object> keyMap = new HashMap<String, Object>();
            keyMap.put("key", localMasterKey);
--- a/source/includes/steps-fle-configure-the-mongodb-client.yaml
+++ b/source/includes/steps-fle-configure-the-mongodb-client.yaml
@@ -40,12 +40,14 @@ content: |
         :tabid: java-sync
 
         .. code-block:: java
-           :emphasize-lines: 2,3
+           :emphasize-lines: 6 
+
+           Map<String, Object> keyMap = new HashMap<String, Object>() {{
+               put("key", localMasterKey);
+           }};
 
            Map<String, Map<String, Object>> kmsProviders = new HashMap<String, Map<String, Object>>() {{
-             put("local", new HashMap<String, Object>() {{
-               put("key", localMasterKey);
-             }});
+               put("local", keyMap);
            }};
      .. tab::
         :tabid: nodejs
@@ -124,7 +126,7 @@ content: |
         .. code-block:: java
            :emphasize-lines: 2
 
-           final Map<String, Object> extraOptions = new HashMap<String, Object>() {{
+           Map<String, Object> extraOptions = new HashMap<String, Object>() {{
              put("mongocryptdSpawnPath", "/usr/local/bin/mongocryptd");
            }};
 
--- a/source/includes/steps-fle-configure-the-mongodb-client.yaml
+++ b/source/includes/steps-fle-configure-the-mongodb-client.yaml
@@ -87,7 +87,7 @@ content: |
            :emphasize-lines: 2
 
            HashMap<String, BsonDocument> schemaMap = new HashMap<String, BsonDocument>() {{
-             put("medicalRecords.patients", BsonDocument.parse(jsonSchema));
+               put("medicalRecords.patients", BsonDocument.parse(jsonSchema));
            }}
      .. tab::
         :tabid: nodejs
@@ -125,7 +125,7 @@ content: |
            :emphasize-lines: 2
 
            Map<String, Object> extraOptions = new HashMap<String, Object>() {{
-             put("mongocryptdSpawnPath", "/usr/local/bin/mongocryptd");
+               put("mongocryptdSpawnPath", "/usr/local/bin/mongocryptd");
            }};
 
         .. admonition:: Encryption Binary Daemon
--- a/source/includes/steps-fle-configure-the-mongodb-client.yaml
+++ b/source/includes/steps-fle-configure-the-mongodb-client.yaml
@@ -86,9 +86,8 @@ content: |
         .. code-block:: java
            :emphasize-lines: 2
 
-           HashMap<String, BsonDocument> schemaMap = new HashMap<String, BsonDocument>() {{
-               put("medicalRecords.patients", BsonDocument.parse(jsonSchema));
-           }}
+           HashMap<String, BsonDocument> schemaMap = new HashMap<String, BsonDocument>();
+           schemaMap.put("medicalRecords.patients", BsonDocument.parse(jsonSchema));
      .. tab::
         :tabid: nodejs
 
@@ -124,9 +123,8 @@ content: |
         .. code-block:: java
            :emphasize-lines: 2
 
-           Map<String, Object> extraOptions = new HashMap<String, Object>() {{
-               put("mongocryptdSpawnPath", "/usr/local/bin/mongocryptd");
-           }};
+           Map<String, Object> extraOptions = new HashMap<String, Object>();
+           extraOptions.put("mongocryptdSpawnPath", "/usr/local/bin/mongocryptd");
 
         .. admonition:: Encryption Binary Daemon
            :class: note
--- a/source/includes/steps-fle-configure-the-mongodb-client.yaml
+++ b/source/includes/steps-fle-configure-the-mongodb-client.yaml
@@ -42,13 +42,11 @@ content: |
         .. code-block:: java
            :emphasize-lines: 6 
 
-           Map<String, Object> keyMap = new HashMap<String, Object>() {{
-               put("key", localMasterKey);
-           }};
+           Map<String, Object> keyMap = new HashMap<String, Object>();
+           keyMap.put("key", localMasterKey);
 
-           Map<String, Map<String, Object>> kmsProviders = new HashMap<String, Map<String, Object>>() {{
-               put("local", keyMap);
-           }};
+           Map<String, Map<String, Object>> kmsProviders = new HashMap<String, Map<String, Object>>();
+           kmsProviders.put("local", keyMap);
      .. tab::
         :tabid: nodejs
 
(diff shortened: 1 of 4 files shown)


source/includes/steps-fle-configure-the-mongodb-client.yaml
```diff
@@ -87,7 +87,7 @@ content: |
            :emphasize-lines: 2
 
            HashMap<String, BsonDocument> schemaMap = new HashMap<String, BsonDocument>() {{
-             put("medicalRecords.patients", BsonDocument.parse(jsonSchema));
+               put("medicalRecords.patients", BsonDocument.parse(jsonSchema));
```
jyemin: Same here: avoid the anonymous class creation

source/includes/steps-fle-configure-the-mongodb-client.yaml
```diff
@@ -124,8 +124,8 @@ content: |
         .. code-block:: java
            :emphasize-lines: 2
 
-           final Map<String, Object> extraOptions = new HashMap<String, Object>() {{
-             put("mongocryptdSpawnPath", "/usr/local/bin/mongocryptd");
+           Map<String, Object> extraOptions = new HashMap<String, Object>() {{
+               put("mongocryptdSpawnPath", "/usr/local/bin/mongocryptd");
```
jyemin: And here

source/includes/steps-fle-convert-to-a-remote-master-key.yaml
```diff
@@ -52,22 +52,22 @@ content: |
            final BsonString awsAccessKeyId = new BsonString("<IAM User Access Key ID>");
            final BsonString awsSecretAccessKey = new BsonString("<IAM User Secret Access Key>");
            Map<String, Map<String, Object>> kmsProviders = new HashMap<String, Map<String, Object>>() {{
-             put("aws", new HashMap<String, Object>() {{
-               put("region", masterKeyRegion);
-               put("accessKeyId", awsAccessKeyId);
-               put("secretAccessKey", awsSecretAccessKey);
-             }});
+               put("aws", new HashMap<String, Object>() {{
```
jyemin: And here.  Once we make Java 9 the minimum we could use `java.util.Map#of()`, but most of our users are still on Java 8

source/includes/steps-fle-convert-to-a-remote-master-key.yaml
```diff
@@ -113,8 +113,8 @@ content: |
            final BsonString masterKeyRegion = new BsonString("<Master Key AWS Region>"); // e.g. "us-east-2"
            final BsonString masterKeyArn = new BsonString("<Master Key ARN>"); // e.g. "arn:aws:kms:us-east-2:111122223333:alias/test-key"
            DataKeyOptions dataKeyOptions = new DataKeyOptions().masterKey(new BsonDocument() {{
-             put("region", masterKeyRegion);
-             put("key", masterKeyArn);
+               put("region", masterKeyRegion);
```
jyemin: For BsonDocument, you can use `append` to avoid this ugliness

source/includes/steps-fle-create-data-encryption-key.yaml
```diff
@@ -11,19 +11,17 @@ content: |
...
-           byte[] fileBytes = new byte[96];
+           byte[] localMasterKey= new byte[96];
 
            try (FileInputStream fis = new FileInputStream(path)) {
-             fileBytes = fis.readAllBytes();
-           } catch (IOException e) {
-             e.printStackTrace();
+               fis.readNBytes(localMasterKey, 0, 96);
+           } catch (Exception e) {
+               e.printStackTrace();
```
jyemin: When you put all the examples together, this catch is problematic, as it swallows the error and proceeds to the next line.  Consider removing the catch clause altogether.

source/includes/steps-fle-create-data-encryption-key.yaml
```diff
@@ -11,19 +11,17 @@ content: |
...
            String path = "master-key.txt";
 
-           byte[] fileBytes = new byte[96];
+           byte[] localMasterKey= new byte[96];
 
            try (FileInputStream fis = new FileInputStream(path)) {
-             fileBytes = fis.readAllBytes();
-           } catch (IOException e) {
-             e.printStackTrace();
+               fis.readNBytes(localMasterKey, 0, 96);
```
jyemin: If you want to be really careful, you could check the return value and throw if 96 bytes were not read.

source/use-cases/client-side-field-level-encryption-guide.txt
```diff
@@ -251,17 +251,17 @@ To begin development, MedcoMD engineers generate a locally-managed master key:
...
+                new SecureRandom().nextBytes(localMasterKey);
 
-            try (FileOutputStream stream = new FileOutputStream("master-key.txt")) {
-              stream.write(localMasterKey);
-            } catch (IOException e)  {
-              e.printStackTrace();
+                try (FileOutputStream stream = new FileOutputStream("master-key.txt")) {
+                    stream.write(localMasterKey);
+                } catch (IOException e) {
+                    e.printStackTrace();
```
jyemin: Same comment as above.  Consider removing the catch clause

source/includes/steps-fle-configure-the-mongodb-client.yaml
ccho-mongodb: 👍 Got it, wasn't sure if you were only referring to the examples you previously identified or in general. Will update.

source/includes/steps-fle-configure-the-mongodb-client.yaml
ccho-mongodb: Will fix.

source/includes/steps-fle-convert-to-a-remote-master-key.yaml
ccho-mongodb: Will fix.

source/includes/steps-fle-convert-to-a-remote-master-key.yaml
ccho-mongodb: Thanks, will fix.

source/includes/steps-fle-create-data-encryption-key.yaml
ccho-mongodb: Would it compile without `catch` and no `finally`? I was getting lint errors in Eclipse running JDK 1.8.

source/includes/steps-fle-create-data-encryption-key.yaml
ccho-mongodb: Sure, that sounds like a good idea -- would you recommend throwing an Exception if it's <96 bytes?

I just discovered `readNBytes()` isn't in Java 1.8, so I will probably switch to `read()`
1.8: https://docs.oracle.com/javase/8/docs/api/java/io/InputStream.html
1.9 https://docs.oracle.com/javase/9/docs/api/java/io/InputStream.html

source/includes/steps-fle-create-data-encryption-key.yaml
jyemin: Hmm, haven't tried with java 8, but you shouldn't need either with try-with-resources

source/includes/steps-fle-create-data-encryption-key.yaml
jyemin: read is problematic because it can return before reading the requested number of bytes.  So you have to call it in a loop.  That's why these newer methods were added.  It's a PITA

ccho-mongodb: Since the 3.11 MongoDB driver is marked as compatible with Java 6, what would be your recommendation on reading the file into a byte array? Perhaps this snippet? https://stackoverflow.com/a/38235210

source/includes/steps-fle-create-data-encryption-key.yaml
ccho-mongodb: Looks like it was introduced in 1.7, but since the driver is compatible with 1.6, would you recommend I use it anyway? Or should I find another way to structure this block?

jyemin: We required Java 8 for FLE so consider Java 8 as the minimum you need to be compatible with.

source/includes/steps-fle-convert-to-a-remote-master-key.yaml
```diff
@@ -112,10 +112,9 @@ content: |
 
            final BsonString masterKeyRegion = new BsonString("<Master Key AWS Region>"); // e.g. "us-east-2"
            final BsonString masterKeyArn = new BsonString("<Master Key ARN>"); // e.g. "arn:aws:kms:us-east-2:111122223333:alias/test-key"
-           DataKeyOptions dataKeyOptions = new DataKeyOptions().masterKey(new BsonDocument() {{
-               put("region", masterKeyRegion);
-               put("key", masterKeyArn);
-           }})
+           DataKeyOptions dataKeyOptions = new DataKeyOptions().masterKey(new BsonDocument();
```
jyemin: Make sure you try to compile all the example.  This one doesn't compile.  I think you meant:

        DataKeyOptions dataKeyOptions = new DataKeyOptions().masterKey(
                new BsonDocument()
                        .append("region", masterKeyRegion)
                        .append("key", masterKeyArn));

source/includes/steps-fle-convert-to-a-remote-master-key.yaml
```diff
@@ -112,10 +112,9 @@ content: |
 
            final BsonString masterKeyRegion = new BsonString("<Master Key AWS Region>"); // e.g. "us-east-2"
```
jyemin: Remember, no `final`.

ccho-mongodb: Ah sorry, missed this file. Thanks for catching that, will fix all instances.

source/includes/steps-fle-convert-to-a-remote-master-key.yaml
ccho-mongodb: Oops, thanks for catching that. 😬 Will fix.

source/includes/steps-fle-create-data-encryption-key.yaml
ccho-mongodb: Good to know, will add that as an explicit requirement in the guide.

ccho-mongodb: I still receive the same compile errors whether setting compliance to 1.8 or 12. Am I doing something incorrectly?

```
Unhandled exception type IOException thrown by automatic close() invocation on fis
	Unhandled exception type FileNotFoundException
	Unhandled exception type IOException
```

jyemin: That's expected.  Just add `IOException` to the throws clause of the containing method.

source/includes/steps-fle-create-data-encryption-key.yaml
jyemin: There's just no good way to one-liner that does this before Java 9.  Not sure what to say

ccho-mongodb: Ok. I'll keep it as `readNBytes` and add an admonition that mentions that if they are using 8, they will need to implement that line differently, and link to stackoverflow or (hopefully) another resource.

source/includes/steps-fle-create-data-encryption-key.yaml
ccho-mongodb: The containing method in the context of the code samples is `main`.  Even if it was in another function, there would need to be a catch statement somewhere. If I added a `System.exit(1)` call after the stacktrace, would that at least meet the requirement of halting on an exception there?

jyemin: This is legal:

    public static void main(String[] args) throws IOException

ccho-mongodb: Ah, you got me there. Will update.

source/use-cases/client-side-field-level-encryption-guide.txt
ccho-mongodb: Will fix

ccho-mongodb:
Ready for re-review @jyemin

source/use-cases/client-side-field-level-encryption-guide.txt
```diff
@@ -181,8 +181,21 @@ Additional Dependencies
...
+
+         * - `JDK 8 or later
+             <https://www.oracle.com/technetwork/java/javase/downloads/index.html>`_
+           - While the current driver is compatible with older versions of
+             the JDK, the CSFLE feature is only compatible with JDK 8
+             and later.
 
+         * - :java-docs-latest:`libmongocrypt <driver/tutorials/client-side-encryption/>`
+           - The `libmongocrypt` library contains bindings to communicate
+             with the process that manages the encryption.
```
jyemin: the process => the native library

ccho-mongodb: Will fix, thanks!

PR #573, opened 2019-11-13
DO NOT MERGE: DOCSP- 6184: Scram & x509 snippets

mongomoe:
JIRA
-----
* https://jira.mongodb.org/browse/DOCSP-6184
* https://jira.mongodb.org/browse/DOCSP-7588


source/includes/connection-snippets/x509-snippets/java-connection.java
```diff
@@ -0,0 +1,25 @@
...
+System.setProperty("javax.net.ssl.keyStore", "/etc/certs/mongodb/client.keystore");
+System.setProperty("javax.net.ssl.keyStorePassword", "mypass");
+String subject = "<your_subject>"; // CN=ChrisChoClient,OU=TestClientCertificateOrgUnit,O=TestClientCertificateOrg,L=TestClientCertificateLocality,ST=TestClientCertificateStateC=US
+MongoCredential credential = MongoCredential.createMongoX509Credential(subject); 
+MongoClientSettings settings = MongoClientSettings.builder()
+        .credential(credential)
+        .applyToClusterSettings(builder -> builder.hosts(Arrays.asList(new ServerAddress("localhost"))))
+        .applyToSslSettings(builder -> builder.enabled(true)).build();
+MongoClient client = MongoClients.create(settings);
+// end x509 connection
```
nathan-contino-mongo: I feel like we should at least put these things in a file that can be run -- this won't run at all because there isn't a main method and it isn't valid java, for instance. Because of the way async methods work in java (and C# and C++, I believe) it's a little weird to omit those details too.

source/includes/connection-snippets/x509-snippets/csharp-connection.cs
```diff
@@ -0,0 +1,30 @@
...
+using MongoDB.Driver;
+using MongoDB.Driver.Core;
+using System;
+using System.Threading.Tasks;
+
+using System.Security.Cryptography.X509Certificates;
+using System.Security.Cryptography;
+using System.Security.Permissions;
+
+var settingObjectOnlySettings = new MongoClientSettings {
```
nathan-contino-mongo: This object is missing its closing brace. (should come after the ``Server = ... `` line)

source/includes/connection-snippets/x509-snippets/c-connection.c
```diff
@@ -0,0 +1,37 @@
...
+};
+
+memcpy( & ssl_opts, ssl_default, sizeof ssl_opts);
+
+ssl_opts.pem_file = "/etc/certs/mongodb/client.pem";
+ssl_opts.ca_file = "/etc/certs/mongodb/ca.pem";
+ssl_opts.weak_cert_validation = false;
+ssl_opts.allow_invalid_hostname = false;
+
+char subject[] = "mongodb://<your_subject>"; // mongodb://CN%3DChris%2COU%3DTestClientCertificateOrgUnit%2CO%3DEducationClientCertificate%2CL%3DTestClientCertificateLocality%2CST%3DTestClientCertificateState%2CC%3DUS
```
ccho-mongodb: Feel free to remove the comment from this line in this repo.

source/includes/connection-snippets/x509-snippets/c-connection.c
```diff
@@ -0,0 +1,37 @@
...
+
+char uri[] = strcat(subject, "@localmongo1/?authMechanism=MONGODB-X509")
```
ccho-mongodb: Should we use `localmongo1` as the host? Perhaps there's a better standard for atlas urls?

source/includes/connection-snippets/x509-snippets/c-connection.c
ccho-mongodb: If this is for Atlas, we should probably use `mongodb+srv://`
https://www.mongodb.com/blog/post/mongodb-3-6-here-to-SRV-you-with-easier-replica-set-connections

source/includes/connection-snippets/x509-snippets/ruby-connection.rb
```diff
@@ -0,0 +1,19 @@
...
+  ssl: true,
+  ssl_cert: '/etc/certs/mongodb/client.pem',
+  ssl_key: '/etc/certs/mongodb/client.pem',
+  ssl_ca_cert: '/etc/certs/mongodb/ca.pem',
+
+)
+
+puts Thread.current.backtrace
+rescue StandardError => e
+puts e.backtrace
```
ccho-mongodb: Nitpick: needs an indentation. Consider running it through `rubocop` linter.

source/includes/connection-snippets/x509-snippets/ruby-connection.rb
```diff
@@ -0,0 +1,19 @@
...
+  ['localmongo1:27017'],
+  auth_mech:: mongodb_x509,
+  ssl: true,
+  ssl_cert: '/etc/certs/mongodb/client.pem',
+  ssl_key: '/etc/certs/mongodb/client.pem',
+  ssl_ca_cert: '/etc/certs/mongodb/ca.pem',
+
+)
+
+puts Thread.current.backtrace
```
ccho-mongodb: This was debug code, please remove

source/includes/connection-snippets/x509-snippets/ruby-connection.rb
```diff
@@ -0,0 +1,19 @@
...
+
+begin
+client = Mongo::Client.new(
+  ['localmongo1:27017'],
+  auth_mech:: mongodb_x509,
+  ssl: true,
+  ssl_cert: '/etc/certs/mongodb/client.pem',
+  ssl_key: '/etc/certs/mongodb/client.pem',
+  ssl_ca_cert: '/etc/certs/mongodb/ca.pem',
+
```
ccho-mongodb: Nitpick: no need for a linebreak here

source/includes/connection-snippets/x509-snippets/python-connection.py
```diff
@@ -0,0 +1,12 @@
+#
+begin x509 connection
```
ccho-mongodb: Did you mean for this to be on a separate line or should it be part of line 1?

source/includes/connection-snippets/scram-snippets/php-connection.php
```diff
@@ -0,0 +1,10 @@
+<!-- begin scram connection -->
+<?php
+
+$client = new MongoDB\Client(
```
ccho-mongodb: Nitpick: I think usually you indent within a php block

source/includes/connection-snippets/x509-snippets/c-connection.c
```diff
@@ -0,0 +1,37 @@
...
+bool retval;
+
+mongoc_init();
+
+const mongoc_ssl_opt_t * ssl_default = mongoc_ssl_opt_get_default();
+mongoc_ssl_opt_t ssl_opts = {
+  0
+};
+
+memcpy( & ssl_opts, ssl_default, sizeof ssl_opts);
```
ccho-mongodb: Nitpick: I think there are a couple extra spaces here. Should probably be:
```
memcpy(&ssl_opts, ssl_default, sizeof ssl_opts);
```

nathan-contino-mongo:
W.R.T. @ccho-mongodb 's last point: java and csharp will happily accept null for the username field when you create a new x509 credential (which is the only one required). If we can just use that for many of these examples we can cut down on a LOT of user errors. Depending on how these are laid out I would also suggest linking to this manual page: https://docs.mongodb.com/manual/tutorial/configure-x509-client-authentication/#add-x-509-certificate-subject-as-a-user
...because it contains an openssl snippet that seems to be the only way to extract a subject from a certificate in such a way that it actually works as the mongodb username via many of the drivers. There are a lot of ways to extract the subject so pointing people here explicitly would be VERY helpful (I know it would have saved me a good 10 hours of my life)

source/includes/connection-snippets/x509-snippets/java-connection.java
mongomoe: I agree that it is a little weird, the password auth implementations currently on the driver docs don't have wrapper classes for the most part (it is a little in-consistent though). See:
https://docs.mongodb.com/ecosystem/drivers/cxx/

@terakilobyte suggested have these only having the client connection code without main methods or classes to wrap the code.

source/includes/connection-snippets/x509-snippets/csharp-connection.cs
mongomoe: Good catch, fixed!

source/includes/connection-snippets/x509-snippets/c-connection.c
mongomoe: Do you think people would be confused about what "<your_subject>" means without the comment? @ccho-mongodb

source/includes/connection-snippets/x509-snippets/c-connection.c
mongomoe: how about `mongodb+srv://<cluster-url>`? @ccho-mongodb

source/includes/connection-snippets/x509-snippets/c-connection.c
ccho-mongodb: You could use:
"x509 client certificate subject" and also comment that it should be urlencoded and provide a link to a sample. If no samples exist, maybe you could create a comment line that contains a sample (the current values are not great -- I think openssl provides some default values when you create a certificate which you could use instead)

source/includes/connection-snippets/x509-snippets/c-connection.c
ccho-mongodb: Sure, whatever you're able to get driver teams to agree to. However, I think you meant 
```
mongodb+srv://<your x509 client subject>@<cluster-url>/?authMechanism=MONGODB-X509")
```

source/includes/connection-snippets/x509-snippets/c-connection.c
nathan-contino-mongo: The username is also unnecessary since it's included in the certificate already. If you omit it when connecting to mongodb it'll just assign you a username matching the subject. Why introduce another point of failure for users?

source/includes/connection-snippets/x509-snippets/java-connection.java
mongomoe: Okay, returned back to the original state where there is a main method. The new code is basically the same as chris's original x509 repo. Resolving this discussion.

source/includes/connection-snippets/x509-snippets/c-connection.c
mongomoe: I've altered this snippet to more closely match the others. Let me know what you think @nathan-contino-mongo

source/includes/connection-snippets/x509-snippets/c-connection.c
mongomoe: Fixed. Thanks!

source/includes/connection-snippets/x509-snippets/c-connection.c
mongomoe: Fixed. Thanks!

source/includes/connection-snippets/x509-snippets/csharp-connection.cs
```diff
@@ -0,0 +1,30 @@
...
+  Credential = MongoCredential.CreateMongoX509Credential(null),
+   SslSettings = new SslSettings {
+    ClientCertificates = new List < X509Certificate > () {
+     new X509Certificate2("client-certificate.pfx", "<your password>")
```
nathan-contino-mongo: probably a good idea to change this from "<your password>" to "<pfx password>"

source/includes/connection-snippets/x509-snippets/csharp-connection.cs
```diff
@@ -0,0 +1,30 @@
...
+    },
+   },
+   UseTls = true,
+   Server = new MongoServerAddress("localhost", 27017)
+
+  var client = new MongoClient(settingObjectOnlySettings);
+  // end x509 connection
```
nathan-contino-mongo: also, not a valid file until it actually has a class/main method

mongomoe: Hey nate, I think you're commenting on an outdated commit

mongomoe: checkout https://github.com/mongodb/docs-ecosystem/pull/573/files for latest versions @nathan-contino-mongo

source/includes/connection-snippets/x509-snippets/java-connection.java
```diff
@@ -0,0 +1,25 @@
...
+import com.mongodb.client.MongoClients;
+import com.mongodb.client.MongoIterable;
+
+
+System.setProperty("javax.net.ssl.trustStore","/etc/certs/mongodb/client.truststore");
+System.setProperty("javax.net.ssl.trustStorePassword","mypass");
+
+System.setProperty("javax.net.ssl.keyStore", "/etc/certs/mongodb/client.keystore");
+System.setProperty("javax.net.ssl.keyStorePassword", "mypass");
+String subject = "<your_subject>"; // CN=ChrisChoClient,OU=TestClientCertificateOrgUnit,O=TestClientCertificateOrg,L=TestClientCertificateLocality,ST=TestClientCertificateStateC=US
```
nathan-contino-mongo: Definitely need to remove this example subject comment from all of the samples. If you really want to put a comment in there, I'd suggest linking to the manual page I mentioned below where it actually describes how to extract your subject in a way that will allow a connection (there are many alternative ways to extract the subject s.t. mongod will reject it)

source/includes/connection-snippets/x509-snippets/csharp-connection.cs
```diff
@@ -0,0 +1,53 @@
...
+    {
+        static void Main(string[] args)
+        {
+            MainAsync().Wait();
+            Console.WriteLine("done");
+        }
+
+        static async Task MainAsync()
+        {
+            var settingObjectOnlySettings = new MongoClientSettings
```
nathan-contino-mongo: Probably should rename this to something like "settings"

source/includes/connection-snippets/x509-snippets/csharp-connection.cs
```diff
@@ -0,0 +1,53 @@
...
+using System;
+using System.Threading.Tasks;
+
+using System.Security.Cryptography.X509Certificates;
+using System.Security.Cryptography;
+using System.Security.Permissions;
+
+namespace WorkingWithMongoDB
+{
+    public class Entity
```
nathan-contino-mongo: Delete this class. No need for an Entity class here (it's not used anywhere I can see) and you can always use the default BsonDocument class instead for simple operations.

source/includes/connection-snippets/x509-snippets/csharp-connection.cs
```diff
@@ -0,0 +1,53 @@
...
+        static async Task MainAsync()
+        {
+            var settingObjectOnlySettings = new MongoClientSettings 
+            {
+                Credential =  MongoCredential.CreateMongoX509Credential(null),
+                SslSettings = new SslSettings
+                {
+                    ClientCertificates = new List<X509Certificate>()
+                    {
+                        new X509Certificate2("/etc/certs/mongodb/client-certificate.pfx", "<your_password>")
```
nathan-contino-mongo: same thoughts as before -> change your_password to pfx_passphrase to make it clear what you need to put in here

source/includes/connection-snippets/x509-snippets/nodejs-connection.js
```diff
@@ -0,0 +1,21 @@
+// begin x509 connection
+const MongoClient = require('mongodb').MongoClient;
+const fs = require('fs');
+
+const cert = fs.readFileSync("/etc/certs/mongodb/client.pem");
+const key = fs.readFileSync("/etc/certs/mongodb/client.pem");
```
nathan-contino-mongo: Out of curiosity is it really necessary to read in cert and key? Why not just use them both under a different name?

source/includes/connection-snippets/x509-snippets/scala-connection.scala
```diff
@@ -0,0 +1,33 @@
...
+import com.mongodb.client.model.FindOneAndUpdateOptions
+
+object MongoDBx509 extends App {
+   System.setProperty("javax.net.ssl.trustStore","client.truststore");
+   System.setProperty("javax.net.ssl.trustStorePassword","<your_password>");
+
+   System.setProperty("javax.net.ssl.keyStore", "client.keystore");
+   System.setProperty("javax.net.ssl.keyStorePassword", "<your_password>");
+
+   // null value will force mongoDB server to use DN from cert
```
nathan-contino-mongo: should change DN to Subject for consistency here

source/includes/connection-snippets/x509-snippets/scala-connection.scala
```diff
@@ -0,0 +1,33 @@
...
+      .applyToClusterSettings(
+         (builder: ClusterSettings.Builder)
+         => builder.hosts(List(new ServerAddress("<cluster-url>")).asJava))
+      .applyToSslSettings(
+         (builder : SslSettings.Builder)
+         => builder.enabled(true))
+      .build();
+
+   val client : MongoClient = MongoClient(settings);
+   val db : MongoDatabase = client.getDatabase("test")
```
nathan-contino-mongo: I noticed that sometimes we connect to "test" database, sometimes we connect to "test" database collection "collection", sometimes something else... I'd advise that we make this consistent everywhere.

nathan-contino-mongo:
Thought: the way these snippets are structured it's going to mislead people in testing because they will copy/paste this code, run it, and it won't throw any errors until they actually interact with a collection. (because the drivers don't actually connect until you execute a command or issue a query)

Snippets are currently inconsistent as far as database/collection connections are concerned.

Would it be a good idea to just findOne from the test/collection collection so that every driver is forced to actually connect and try to execute something?

Otherwise this is going to expose the lazy connection aspect of the drivers really quickly in a really confusing way for users (because the connection, if it doesn't actually execute anything, will ALWAYS appear to work no matter what settings they use and how valid their certs are)

source/includes/connection-snippets/x509-snippets/ruby-connection.rb
mongomoe: Fixed - ran through rubocop. Thanks!

source/includes/connection-snippets/x509-snippets/ruby-connection.rb
mongomoe: Fixed - removed debug code

source/includes/connection-snippets/x509-snippets/ruby-connection.rb
mongomoe: Fixed. Thanks!

source/includes/connection-snippets/x509-snippets/python-connection.py
mongomoe: I did not mean this to be a separate line. Fixed, thanks for pointing it out

source/includes/connection-snippets/scram-snippets/php-connection.php
mongomoe: Fixed - indentation. Thanks!

source/includes/connection-snippets/x509-snippets/c-connection.c
mongomoe: Fixed. Thanks!

source/includes/connection-snippets/x509-snippets/c-connection.c
mongomoe: Removed chris's x509 cert subject with the comment `// x509 client certificate subject`

source/includes/connection-snippets/x509-snippets/java-connection.java
mongomoe: Fixed - replaced with `// x509 client certificate subject`

source/includes/connection-snippets/x509-snippets/csharp-connection.cs
mongomoe: @nathan-contino-mongo can you send me a snippet of how you want me to replace this, I copied and pasted this snippet almost verbatim from chris's repo and don't have my enviroment set up for csharp, so I'm hesitant on altering code like this without being able to run it

nathan-contino-mongo: The entire snippet within: "public class Entity{...}" (including the class declaration and the curly braces) should be deleted. It isn't used and it isn't necessary :)

source/includes/connection-snippets/x509-snippets/csharp-connection.cs
mongomoe: Fixed

source/includes/connection-snippets/x509-snippets/csharp-connection.cs
mongomoe: Fixed - thanks!

source/includes/connection-snippets/x509-snippets/nodejs-connection.js
mongomoe: It is not necessary, I don't particularly like that we're reading this same file twice - but this is consistent w/ the node-driver example for x509 (see: https://mongodb.github.io/node-mongodb-native/3.3/tutorials/connect/ssl/ > heading: "Validate Server Certificate and Present Valid Certificate")

source/includes/connection-snippets/x509-snippets/scala-connection.scala
mongomoe: Fixed

source/includes/connection-snippets/x509-snippets/csharp-connection.cs
mongomoe: Fixed

source/includes/connection-snippets/x509-snippets/scala-connection.scala
mongomoe: Fair point - I've removed any variables referencing `db` from these snippets to make it consistent and since including `db` is out of scope for this.

mongomoe:
# Response to nate's top-level comments:

Question/Comment
----------

> Thought: the way these snippets are structured it's going to mislead people in testing because they will copy/paste this code, run it, and it won't throw any errors until they actually interact with a collection. (because the drivers don't actually connect until you execute a command or issue a query)
> 
> Snippets are currently inconsistent as far as database/collection connections are concerned.
> 
> Would it be a good idea to just findOne from the test/collection collection so that every driver is forced to actually connect and try to execute something?
> 
> Otherwise this is going to expose the lazy connection aspect of the drivers really quickly in a really confusing way for users (because the connection, if it doesn't actually execute anything, will ALWAYS appear to work no matter what settings they use and how valid their certs are)

Answer
----------
* I don't think these snippets will mislead people, the scope of this ticket was to create client connections with x509, not to include inserts. Also, other than the nodejs-driver I believe all drivers do actually connect without inserts (though I may be wrong about this). Another way users can test if they actually connected is by checking their `mongod` instance has received a connection.
* As for snippets being inconsistent for db/collections, now this is fixed. The snippets now just show the client connection.
* Adding a findOne would mean there would have to be documents in the test db. Also, that's quite a bit of scope creep because it would mean adding the findOne statements for every driver snippet and testing out all the code (and having the environment set up for each snippet).

Thanks for the detailed comment and review, good ideas all together, but I think it misses the context of this ticket - this ticket is just supposed to show the code snippet for client connection, not be a doc on how to connect via x509.

source/includes/connection-snippets/scram-snippets/java-connection.java
```diff
@@ -0,0 +1,19 @@
...
+import com.mongodb.client.MongoDatabase;
+import com.mongodb.MongoClientSettings;
+
+// ...
+ConnectionString connString = new ConnectionString(
+    "mongodb+srv://<username>:<password>@<cluster-address>/test?w=majority"
+);
+MongoClientSettings settings = MongoClientSettings.builder()
+    .applyConnectionString(connString)
+    .retryWrites(true)
```
jyemin: Why do you not use the connection string for this, as with other drivers?  Did it not work?

source/includes/connection-snippets/x509-snippets/java-connection.java
```diff
@@ -0,0 +1,37 @@
...
+import com.mongodb.ServerAddress;
+import com.mongodb.client.MongoClient;
+import com.mongodb.client.MongoClients;
+import com.mongodb.client.MongoIterable;
+
+public class X509Connection {
+
+    public static void main(String args[]) {
+
+        System.setProperty("javax.net.ssl.trustStore", "/etc/certs/mongodb/v3/client.truststore");
```
jyemin: Normally applications don't need their own trust store, unless they are self-signing the server certificate.  For Atlas, you shouldn't need these two lines

source/includes/connection-snippets/x509-snippets/java-connection.java
```diff
@@ -0,0 +1,37 @@
...
+        System.setProperty("javax.net.ssl.trustStorePassword", "<your_password>");
+
+        System.setProperty("javax.net.ssl.keyStore", "/etc/certs/mongodb/v3/client.keystore");
+        System.setProperty("javax.net.ssl.keyStorePassword", "<your_password>");
+
+        try {
+            String subject = "<your_subject>"; // x509 client certificate subject
```
jyemin: Why aren't you using a connection string? Were you unable to get it to work?

source/includes/connection-snippets/scram-snippets/c-connection.c
```diff
@@ -0,0 +1,24 @@
...
+
+int
+main (int argc, char *argv[])
+{
+   mongoc_database_t *database;
+   mongoc_client_t *client;
+
+   mongoc_init ();
+
+   client = mongoc_client_new(
```
ccho-mongodb: The spacing for function calls seems inconsistent. This line does not include a space before the opening parens, but 5, 15, 17, 18, and 20 do. Should this one include a space too?

source/includes/connection-snippets/scram-snippets/java-connection.java
```diff
@@ -0,0 +1,19 @@
...
+    .build();
+MongoClient mongoClient = MongoClients.create(settings);
+MongoDatabase database = mongoClient.getDatabase("test");
+// begin scram connection
```
ccho-mongodb: Should this include a `.close()` and the comment then read ```// end scram connection``` or is this one meant to be different?

source/includes/connection-snippets/scram-snippets/csharp-connection.cs
```diff
@@ -0,0 +1,9 @@
+// begin scram connection
+using MongoDB.Bson;
+using MongoDB.Driver;
+// ...
+var client = new MongoClient(
+    "mongodb+srv://<username>:<password>@<cluster-address>/test?w=majority"
```
ccho-mongodb: Nitpick: sometimes the connection string includes `cluster-url` and others it includes `cluster-address`. Is this on purpose and if so, what is the reason?

source/includes/connection-snippets/scram-snippets/nodejs-connection.js
```diff
@@ -0,0 +1,10 @@
+// begin scram connection
+const MongoClient = require('mongodb').MongoClient;
+const uri = "mongodb+srv://<username>:<password>@<your-cluster-url>/test?retryWrites=true&w=majority";
+const client = new MongoClient(uri, { useNewUrlParser: true });
+client.connect(err => {
+  const collection = client.db("test").collection("devices");
```
ccho-mongodb: The other SCRAM driver snippets only retrieve a reference to the database. Could you make sure that they are consistent in which reference they store?

source/includes/connection-snippets/scram-snippets/ruby-connection.rb
```diff
@@ -0,0 +1,5 @@
+# begin scram connection
+require 'mongo'
+
+client = Mongo::Client.new('mongodb+srv://kay:myRealPassword@cluster0.mongodb.net/?ssl=true&w=majority')
```
ccho-mongodb: The user/password/cluster placeholder `<>` used in the other drivers is missing here. Could you make this one consistent?

source/includes/connection-snippets/x509-snippets/go-connection.go
```diff
@@ -0,0 +1,33 @@
...
+    Age  int
+    City string
+}
+
+func main() {
+	ctx := context.TODO()
+	caFilePath := "/etc/certs/mongodb/ca.pem"
+	certificateKeyFilePath := "/etc/certs/mongodb/client.pem"
+
+	uri := "mongodb+srv://<your-x509-client-subject>@<cluster-url>/?tlsCAFile=%s&tlsCertificateKeyFile=%s&authMechanism=MONGODB-X509"
```
divjotarora: I don't think the `<your-x509-client-subject>` part is necessary because we should read the username from the client cert automatically.

source/includes/connection-snippets/x509-snippets/go-connection.go
```diff
@@ -0,0 +1,33 @@
...
+	uri = fmt.Sprintf(uri, caFilePath, certificateKeyFilePath)
+	clientOpts := options.Client().ApplyURI(uri).SetServerSelectionTimeout(2 * time.Second)
```
divjotarora: Why is `SetServerSelectionTimeout` called here? I don't see it being set in the Java and Node examples.

source/includes/connection-snippets/x509-snippets/csharp-connection.cs
```diff
@@ -0,0 +1,48 @@
...
+                    {
+                        new X509Certificate2("/etc/certs/mongodb/client-certificate.pfx", "<pfx_passphrase>")
+                    },
+                },
+                UseTls = true,
+                Server = new MongoServerAddress("<cluster-url>")
+            };
+
+            var client = new MongoClient(settings);
+        }
```
rstam: It might be better to use the sample code from the recently revised C# documentation for X509. It's not yet live, but you can see the changes made in github.

In particular, let's show starting from a connection string and adding in via code the things that can only be added via code:

https://github.com/mongodb/mongo-csharp-driver/commit/ee9f21e423ffb02e8c42c5c7192b352403800499#diff-c2ca42c0112df581679f2ede5989ee0aR83

source/includes/connection-snippets/scram-snippets/nodejs-connection.js
```diff
@@ -0,0 +1,10 @@
+// begin scram connection
```
mbroadst: for our recent documentation work we've been using a more complex template using async/await:
```
const { MongoClient } = require('mongodb');

const client = new MongoClient(
  'mongodb+srv://<username>:<password>@<your-cluster-url>/test?retryWrites=true&w=majority'
);

async function run() {
  try {
    await client.connect();
    const collection = client.db('test').collection('devices');
    // perform actions on the collection object
  } finally {
    // Ensures that the client will close when you finish/error
    await client.close();
  }
}

run().catch(console.dir);
```

source/includes/connection-snippets/x509-snippets/nodejs-connection.js
```diff
@@ -0,0 +1,20 @@
+// begin x509 connection
+const MongoClient = require('mongodb').MongoClient;
```
mbroadst: per the style above, here's what this looks like now:

```
const { MongoClient } = require('mongodb');
const fs = require('fs');

const client = new MongoClient('mongodb+srv://<cluster-url>?authMechanism=MONGODB-X509&ssl=true', {
  sslCA: fs.readFileSync('/etc/certs/mongodb/ca.pem'),
  sslKey: fs.readFileSync('/etc/certs/mongodb/client.pem'),
  sslCert: fs.readFileSync('/etc/certs/mongodb/client.pem')
});

async function run() {
  try {
    await client.connect();
    const collection = client.db('test').collection('devices');
    // perform actions on the collection object
  } finally {
    // Ensures that the client will close when you finish/error
    await client.close();
  }
}

run().catch(console.dir);
```

source/includes/connection-snippets/x509-snippets/python-connection.py
```diff
@@ -0,0 +1,12 @@
+#begin x509 connection
+import ssl
+from pymongo
+import MongoClient
```
ccho-mongodb: I'm not familiar with formatting style -- did you mean for a single import to extend to two lines?

source/includes/connection-snippets/x509-snippets/ruby-connection.rb
```diff
@@ -0,0 +1,14 @@
...
+begin
+  client = Mongo::Client.new(
+    ['localmongo1:27017'],
+    auth_mech. mongodb_x509,
+    ssl: true,
+    ssl_cert: '/etc/certs/mongodb/client.pem',
+    ssl_key: '/etc/certs/mongodb/client.pem',
+    ssl_ca_cert: '/etc/certs/mongodb/ca.pem'
+  )
```
p-mongo: We documented how to enable x509 for local testing in https://github.com/mongodb/mongo-ruby-driver/blob/master/spec/README.md#x509-authentication. Unless creation of the client using Ruby options is specifically desired, I think using the URI format is a better idea.

source/includes/connection-snippets/x509-snippets/ruby-connection.rb
```diff
@@ -0,0 +1,14 @@
+# begin x509 connection
+require 'mongo'
+
+begin
```
p-mongo: The begin/end statements in this snippet are unnecessary as they do not do anything.

source/includes/connection-snippets/scram-snippets/ruby-connection.rb
p-mongo: Per https://github.com/mongodb/specifications/blob/master/source/initial-dns-seedlist-discovery/initial-dns-seedlist-discovery.rst#seedlist-discovery `ssl=true` should be the default, and thus should not be needed.

source/includes/connection-snippets/x509-snippets/python-connection.py
```diff
@@ -0,0 +1,12 @@
...
+
+uri = "mongodb+srv://<cluster-url>/?authMechanism=MONGODB-X509"
+client = MongoClient(uri,
```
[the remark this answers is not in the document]
p-mongo: +1 on using tls option names.

source/includes/connection-snippets/scram-snippets/java-connection.java
mongomoe: @ccho-mongodb might be able to shed some light on this, since this was his original code that I've adapted for this repo

source/includes/connection-snippets/scram-snippets/nodejs-connection.js
mongomoe: The scram snippets are actually verabitim copied and pasted from their driver docs. But I can fix them to be more consistent

source/includes/connection-snippets/x509-snippets/java-connection.java
mongomoe: Removed the two lines related to the trustStore

source/includes/connection-snippets/x509-snippets/go-connection.go
mongomoe: Cool, thanks for the info. I've removed it from the uri

source/includes/connection-snippets/x509-snippets/go-connection.go
mongomoe: Sorry about that, you're right `SetServerSelectionTimeout` doesn't need to be in here. I've removed it now.

source/includes/connection-snippets/x509-snippets/java-connection.java
nathan-contino-mongo: This should also apply to scala.

source/includes/connection-snippets/x509-snippets/nodejs-connection.js
mongomoe: Fixed - replaced my node snippet almost verbatim with the one you provided above ^ the only thing I changed was the removal of the line ` const collection = client.db('test').collection('devices');` to make it consistent w/ the other examples in this repo. This ticket only entails client connection, not specifying a db or collection. Let me know if that's fine with you @mbroadst

source/includes/connection-snippets/x509-snippets/java-connection.java
mongomoe: Removed the two lines related to the trustStore for scala as well @nathan-contino-mongo

source/includes/connection-snippets/x509-snippets/java-connection.java
mongomoe: @ccho-mongodb might be able to shed some light on this, since this was his original code that I've adapted for this repo

source/includes/connection-snippets/scram-snippets/java-connection.java
mongomoe: For the SCRAM connection snippets, these were taken verbatim from the drivers docs. For instance: https://docs.mongodb.com/ecosystem/drivers/java/#connect-to-mongodb-atlas

rozza: Lets just use the connection string as a string and pass it to `MongoClients.create()` it will greatly simplify the creation of the client.

source/includes/connection-snippets/scram-snippets/scala-connection.scala
```diff
@@ -0,0 +1,8 @@
+// begin scram connection
+import org.mongodb.scala._
+
+val uri: String = "mongodb+srv://<username>:<password>@<cluster-address>/test?retryWrites=true&w=majority"
+System.setProperty("org.mongodb.async.type", "netty")
```
rozza: Remove the system property - netty is no longer a requirement and system properties for configuration of the driver settings are being removed in the near future.

rozza: Could use `streamType=netty` in the uri string.

source/includes/connection-snippets/x509-snippets/java-connection.java
ccho-mongodb: @mongomoe It seems related to Jeff's other comment here: `source/includes/connection-snippets/scram-snippets/java-connection.java`. I think the request was to try using a connection string (e.g. similar to other examples such as the golang one) rather than using the builder pattern.

source/includes/connection-snippets/scram-snippets/java-connection.java
mongomoe: Thanks for the suggestion - I've updated the snippet

source/includes/connection-snippets/scram-snippets/nodejs-connection.js
mongomoe: Fixed - similar to my comment below: copied basically verbatim and removed the `const collection = client.db('test').collection('devices');` line since it's not needed for this ticket

source/includes/connection-snippets/scram-snippets/java-connection.java
```diff
@@ -0,0 +1,7 @@
...
```
jyemin: Let's be consistent about two things

1. Whether a full program is included.  This is a snippet, but the X509 example is a complete program.  Looking at examples for other languages, seems like complete programs are favored.
2. (Nit) Whether there is a separate local variable for the URI string.  Here there is not, but in the X509 example there is.  Looking at examples from other languages, seems like _not_ having a separate variable for the connection string is favored.

source/includes/connection-snippets/scram-snippets/csharp-connection.cs
mongomoe: Fixed. All switched to `<cluster-url>`

source/includes/connection-snippets/x509-snippets/python-connection.py
mongomoe: Typo - fixed in latest commit

source/includes/connection-snippets/x509-snippets/python-connection.py
```diff
@@ -0,0 +1,11 @@
+#begin x509 connection
+import ssl 
+from pymongo import MongoClient
+
+uri = "mongodb+srv://<cluster-url>/?authMechanism=MONGODB-X509"
+client = MongoClient(uri,
+  ssl = True,
+  ssl_certfile = '/etc/certs/mongodb/client.pem',
```
prashantmital: This still needs to be updated to use the equivalent `tls*` option names.

source/includes/connection-snippets/scram-snippets/php-connection.php
```diff
@@ -0,0 +1,10 @@
...
+    'mongodb+srv://<username>:<password>@<cluster-url>/test?retryWrites=true&w=majority'
+);
+
+$db = $client->test;
```
jmikola: Do these examples acknowledge the fact that [driver client objects do no IO in their constructor](https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst#clients-do-no-i-o-in-the-constructor)? In most of these examples, the only IO performed will be the initial DNS lookup for the SRV/TXT record. Selecting a database or collection object from the client is no enough to initiate SDAM and perform an auth handshake with the server. This means that users could run many of these examples and SCRAM would not actually be exercised at all.

An easy, universal way to ensure the SCRAM handshake _would_ happen may be to execute a `ping` command using the newly constructed clients, or perhaps insert a test document.

source/includes/connection-snippets/scram-snippets/c-connection.c
```diff
@@ -0,0 +1,25 @@
+// begin scram connection
```
jmikola: Is the purpose of all of these examples to _not_ specify the `authMechanism`? IIRC, modern drivers can infer whether to use SCRAM-SHA-1 or SCRAM-SHA-256 for a particular user, since 256 was introduced in MongoDB 4.0, so perhaps this was intentional.

source/includes/connection-snippets/x509-snippets/c-connection.c
```diff
@@ -0,0 +1,30 @@
...
+
+int
+main (int argc, char *argv[])
+{
+   mongoc_database_t *database;
+   mongoc_client_t *client;
+
+   mongoc_init ();
+
+   const mongoc_ssl_opt_t *ssl_default = mongoc_ssl_opt_get_default ();
```
jmikola: I assume this code was inspired by the [Connecting to a server over SSL](http://mongoc.org/libmongoc/current/advanced-connections.html#connecting-to-a-server-over-ssl) docs, as the [X509 examples in the libmongoc docs](http://mongoc.org/libmongoc/current/authentication.html#x-509-certificate-authentication) don't use `mongoc_ssl_opt_get_default` at all.

AFAIK, the only purpose of copying in `mongoc_ssl_opt_get_default ()` instead of starting with totally empty SSL opts is that we make it possible to pick up a custom CA file/directory that might have been configured when libmongoc was originally compiled. The X509 example overrides the CA _file_, but I suppose there may still be a use case for inheriting some default CA _directory_. No harm in keeping this as-is, I just wanted to example what's going on.

cc: @bazile-clyde @kevinAlbs

source/includes/connection-snippets/x509-snippets/c-connection.c
```diff
@@ -0,0 +1,30 @@
...
+   mongoc_ssl_opt_t ssl_opts = { 0 };
+
+   memcpy(&ssl_opts, ssl_default, sizeof ssl_opts);
```
jmikola: Per earlier comments about coding style (https://github.com/mongodb/docs-ecosystem/pull/573/files#r347552954), you may want a space between `memcpy` and the opening paren here.

source/includes/connection-snippets/x509-snippets/c-connection.c
```diff
@@ -0,0 +1,30 @@
...
+
+   ssl_opts.pem_file = "/etc/certs/mongodb/client.pem";
+   ssl_opts.ca_file = "/etc/certs/mongodb/ca.pem";
+   ssl_opts.weak_cert_validation = false;
+   ssl_opts.allow_invalid_hostname = false;
```
jmikola: These `false` assignments are redundant because the `ssl_opts` struct was initialized with zero values for all fields (i.e. `{ 0 }` initializing struct).

source/includes/connection-snippets/x509-snippets/c-connection.c
```diff
@@ -0,0 +1,30 @@
...
+
+   client = mongoc_client_new ("mongodb+srv://<your-x509-client-subject>@<cluster-url>/?authMechanism=MONGODB-X509");
```
jmikola: Comparing this with other drivers, I think we don't even need to use a `mongoc_ssl_opt_t` struct in this example. The C driver supports the same `tlsCAFile` and `tlsCertificateKeyFile` options used in the Go example. Given that, I think this example file can be drastically reduced and more closely resemble the SCRAM example with the connection string changed.

In fact, every 4.2-compat driver should support those options per the [Unified URI options spec](https://github.com/mongodb/specifications/blob/master/source/uri-options/uri-options.rst). You can see the list of compatible drivers by looking at the language tickets linked from https://jira.mongodb.org/browse/DRIVERS-557.

----

@bazile-clyde @kevinAlbs: Since http://mongoc.org/libmongoc/current/authentication.html#x-509-certificate-authentication still demonstrates using `mongoc_ssl_opt_t`, can one of you open a ticket to track updating that to use the new unified options? I think it's sensible to fix it for the 1.15 docs and later, but we could at least do so in the master branch for 1.16+.

source/includes/connection-snippets/x509-snippets/cpp-connection.cpp
```diff
@@ -0,0 +1,25 @@
...
+  mongocxx::instance inst{};
+
+  mongocxx::options::ssl ssl_opts{};
+  ssl_opts.ca_file("/etc/certs/mongodb/ca.pem");
+  ssl_opts.pem_file("/etc/certs/mongodb/client.pem");
+
+  mongocxx::options::client client_opts{};
+  client_opts.ssl_opts(ssl_opts);
+
+  mongocxx::uri uri("mongodb+srv://<your-x509-client-subject>@<cluster-url>/?authMechanism=MONGODB-X509&ssl=true");
```
jmikola: Note that C++ doesn't yet support the Unified URI options spec (per https://jira.mongodb.org/browse/DRIVERS-557), so you're correct to use `ssl_opts` here.

I do think that `ssl=true` may be redundant in the URI, though. In libmongoc, which C++ uses, SSL is enabled for a client when SSL opts are assigned, even if the connection string did not specify `ssl=true` or `tls=true` on its own.

cc: @bazile-clyde @kevinAlbs

source/includes/connection-snippets/x509-snippets/csharp-connection.cs
```diff
@@ -0,0 +1,48 @@
+// begin x509 connection
```
jmikola: I think @jyemin already pointed this out, but the corresponding SCRAM example for C# is a snippet, while this appears to be a complete source file.

source/includes/connection-snippets/x509-snippets/java-connection.java
```diff
@@ -0,0 +1,18 @@
+// begin x509 connection
+import com.mongodb.client.MongoClient;
+import com.mongodb.client.MongoClients;
+
+public class X509Connection {
+
+public static void main(String args[]) {
+        // Set system properties via commandline or programmatically
+        System.setProperty("javax.net.ssl.keyStore", "/etc/certs/mongodb/v3/client.keystore");
+        System.setProperty("javax.net.ssl.keyStorePassword", "<your_password>");
```
jmikola: @jyemin should this use tls URI options instead (per [DRIVERS-557](https://jira.mongodb.org/browse/DRIVERS-557))?

source/includes/connection-snippets/x509-snippets/nodejs-connection.js
```diff
@@ -0,0 +1,22 @@
+// begin x509 connection
+const { MongoClient } = require('mongodb');
+const fs = require('fs');
+
+const client = new MongoClient('mongodb+srv://<cluster-url>?authMechanism=MONGODB-X509&ssl=true', {
+  sslCA: fs.readFileSync('/etc/certs/mongodb/ca.pem'),
+  sslKey: fs.readFileSync('/etc/certs/mongodb/client.pem'),
+  sslCert: fs.readFileSync('/etc/certs/mongodb/client.pem')
```
jmikola: @mbroadst should this use tls URI options instead (per [DRIVERS-557](https://jira.mongodb.org/browse/DRIVERS-557))?

source/includes/connection-snippets/x509-snippets/php-connection.php
```diff
@@ -0,0 +1,13 @@
+<!-- Begin x509 connection -->
```
jmikola: Same comments as I left on the SCRAM PHP example:

 * Don't use HTML comments
 * Start the source file with `<?php` and omit the final `?>
 * Use PHP comment syntax within the `<?php` tag

Also, I'll note that this file is using the PHP `'mongodb` extension (our low-level API) directly instead of the PHP library used in the SCRAM example. I think you should be consistent and use the library code, as it's what we have for most all PHP documentation examples.

So, construct a `MongoDB\Client` instead of a Manager object.

source/includes/connection-snippets/x509-snippets/php-connection.php
```diff
@@ -0,0 +1,13 @@
...
+<?php
+$subject = "mongodb+srv://<your_subject>"; 
+$manager = new MongoDB\Driver\Manager(
+    $subject + "@<cluster-url>/?ssl=true&authMechanism=MONGODB-X509",
+  [],
+  [
+    "pem_file" => "/etc/certs/mongodb/v3/client.pem",
+    "ca_file" => "/etc/certs/mongodb/v3/ca.pem",
+  ]
```
jmikola: Similar to my comment on the C example, this can use the `tlsCAfile` and `tlsCertificateKeyFile` URI options instead of the special driver options array being used here (per [DRIVERS-557](https://jira.mongodb.org/browse/DRIVERS-557)). That works for the 4.2-compat version of the PHP driver (extension 1.6.0+ and library 1.5.0+), and would let you only need to provide a URI string (first argument).

Similar to my other comments in the SCRAM PHP example, you don't do anything after creating the client object here so you should be aware that no IO is happening. A `ping` command would avoid that.

source/includes/connection-snippets/x509-snippets/php-connection.php
```diff
@@ -0,0 +1,13 @@
+<!-- Begin x509 connection -->
+<?php
+$subject = "mongodb+srv://<your_subject>";
```
jmikola: This is technically the start of the URI and not just the subject. I don't agree with the variable naming here and think you'd be better off just to include this in the URI string below (as you've already embedded "<cluster-url>" there.

source/includes/connection-snippets/x509-snippets/php-connection.php
```diff
@@ -0,0 +1,13 @@
+<!-- Begin x509 connection -->
+<?php
+$subject = "mongodb+srv://<your_subject>"; 
+$manager = new MongoDB\Driver\Manager(
+    $subject + "@<cluster-url>/?ssl=true&authMechanism=MONGODB-X509",
```
jmikola: `+` is not a string concat operator in PHP. This should be `.` if so; however, I don't think you actually need to store the first part of the URI in a `$subject` variable.

source/includes/connection-snippets/x509-snippets/c-connection.c
kevinAlbs: Currently I think `const mongoc_ssl_opt_t *ssl_default = mongoc_ssl_opt_get_default()` is equivalent to `mongoc_ssl_opt_t ssl_default = {0}`.

But there's an old TODO here: https://github.com/mongodb/mongo-c-driver/blob/1.15.0/src/libmongoc/src/mongoc/mongoc-ssl.c/#L37-L44 indicating we could change the behavior of `mongoc_ssl_opt_get_default`. I'm not sure how useful that would be, or if that's something we really plan to do though.

source/includes/connection-snippets/x509-snippets/c-connection.c
kevinAlbs: +1 for using  the URI options instead  of `mongoc_ssl_opt_t`.

Created a ticket to update docs, and to consider deprecating use of that struct entirely:
https://jira.mongodb.org/browse/CDRIVER-3436

source/includes/connection-snippets/x509-snippets/cpp-connection.cpp
kevinAlbs: Agreed, `ssl=true` or `tls=true` is redundant if setting via `ssl_opts`.

source/includes/connection-snippets/scram-snippets/c-connection.c
mongomoe: This snippet was actually taken verbatim from https://docs.mongodb.com/ecosystem/drivers/c/#connect-to-mongodb-atlas (all the SCRAM connection snippets were taken verbatim from the current docs). I can't speak to why the original developer who made this snippet didn't explicitly specify `authMechanism` but the default is SCRAM-SHA-1 and you can specify to use SCRAM-SHA-256 if needed. 

However, I don't see the need to specify this explicitly for these snippets, and I would prefer these snippets to be as consistent to the adjacent driver docs as possible.

source/includes/connection-snippets/scram-snippets/c-connection.c
```diff
@@ -0,0 +1,25 @@
...
+int
+main (int argc, char *argv[])
+{
+   mongoc_database_t *database;
+   mongoc_client_t *client;
+
+   mongoc_init ();
+
+   client = mongoc_client_new(
+      "mongodb+srv://<username>:<password>@<cluster-url>/test?retryWrites=true&w=majority"
```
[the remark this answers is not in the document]
mongomoe: The SCRAM snippets are taken verbatim from the driver docs (see: https://docs.mongodb.com/ecosystem/drivers/c/#connect-to-mongodb-atlas)

The atlas UI does specify to use `/test`. I personally would prefer to stay consistent with the atlas UI, but if you and the other leads are against it, I can change the snippets. 

Screen Shot 2019-11-20 at 6 24 16 PM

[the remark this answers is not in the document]
mongomoe: +1 Nathan for his thoughts on this:

@terakilobyte

source/includes/connection-snippets/scram-snippets/cpp-connection.cpp
```diff
@@ -0,0 +1,12 @@
+// begin scram connection
+#include <mongocxx/client.hpp>
+#include <mongocxx/instance.hpp>
+
+mongocxx::instance inst{}; 
+mongocxx::client conn{
+   mongocxx::uri{
+      "mongodb+srv://<username>:<password>@<cluster-url>/test?retryWrites=true&w=majority"
```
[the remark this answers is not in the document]
mongomoe: The atlas UI suggests using w=majority, see pic below: 

Screen Shot 2019-11-20 at 6 28 12 PM

I would prefer to stay consistent with the UI but don't have a strong preference specifically around including w=majority or not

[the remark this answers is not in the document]
mongomoe: +1 nathan @terakilobyte

source/includes/connection-snippets/scram-snippets/cpp-connection.cpp
```diff
@@ -0,0 +1,12 @@
...
+   }
+};
+mongocxx::database db = conn["test"];
+// end scram connection
```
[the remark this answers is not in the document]
mongomoe: Fair point, I've removed all newlines at the eof for all these snippets now.

source/includes/connection-snippets/scram-snippets/java-connection.java
mongomoe: Hey Jeff, 

to address your two points:
* SCRAM Connection snippets use a snippet since they were taken pretty much verbatim from the driver docs (See:  https://docs.mongodb.com/ecosystem/drivers/java/#connect-to-mongodb-atlas). On the other hand, X509 snippets are full runnable programs. I'm open to adjusting the SCRAM snippets to be full runnable programs but that would mean SCRAM snippets aren't consistent with the driver docs for connecting to atlas (which might be a necessary evil anyway). I would prefer to keep x509 code as full runnable code. 
* You might be referring to an old version of this PR, but now the code snippets use <cluster-url>. Let me know if this is satisfactory

source/includes/connection-snippets/scram-snippets/php-connection.php
```diff
@@ -0,0 +1,10 @@
+<!-- begin scram connection -->
```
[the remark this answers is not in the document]
mongomoe: Fixed

source/includes/connection-snippets/scram-snippets/php-connection.php
```diff
@@ -0,0 +1,10 @@
...
+<?php
+
+$client = new MongoDB\Client(
+    'mongodb+srv://<username>:<password>@<cluster-url>/test?retryWrites=true&w=majority'
+);
+
+$db = $client->test;
+?>
```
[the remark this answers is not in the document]
mongomoe: Fixed

source/includes/connection-snippets/scram-snippets/php-connection.php
```diff
@@ -0,0 +1,10 @@
...
+<!-- end scram connection -->
```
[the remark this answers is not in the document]
mongomoe: Fixed

source/includes/connection-snippets/scram-snippets/scala-connection.scala
mongomoe: Fixed - removed system property & updated uri to streamType=netty

source/includes/connection-snippets/x509-snippets/ruby-connection.rb
```diff
@@ -0,0 +1,14 @@
...
+  client = Mongo::Client.new(
+    ['<cluster-url>'],
+    auth_mech. mongodb_x509,
+    ssl: true,
```
prashantmital: @p-mongo shouldn't this be updated to use tls* option names as well?

source/includes/connection-snippets/scram-snippets/cpp-connection.cpp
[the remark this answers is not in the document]
jmikola: I wasn't aware Atlas was now suggesting this. I don't feel strongly either way (other than thinking the extra options might distract from the auth-focus here), but I believe the X509 examples weren't including "w=majority" so you probably want to enforce consistency throughout the PR if that's still the case.

source/includes/connection-snippets/x509-snippets/c-connection.c
jmikola: Ah, I see that `MONGOC_SSL_DEFAULT_TRUST_FILE` and `MONGOC_SSL_DEFAULT_TRUST_DIR` are always defined as `NULL` if undefined at compile time. And I found no references to them in CMake so I assume they are simply unused and reserved for future use.

@kevinAlbs: In that case, do you suppose it makes sense to omit `mongoc_ssl_opt_get_default ()` from this example entirely?

source/includes/connection-snippets/scram-snippets/c-connection.c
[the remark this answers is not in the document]
terakilobyte: I'm fine with it having "test" in the uri. It doesn't impact auth, it's just the default database to connect to. Specifying anything else is equally valid, and so not an item we should spend much time fussing over imo.

source/includes/connection-snippets/scram-snippets/cpp-connection.cpp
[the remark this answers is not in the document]
terakilobyte: I'd like to mimic the existing Atlas UI as much as possible. The idea is these connection snippets will eventually be the ones used in Atlas, on the driver pages, and more. `w=majority` is a nice, sane default that benefits copypasta users. Is there a good reason to not choose this as the default?

source/includes/connection-snippets/x509-snippets/ruby-connection.rb
terakilobyte: From a developer perspective, I think most users would rather use an options structure. Their IDEs will give code hints to help prevent typos and offer suggestions, and API line documentation if it exists. 

Why is the URI format a better idea and different from the other drivers?

source/includes/connection-snippets/x509-snippets/ruby-connection.rb
p-mongo: Yes the tls names would be preferred.

source/includes/connection-snippets/x509-snippets/c-connection.c
kevinAlbs: Per this thread  https://github.com/mongodb/docs-ecosystem/pull/573/files/c5875fa5382cf3e2ccbf27cc2d8843d3359031e1#r348478333
I'd prefer removing the reference to mongoc_ssl_opt_t entirely and use the URI opts instead.

source/includes/connection-snippets/x509-snippets/csharp-connection.cs
mongomoe: Sure, we're going to update the SCRAM snippets shortly  to be complete source files

source/includes/connection-snippets/x509-snippets/csharp-connection.cs
mongomoe: Sure we've now updated the code to be very similar to that link. Let me know what you think

source/includes/connection-snippets/x509-snippets/csharp-connection.cs
```diff
@@ -0,0 +1,51 @@
...
+    class Program
+    {
+        static void Main(string[] args)
+        {
+            MainAsync().Wait();
+        }
+
+        static async Task MainAsync()
+        {
+            var connectionString = "mongodb+srv://<cluster-url>/test?authSource=$externalretryWrites=true&w=majority&authMechanism=MONGODB-X509";
```
rstam: Is `authSource=$external` actually required?

There seems to be a separator missing between the `authSource` value and the `retryWrites` keyword.

mongomoe: * `authSource=$external` is technically not needed for csharp, but it is needed for other drivers, and for these snippets, we would ideally be as consistent amongst drivers as possible
* Updated the uri to include the separator

source/includes/connection-snippets/scram-snippets/c-connection.c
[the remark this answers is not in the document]
jmikola: Sorry to nit on this, but I want to ensure we're all on the same page.

> It doesn't impact auth

This is only the case if the SRV/TXT response specifies an "authSource" option, which I believe is the case for Atlas but not generally. In the absence of "authSource" (in either the SRV/TXT response or inline connection string option), this _would_ impact auth my instructing drivers to authenticate against the "test" database.

> it's just the default database to connect to

I'm not aware of any drivers that have a notion of a "default database" (I cc-ed @mbroadst and @jyemin to confirm). This applies to the shell and perhaps to some higher level libraries (e.g. Mongoose ODM).

source/includes/connection-snippets/scram-snippets/cpp-connection.cpp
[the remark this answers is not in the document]
jmikola: Per my previous comment, I'm OK including it if you want to be consistent with the Atlas UI. In that case, I think my point about the X509 examples being inconsistent is outstanding. Perhaps you want to add `w=majority` to the X509 URIs.

This is related to my previous comment about `/<db-name>` appearing in the SCRAM examples but not X509.

source/includes/connection-snippets/scram-snippets/php-connection.php
jmikola: @mongomoe poke, in case you missed this comment.

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+    "ruby.rubocop.executePath": "/usr/local/bin/rubocop"
+}
```
jmikola: I don't think this file should have been committed. It may be wise to include your IDE's files in your personal `.gitignore` to avoid this in the future.

source/includes/connection-snippets/x509-snippets/c-connection.c
jmikola: @kevinAlbs To speed this along, would you mind modifying this file to the preferred form and sending that over to @mongomoe so he can copy it into this PR?

source/includes/connection-snippets/x509-snippets/cpp-connection.cpp
jmikola: @mongomoe let's remove `ssl=true` here, and then I think this can be resolved. Doing so would also be consistent with the C driver example, since that connection string omits an `ssl` or `tls` option as well.

source/includes/connection-snippets/scram-snippets/c-connection.c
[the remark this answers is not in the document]
jyemin: IMO we should not include a default database in our examples.  It's meaningless in most drivers except as a non-canonical alternative to authSource.

source/includes/connection-snippets/scram-snippets/php-connection.php
```diff
@@ -0,0 +1,8 @@
+<?php
+/* Begin x509 connection */
```
jmikola: This is a SCRAM example, not X509.

source/includes/connection-snippets/scram-snippets/php-connection.php
```diff
@@ -0,0 +1,8 @@
...
+$client = new MongoDB\Client(
+    'mongodb+srv://<username>:<password>@<cluster-url>/test?retryWrites=true&w=majority'
+);
+
+$db = $client->test;
+/* End x509 connection */
```
jmikola: This is a SCRAM example, not X509.

source/includes/connection-snippets/x509-snippets/perl-connection.pl
```diff
@@ -0,0 +1,14 @@
+# begin x509 connection
+use IO::Socket::SSL;
+use MongoDB;
+
+my $client = MongoDB::MongoClient->new(
+    host => "mongodb+srv://<cluster-url>",
```
jmikola: Is this missing the X509 subject? Connection strings for other language include that. /cc @xdg

xdg: For 3.4+, we shouldn't need to give the subject as username, as it can be picked out of the ticket.  And drivers can (as the Perl driver does) pick it out of the ticket and send it for earlier servers.  See https://github.com/mongodb/specifications/blob/master/source/auth/auth.rst#mongodb-x509

source/includes/connection-snippets/x509-snippets/php-connection.php
```diff
@@ -0,0 +1,12 @@
...
+  [],
+  [
+    "pem_file" => "/etc/certs/mongodb/v3/client.pem",
+    "ca_file" => "/etc/certs/mongodb/v3/ca.pem",
+  ]
+);
+/* End x509 connection */
```
jmikola: I propose the following source for this file:

```php
/* Begin x509 connection */
$client = new MongoDB\Client(
    'mongodb+srv://<your_subject>@<cluster-url>/?authMechanism=MONGODB-X509',
    [
        'tlsCAfile' => '/etc/certs/mongodb/v3/ca.pem',
        'tlsCertificateKeyFile' => '/etc/certs/mongodb/v3/client.pem',
    ]
);
/* End x509 connection */
```

Note the following:

 * I replaced usage of the extension (i.e. MongoDB\Driver\Manager) with the library (i.e. MongoDB\Client). This is consistent with what you were doing in the SCRAM example. For all intents and purposes, the constructors take and support the same arguments. The library's MongoDB\Client class passes its constructor arguments directly to the MongoDB\Driver\Manager class, which is creates internally. The MongoDB\Client supports some extra options in the third argument, but we're not using it in any of these examples and it's not relevant to auth at all.
 * I replaced the deprecated driver options (third constructor argument) with [unified URI options](https://github.com/mongodb/specifications/blob/master/source/uri-options/uri-options.rst) in the second argument, which I also mentioned in some other comments for other language examples. Since we're now using URI options, it's certainly possible to stuff _everything_ into a single connection string (i.e. first argument), but I broke it out in the interest of readability. When constructing a client, the PHP driver basically merges any options in the second constructor parameter on top of the connection string (i.e. second parameter options take precedence). The third parameter is used for driver options and is handled entirely separately (different keys and rules for merging), but it's no longer relevant for any of these examples if we use can use unified URI options. 
 * I removed `ssl=true` from the URI, since the current version of the driver implicitly enables TLS when one or more TLS-related options are specified (we're using two here). For an explanation of this, you can see the [extension's changlog](https://www.php.net/manual/en/mongodb-driver-manager.construct.php#refsect1-mongodb-driver-manager.construct-changelog).

source/includes/connection-snippets/x509-snippets/ruby-connection.rb
jmikola: @p-mongo to clarify, does Ruby's client constructor solicit URI options outside of the connection string using the same keys as would be used in the connection string? This came up in the PHP example, where @mongomoe was passing the CA file and cert via a constructor parameter with different names for these options in an associative array (since deprecated), and I suggested using another constructor parameter that used the unified URI option names (also an associative array). There, I opted _not_ to stuff everything into a single connection string for the sake of readability. See: https://github.com/mongodb/docs-ecosystem/pull/573/files#r350349009

I don't know if Ruby allows specifying the unified URI options via the same option names anywhere beyond the connection string (i.e. first parameter), but I want to make sure you and @terakilobyte are on the same page there.

p-mongo: No, in Ruby URI options only apply when they are given in the URI. When given as Ruby hash keys the option names can be entirely different.

I don't know if I personally care which syntax is used - we support both - I am only under the impression that in general the URI syntax is favored these days.

p-mongo: Also, almost no developers initialize Ruby client programmatically. Almost all applications use Mongoid. Therefore while the given code example illustrates which Ruby options can be set and what they are called, it's not something that 95% of users will directly copy/paste into their configuration (they would have to convert this to Mongoid yaml syntax). A URI is more easily copy pastable.

source/includes/connection-snippets/scram-snippets/c-connection.c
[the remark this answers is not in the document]
jmikola: I just remembered that I have an outstanding request to remove the `/<db-name>` component from connection strings in Atlas in [MMS-5936](https://jira.mongodb.org/browse/MMS-5936).

source/includes/connection-snippets/x509-snippets/nodejs-connection.js
mongomoe: The latest version of the node-docs use `ssl` (see: https://mongodb.github.io/node-mongodb-native/3.3/reference/connecting/connection-settings/)

mongomoe:
Closing this PR because I'm going to split this up into smaller PRs for each driver for x509 code

source/includes/connection-snippets/x509-snippets/perl-connection.pl
jmikola: Ah, in that case it was probably a mistake for other driver examples in this PR to include a subject. Thanks for the clarification!

source/includes/connection-snippets/x509-snippets/cpp-connection.cpp
kevinAlbs: Apologies, `ssl=true` is actually not redundant because the C++ driver has an [extra validation check](https://github.com/mongodb/mongo-cxx-driver/blob/fe4b1dcc2b9b4402b7f7b8bb0202bbc8a227ccf1/src/mongocxx/client.cpp#L70-L74)  which throws an exception if attempting to set SSL options without having `ssl=true` in the URI.
I'll note in the [new review](https://github.com/mongodb/docs-ecosystem/pull/583/files)

PR #574, opened 2019-11-14
DOCSP-7006 fix admonition, add link to server-side encryption schema

ccho-mongodb:
JIRA: https://jira.mongodb.org/browse/DOCSP-7006

Staging: https://docs-mongodbcom-staging.corp.mongodb.com/ecosystem/docsworker/DOCSP-7006-csfle-link-server-json-schema/use-cases/client-side-field-level-encryption-guide.html

### DIFF
--- a/source/use-cases/client-side-field-level-encryption-guide.txt
+++ b/source/use-cases/client-side-field-level-encryption-guide.txt
@@ -394,7 +394,6 @@ standard requires the following information:
    see :manual:`Enforce Field Level Encryption Schema
    </core/security-client-side-encryption/#enforce-field-level-encryption-schema>`.
 
-
 The MedcoMD engineers receive specific requirements for the fields of
 data and their encryption strategies. The following table illustrates
 the data model of the Medco Management System.
@@ -475,8 +474,6 @@ upon deterministic encryption based on the following properties:
         }
     }
 
-
-
 Blood Type
 ++++++++++
 
@@ -495,7 +492,6 @@ random encryption based on the following properties:
             "algorithm": "AEAD_AES_256_CBC_HMAC_SHA_512-Random"
         }
     }
-
 
 Medical Records
 +++++++++++++++
@@ -544,7 +540,6 @@ decide upon deterministic encryption based on the following properties:
         }
     }
 
-
 Recap
 +++++
 
@@ -552,7 +547,6 @@ MedcoMD engineers created a JSON Schema that satisfies their
 requirements of making sensitive data queryable and secure. View the
 full `JSON Schema for the Medco Medical Management System
 <https://raw.githubusercontent.com/mongodb/docs-assets/DOCSP-json-schema-helper-and-json/MedcoMDSchema.json>`_.
-
 
 .. tabs-drivers::
 
@@ -712,7 +706,6 @@ unencrypted:
             "policyNumber": 123142
         }
     }
-
 
 .. note::
 
--- a/source/use-cases/client-side-field-level-encryption-guide.txt
+++ b/source/use-cases/client-side-field-level-encryption-guide.txt
@@ -354,12 +354,14 @@ locally-managed master key.
 C. Define a JSON Schema
 ~~~~~~~~~~~~~~~~~~~~~~~
 
-`JSON Schema <http://json-schema.org/>`_ is a vocabulary that allows you
-to annotate and validate JSON documents. MongoDB drivers use an extended
-version of the JSON Schema standard to configure automatic client-side
-encryption and decryption of specific fields of the documents in a
-collection. The extended MongoDB CSFLE standard requires the following
-information:
+In this section, MedcoMD engineers configure the fields that the client
+automatically encrypts and decrypts using JSON Schema.
+`JSON Schema <http://json-schema.org/>`_, a vocabulary
+that allows you to annotate and validate JSON documents.
+MongoDB drivers use an extended version of the JSON Schema standard to
+configure automatic client-side encryption and decryption of specific fields
+of the documents in a collection. The extended MongoDB CSFLE standard requires
+the following information:
 
 - The encryption algorithm to use when encrypting each field
   (:manual:`Deterministic Encryption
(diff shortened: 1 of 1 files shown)


ccho-mongodb:
While the `note` addition containing the link to the server-side enforcement of CSFLE is directly after an `admonition`, it seemed an appropriate place to mention it.
Also added a sentence to the section to explain what based on user research feedback.

source/use-cases/client-side-field-level-encryption-guide.txt
```diff
@@ -354,12 +354,14 @@ locally-managed master key.
...
 
-`JSON Schema <http://json-schema.org/>`_ is a vocabulary that allows you
-to annotate and validate JSON documents. MongoDB drivers use an extended
-version of the JSON Schema standard to configure automatic client-side
-encryption and decryption of specific fields of the documents in a
-collection. The extended MongoDB CSFLE standard requires the following
-information:
+In this section, MedcoMD engineers configure the fields that the client
+automatically encrypts and decrypts using JSON Schema.
+`JSON Schema <http://json-schema.org/>`_, a vocabulary
```
nlarew: Grammar is off here. Either combine the two sentences or add an **is**

JSON Schema **is** a vocabulary...

source/use-cases/client-side-field-level-encryption-guide.txt
```diff
@@ -382,6 +384,14 @@ information:
    Any other document validation instances will cause the client to throw
    an error.
 
+.. note::
+
+   You can prevent clients that are not configured with the appropriate
+   client-side JSON Schema from writing unencrypted data to a field by
+   using server-side document validation. For server-side document validation
+   implementation details, see :manual:`Enforce Field Level Encryption Schema
+   </core/security-client-side-encryption/#enforce-field-level-encryption-schema>`.
```
nlarew: - Could give this note a title: **Server-side JSON Schema**
- unencrypted data OR data that is improperly encrypted
- Maybe add a sentence that explicitly says that this is only supplemental to CSFLE

source/use-cases/client-side-field-level-encryption-guide.txt
nathan-contino-mongo: I feel like it reads a little awkwardly to have JSON Schema stated twice back-to-back. Would it be better to phrase it like this (?) :

In this section, MedcoMD engineers configure the fields that the client automatically encrypts and decrypts using `JSON Schema <http://json-schema.org/>`_, a vocabulary that allows you to annotate and validate JSON documents.

source/use-cases/client-side-field-level-encryption-guide.txt
```diff
@@ -372,8 +374,8 @@ information:
 - The BSON Type of each field (only required for deterministically
   encrypted fields)
 
-.. admonition::Automatic Client-Side Field Level Encryption JSON Schema Does Not Support Document Validation
-   :important:
+.. admonition:: Automatic Client-Side Field Level Encryption JSON Schema Does Not Support Document Validation
```
nathan-contino-mongo: Is it possible to include a link here to the manual page on automatic CSFLE? I realize you were just fixing whitespace here, but I personally had to google automatic vs. normal CSFLE when I read this.

nathan-contino-mongo: It would also help this feel slightly less like a Demented randOm Capitalization statement

nathan-contino-mongo:
Two nits

source/use-cases/client-side-field-level-encryption-guide.txt
ccho-mongodb: > Could give this note a title: Server-side JSON Schema
Sure. I avoided that term because: 
1) It hasn't been declared as the de facto term in the server manual.
2) It could cause confusion since the section title doesn't specify which type (server-side or client-side) of schema we're creating in this step

> unencrypted data OR data that is improperly encrypted
Could you explain what you mean by "improperly encrypted"? Do you mean that the client-side JSON Schema references the wrong key? I wonder if there is a way to make this more general to encompass all the ways it could go wrong as it would otherwise require listing out that info.

> Maybe add a sentence that explicitly says that this is only supplemental to CSFLE
👍

source/use-cases/client-side-field-level-encryption-guide.txt
ccho-mongodb: That was my original edit, but decided the sentence contained too much information (both what is being accomplished by defining a JSON Schema and what JSON Schema is) which resulted in me creating a grammatical error in my manual undo which Nick spotted.

ccho-mongodb: Oops, will fix.

source/use-cases/client-side-field-level-encryption-guide.txt
ccho-mongodb: I think this does not include "improperly encrypted" for the following reason:

> Since the MongoDB server cannot decrypt nor introspect the contents of the encrypted field, it cannot validate that clients used the specified encryption options to encrypt a given field. This allows two clients to insert encrypted data using different keyIDs or encryption algorithms for a specific field. While some workloads may require independent field level encryption implementations, inconsistent implementation of encryption options for a field across clients may result in incorrect or unexpected behavior of queries against the encrypted field.

source/use-cases/client-side-field-level-encryption-guide.txt
ccho-mongodb: That said, if you think it's of value to join in a single sentence, I will update it.

source/use-cases/client-side-field-level-encryption-guide.txt
ccho-mongodb: Great feedback. I will omit "Automatic" as it does not belong. When writing this, I was probably in the frame of mind of emphasizing that CSFLE is automatic. While it could always be manually implemented, this feature is our "official" implementation.
From: https://docs.mongodb.com/manual/core/security-encryption-at-rest/#application-level-encryption

> New in version 4.2: MongoDB 4.2-series drivers provides a client-side field level encryption framework. For more information, see Client-Side Field Level Encryption.
> To encrypt full documents, write custom encryption and decryption routines or use a commercial solution.

ccho-mongodb: (I wish the server doc linked in this admonition did not contain automatic in the title)

nathan-contino-mongo: Sounds good, slightly more sane now :)

source/use-cases/client-side-field-level-encryption-guide.txt
nathan-contino-mongo: With the revised second sentence this should be good now.

nathan-contino-mongo:
LGTM

PR #576, opened 2019-11-25
(DOCSP-7796): separate out kms step

terakilobyte:
[Staged](https://docs-mongodbcom-staging.corp.mongodb.com/ecosystem/docsworker/DOCSP-7796/index.html)

### DIFF
--- a/source/includes/steps-fle-convert-to-a-remote-master-key.yaml
+++ b/source/includes/steps-fle-convert-to-a-remote-master-key.yaml
@@ -176,6 +176,7 @@ content: |
 title: Update the JSON Schema
 ref: update-the-json-schema
 content: |
-  If you have embedded the base64 key id for your data encryption key, you will
-  need to update your :ref:`JSON Schema <fle-define-a-json-schema>` with the new
-  base64 key id of your data encryption key.
+  If you have embedded the key id for your data encryption key in your JSON
+  Schema by hardcoding the string value, you will need to update your
+  :ref:`JSON Schema <fle-define-a-json-schema>` with the new key id of your data
+  encryption key.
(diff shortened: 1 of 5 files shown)


source/use-cases/client-side-field-level-encryption-local-key-to-kms.txt
```diff
@@ -0,0 +1,26 @@
...
+   :local:
+   :backlinks: none
+   :depth: 2
+   :class: singlecol
+
+.. _fle-convert-to-a-remote-master-key:
+
+Convert to a Remote Master Key
+------------------------------
+
```
ccho-mongodb: While we have a lead-in into this document from the CSFLE Guide, I think it would be helpful to include a couple sentences about who should read the guide, stage of implementation (ready for production), a reference to the CSFLE guide (in case someone lands on the wrong page), and what the benefits of a KMS are.

ccho-mongodb: Another item important to mention: any data that was inserted with the old data encryption key needs to be migrated or else it will not be decrypt-able.

source/includes/steps-fle-convert-to-a-remote-master-key.yaml
```diff
@@ -176,7 +176,6 @@ content: |
 title: Update the JSON Schema
 ref: update-the-json-schema
 content: |
-  The development JSON Schema directly referenced the development data
-  key id, so you need to generate a new JSON Schema. To do so, repeat
-  the :ref:`Define a JSON Schema <fle-define-a-json-schema>` step with
-  the new data key that was created with the remote master key.
+  If you have embedded the base64 key id for your data encryption key, you will
+  need to update your :ref:`JSON Schema <fle-define-a-json-schema>` with the new
+  base64 key id of your data encryption key.
```
ccho-mongodb: Have we defined the term "embedded" in this doc or step file? Perhaps it would be useful to specify the two different ways they may be referencing the data key -- either by query, or hardcoded schema. Perhaps something like this:

"Replace all references to the old data encryption key in your JSON schema with the new one you generated in the previous step whether directly embedded in your code or programmatically read."

terakilobyte:
@ccho-mongodb ready for another look

PR #578, opened 2019-11-25
(DOCSP-7888): x509 go snippet

mongomoe:
Ticket:  https://jira.mongodb.org/browse/DOCSP-7888


mongomoe:
@divjotarora pinging you for review on this, you've seen most of this code - but lines x and x are new

source/includes/connection-snippets/x509-snippets/go-connection.go
```diff
@@ -0,0 +1,29 @@
...
+
+	client, err := mongo.Connect(ctx, clientOpts)
+	if err != nil { log.Fatal(err) }
+
+
+	collection := client.Database("testDB").Collection("testCol")
+	docCount, err := collection.CountDocuments(ctx, nil, nil )
+	fmt.Println(docCount)
+
+	defer client.Disconnect(ctx)
```
divjotarora: This defer should be on line 21, right after checking the error from `mongo.Connect`.

source/includes/connection-snippets/x509-snippets/go-connection.go
```diff
@@ -0,0 +1,29 @@
...
+	uri := "mongodb+srv://<cluster-url>/test?authSource=$external&tlsCertificateKeyFile=%s&authMechanism=MONGODB-X509"
+	uri = fmt.Sprintf(uri, certificateKeyFilePath)
+	clientOpts := options.Client().ApplyURI(uri)
+
+	client, err := mongo.Connect(ctx, clientOpts)
+	if err != nil { log.Fatal(err) }
+
+
+	collection := client.Database("testDB").Collection("testCol")
+	docCount, err := collection.CountDocuments(ctx, nil, nil )
```
divjotarora: `err` should be checked.

source/includes/connection-snippets/x509-snippets/go-connection.go
mongomoe: Fixed

source/includes/connection-snippets/x509-snippets/go-connection.go
mongomoe: Fixed

mongomoe: Fixed

source/includes/connection-snippets/x509-snippets/go-connection.go
```diff
@@ -0,0 +1,28 @@
...
+	uri := "mongodb+srv://<cluster-url>/test?authSource=$external&tlsCertificateKeyFile=%s&authMechanism=MONGODB-X509"
+	uri = fmt.Sprintf(uri, certificateKeyFilePath)
+	clientOpts := options.Client().ApplyURI(uri)
+
+	client, err := mongo.Connect(ctx, clientOpts)
+	if err != nil { log.Fatal(err) }
+	defer client.Disconnect(ctx)
+
+	collection := client.Database("testDB").Collection("testCol")
+	docCount, err := collection.CountDocuments(ctx, nil, nil )
```
divjotarora: This will return an error because `nil` is not a valid filter. To give an empty filter, you should use an empty document like `bson.D{}`. Also, the second `nil` is unnecessary because the options are optional. So this line can just be `coll.CountDocuments(ctx, bson.D{})`

mongomoe: Fixed

PR #579, opened 2019-11-26
explain mongocryptd and configuration opts

ccho-mongodb:
[JIRA DOCSP-7797](https://jira.mongodb.org/browse/DOCSP-7797)

[Staged](https://docs-mongodbcom-staging.corp.mongodb.com/ecosystem/docsworker/DOCSP-7797-explain-mongocryptd-csfle/use-cases/client-side-field-level-encryption-guide.html)

### DIFF
--- a/source/use-cases/client-side-field-level-encryption-guide.txt
+++ b/source/use-cases/client-side-field-level-encryption-guide.txt
@@ -328,6 +328,7 @@ data encryption key:
 Follow the steps below to generate a single data encryption key from the
 locally-managed master key.
 
+
 .. tabs-drivers::
    :hidden: true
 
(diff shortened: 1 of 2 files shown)


source/use-cases/client-side-field-level-encryption-guide.txt
```diff
@@ -580,9 +581,180 @@ full `JSON Schema for the Medco Medical Management System
...
+                   spawnArgs.add("--idleShutdownTimeoutSecs=60");
+
+                   Map<String, Object> extraOpts = new HashMap<String, Object>();
+                   extraOpts.put("mongocryptdSpawnArgs", spawnArgs);
+
+                   AutoEncryptionSettings autoEncryptionSettings = AutoEncryptionSettings.builder()
+                       ...
+                       .extraOptions(extraOpts);
+
+             | TODO: THIS DOES NOT WORK, BUT SHOULD!
```
ccho-mongodb: Note to self

source/use-cases/client-side-field-level-encryption-guide.txt
```diff
@@ -580,9 +581,180 @@ full `JSON Schema for the Medco Medical Management System
...
+
+             .. example::
+
+                .. code-block:: javascript
+                   :emphasize-lines: 4
+
+                   autoEncryption: {
+                     ...
+                     extraOptions: {
+                       mongocryptdSpawnArgs: ["--port=30000"]
```
ccho-mongodb: Note to self: does not appear to work

source/use-cases/client-side-field-level-encryption-guide.txt
```diff
@@ -580,9 +581,180 @@ full `JSON Schema for the Medco Medical Management System
...
+
+             .. example::
+
+                .. code-block:: javascript
+                   :emphasize-lines: 4
+
+                   autoEncryption: {
+                     ...
+                     extraOptions: {
+                       mongocryptdSpawnArgs: ["--idleShutdownTimeoutSecs=75"]
```
ccho-mongodb: Note to self: does not appear to work

ccho-mongodb:
Would like an initial review even though the Java and Node examples are not working completely. @terakilobyte

ccho-mongodb:
> This addition looks good to me once we've verified the way to pass args is right/work.
> 
> One question. With the information embedded in a step users won't be able to quickly find this in the "on this page" section. Should we have it separated out into a section so it's more highlighted?

I considered this, but the rest of the guide explains concepts only when they are sequentially relevant (e.g. JSON Schema). This is the unfortunate combination of both use case and reference material that we agreed could be better organized by splitting the entire guide into two: one piece would be the narrative and example implementation, and the other would be a deeper dive into relevant concepts.

ccho-mongodb:
Already merged into a different integration branch

PR #583, opened 2019-11-26
(DOCSP-7886): added cpp x509 connection snippet

mongomoe:
Ticket: https://jira.mongodb.org/browse/DOCSP-7886
C++ reviewer(s): either @jmikola or @bazile-clyde
Docs reviewer: @zach-carr

### DIFF
--- a/source/includes/connection-snippets/x509-snippets/cpp-connection.cpp
+++ b/source/includes/connection-snippets/x509-snippets/cpp-connection.cpp
@@ -25,10 +25,7 @@ int main(int, char**) {
   auto client = mongocxx::client{uri, client_opts};
   mongocxx::database db = client["testDB"];
   mongocxx::collection collection = db["testCol"];
-
-  bsoncxx::builder::basic::document basic_builder{};
-  basic_builder.append(kvp("", ""));
-  bsoncxx::document::value filter = basic_builder.extract();
-  collection.count_documents(client, filter);
+  
+  std::cout << collection.count_documents({});
 }
 // end x509 connection
--- a/source/includes/connection-snippets/x509-snippets/cpp-connection.cpp
+++ b/source/includes/connection-snippets/x509-snippets/cpp-connection.cpp
@@ -1,0 +1,35 @@
+// begin x509 connection
+#include <iostream>
+#include <bsoncxx/builder/stream/document.hpp>
+#include <bsoncxx/json.hpp>
+#include <mongocxx/client.hpp>
+#include <mongocxx/uri.hpp>
+#include <mongocxx/instance.hpp>
+#include <mongocxx/options/client.hpp>
+#include <mongocxx/options/ssl.hpp>
+#include <bsoncxx/builder/basic/kvp.hpp>
+
+using bsoncxx::builder::basic::kvp;
+
+int main(int, char**) {
+  mongocxx::instance inst{};
+
+  mongocxx::options::ssl ssl_opts{};
+  ssl_opts.ca_file("/etc/certs/mongodb/ca.pem");
+  ssl_opts.pem_file("/etc/certs/mongodb/client.pem");
+
+  mongocxx::options::client client_opts{};
+  client_opts.ssl_opts(ssl_opts);
+
+  mongocxx::uri uri("mongodb+srv://<cluster-url>/test?authSource=$external&retryWrites=true&w=majority&authMechanism=MONGODB-X509");
+
+  auto client = mongocxx::client{uri, client_opts};
+  mongocxx::database db = client["testDB"];
+  mongocxx::collection collection = db["testCol"];
+
+  bsoncxx::builder::basic::document basic_builder{};
+  basic_builder.append(kvp("", ""));
+  bsoncxx::document::value filter = basic_builder.extract();
+  collection.count_documents(client, filter);
+}
+// end x509 connection
--- a/source/includes/connection-snippets/x509-snippets/cpp-connection.cpp
+++ b/source/includes/connection-snippets/x509-snippets/cpp-connection.cpp
@@ -1,15 +1,9 @@
 // begin x509 connection
 #include <iostream>
-#include <bsoncxx/builder/stream/document.hpp>
-#include <bsoncxx/json.hpp>
 #include <mongocxx/client.hpp>
 #include <mongocxx/uri.hpp>
 #include <mongocxx/instance.hpp>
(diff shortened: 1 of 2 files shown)


jmikola:
@kevinAlbs: please review

source/includes/connection-snippets/x509-snippets/cpp-connection.cpp
```diff
@@ -0,0 +1,34 @@
...
+int main(int, char**) {
+  mongocxx::instance inst{};
+
+  mongocxx::options::ssl ssl_opts{};
+  ssl_opts.pem_file("/etc/certs/mongodb/client.pem");
+
+  mongocxx::options::client client_opts{};
+  client_opts.ssl_opts(ssl_opts);
+
+  mongocxx::uri uri("mongodb+srv://<cluster-url>/test?authSource=$external&retryWrites=true&w=majority&authMechanism=MONGODB-X509");
```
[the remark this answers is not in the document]
mongomoe: Fixed: 
* Added `tls=true` to uri

source/includes/connection-snippets/x509-snippets/go-connection.go
```diff
@@ -13,7 +13,7 @@ import (
 func main() {
 	ctx := context.TODO()
 	certificateKeyFilePath := "/etc/certs/mongodb/client.pem"
-	uri := "mongodb+srv://<cluster-url>/test?authSource=$external&tlsCertificateKeyFile=%s&retryWrites=true&w=majority&authMechanism=MONGODB-X509"
+	uri := "mongodb+srv://<cluster-url>/test?authSource=$external&tlsCertificateKeyFile=%s&retryWrites=true&w=majority&authMechanism=MONGODB-X509&tls=true"
```
jmikola: Since this is injecting an arbitrary string into the URI via `%s`, I think you'll want to ensure that the value is URL encoded (as discussed in the [connection string spec](https://github.com/mongodb/specifications/blob/master/source/connection-string/connection-string-spec.rst)). This also relates to [Section 3.4 Query of RFC 3986](https://tools.ietf.org/html/rfc3986#section-3.4)). If there is an alternative method to provide this option to the Client, that may be easier.

@mongomoe: note that this applies to connection strings for _all_ languages, not just Go.

jmikola: Second thought: perhaps we don't care about URL encoding because the path is defined on the line above and does not contain any special characters. Feel free to disregard this comment.

source/includes/connection-snippets/x509-snippets/cpp-connection.cpp
```diff
@@ -0,0 +1,34 @@
...
+
+  auto client = mongocxx::client{uri, client_opts};
+  mongocxx::database db = client["testDB"];
+  mongocxx::collection collection = db["testCol"];
+
+  bsoncxx::builder::basic::document basic_builder{};
+  basic_builder.append(kvp("", ""));
+  bsoncxx::document::value filter = basic_builder.extract();
+  collection.count_documents(client, filter);
```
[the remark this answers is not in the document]
mongomoe: fixed

source/includes/connection-snippets/x509-snippets/cpp-connection.cpp
```diff
@@ -0,0 +1,31 @@
...
+int main(int, char**) {
+  mongocxx::instance inst{};
+
+  mongocxx::options::ssl ssl_opts{};
+  ssl_opts.pem_file("/etc/certs/mongodb/client.pem");
+
+  mongocxx::options::client client_opts{};
+  client_opts.ssl_opts(ssl_opts);
+
+  mongocxx::uri uri("mongodb+srv://<cluster-url>/test?authSource=$external&retryWrites=true&w=majority&authMechanism=MONGODB-X509&tls=true");
```
kevinAlbs: I'd recommend removing "test" from the URI since it is not used as the database to authenticate against for X509, so it serves no purpose. I.e. make it

`mongodb+srv://<cluster-url>/?authSource=$external&retryWrites=true&w=majority&authMechanism=MONGODB-X509&tls=true`

source/includes/connection-snippets/x509-snippets/cpp-connection.cpp
```diff
@@ -0,0 +1,31 @@
+// begin x509 connection
+#include <iostream>
+#include <bsoncxx/builder/stream/document.hpp>
+#include <bsoncxx/json.hpp>
+#include <mongocxx/client.hpp>
+#include <mongocxx/uri.hpp>
+#include <mongocxx/instance.hpp>
+#include <mongocxx/options/client.hpp>
+#include <mongocxx/options/ssl.hpp>
+#include <bsoncxx/builder/basic/kvp.hpp>
```
kevinAlbs: I believe the following includes aren't necessary anymore.

- #include <bsoncxx/builder/stream/document.hpp>
- #include <bsoncxx/json.hpp>
- #include <bsoncxx/builder/basic/kvp.hpp>

source/includes/connection-snippets/x509-snippets/cpp-connection.cpp
```diff
@@ -0,0 +1,31 @@
...
+
+using bsoncxx::builder::basic::kvp;
```
kevinAlbs: This line is no longer necessary.

source/includes/connection-snippets/x509-snippets/cpp-connection.cpp
mongomoe: Unfortunately, I can't remove `/test` from the uri. Although the spec doesn't require it, some of the drivers (for example, go) cause an error without it. So for this PR the goal is to have the drivers examples as similar as possible. We can remove `/test` from all the snippets eventually but for right now, if one of the drivers requires it, for unification purposes we'll have to keep it for the other drivers as well

kevinAlbs: Gotcha, that's fine with me then.

source/includes/connection-snippets/x509-snippets/cpp-connection.cpp
mongomoe: fixed

source/includes/connection-snippets/x509-snippets/cpp-connection.cpp
mongomoe: fixed

PR #584, opened 2019-11-26
(DOCSP-7893): added python client connection for x509

mongomoe:
Ticket: https://jira.mongodb.org/browse/DOCSP-7893
Python reviewer(s): either @prashantmital  or @ShaneHarvey 
Docs reviewer: @zach-carr

### DIFF
--- a/source/includes/connection-snippets/x509-snippets/python-connection.py
+++ b/source/includes/connection-snippets/x509-snippets/python-connection.py
@@ -1,5 +1,4 @@
 # begin x509 connection
-import ssl
 from pymongo import MongoClient
 
 uri = "mongodb+srv://<cluster-url>/?authMechanism=MONGODB-X509"
@@ -10,6 +9,6 @@ client = MongoClient(uri,
 
 db = client['test-database']
 collection = db['test-collection']
-docCount = collection.count_documents({'x': 1})
+docCount = collection.count_documents({})
 print(docCount)
 # end x509 connection
--- a/source/includes/connection-snippets/x509-snippets/python-connection.py
+++ b/source/includes/connection-snippets/x509-snippets/python-connection.py
@@ -1,11 +1,10 @@
 # begin x509 connection
 from pymongo import MongoClient
 
-uri = "mongodb+srv://<cluster-url>/?authMechanism=MONGODB-X509"
+uri = "mongodb+srv://<cluster-url>/test?authSource=$external&retryWrites=true&w=majority&authMechanism=MONGODB-X509"
 client = MongoClient(uri,
                      tls=True,
-                     tlsCertificateKeyFile='/etc/certs/mongodb/client.pem',
-                     tlsAllowInvalidCertificates=False)
+                     tlsCertificateKeyFile='/etc/certs/mongodb/client.pem')
 
 db = client['testDB']
 collection = db['testCol']
--- a/source/includes/connection-snippets/x509-snippets/python-connection.py
+++ b/source/includes/connection-snippets/x509-snippets/python-connection.py
@@ -8,6 +8,6 @@ client = MongoClient(uri,
 
 db = client['testDB']
 collection = db['testCol']
-docCount = collection.count_documents({})
-print(docCount)
+doc_count = collection.count_documents({})
+print(doc_count)
 # end x509 connection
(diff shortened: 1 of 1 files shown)


jmikola:
@ShaneHarvey: please review

source/includes/connection-snippets/x509-snippets/python-connection.py
```diff
@@ -0,0 +1,14 @@
+# begin x509 connection
+from pymongo import MongoClient
+
+uri = "mongodb+srv://<cluster-url>/?authMechanism=MONGODB-X509"
+client = MongoClient(uri,
+                     tls=True,
+                     tlsCertificateKeyFile='/etc/certs/mongodb/client.pem',
+                     tlsAllowInvalidCertificates=False)
```
jmikola: AFAIK, `tlsAllowInvalidCertificates` should already default to `false` in all drivers, so this might be redundant. I also didn't see this in any other language examples.

source/includes/connection-snippets/x509-snippets/python-connection.py
```diff
@@ -0,0 +1,14 @@
+# begin x509 connection
+from pymongo import MongoClient
+
+uri = "mongodb+srv://<cluster-url>/?authMechanism=MONGODB-X509"
+client = MongoClient(uri,
+                     tls=True,
```
jmikola: I don't think this is necessary, as PyMongo should enable TLS by virtue of `mongodb+srv` being used (per [driver spec](https://github.com/mongodb/specifications/blob/master/source/initial-dns-seedlist-discovery/initial-dns-seedlist-discovery.rst#seedlist-discovery)). The [unified URI options spec](https://github.com/mongodb/specifications/blob/master/source/uri-options/uri-options.rst) also suggests enabling TLS if one or more `tls*` options are used (e.g. `tlsCertificateKeyFile` below), although I'd defer to @ShaneHarvey and @prashantmital to confirm whether PyMongo honors that.

source/includes/connection-snippets/x509-snippets/python-connection.py
mongomoe: Sorry, `tlsAllowInvalidCertificates` should not be in there. I've removed it now

source/includes/connection-snippets/x509-snippets/python-connection.py
mongomoe: `tls=True` should technically not be required, but it's needed for this snippet. I actually need to update all the snippets to have `tls=True` in all the URI's because some drivers currently don't work without it.

jmikola: Can you clarify which drivers require `tls=true` to be explicitly specified? If any driver is not complying with the [Initial DNS Seedlist Discovery](https://github.com/mongodb/specifications/blob/master/source/initial-dns-seedlist-discovery/initial-dns-seedlist-discovery.rst) spec requirement to enable TLS by default when the `mongodb+srv` scheme is used that would be something we should fix immediately.

mongomoe: I don't have a full list of which drivers require tls=true to be explicitly specified. I can create a list of drivers that don't follow the spec exactly.  I do know for sure that Go requires it to be explicitly specified and throws an error without it (I think there's at least one other driver that does to, but I'd have to test it out).

ShaneHarvey: I can confirm that PyMongo will automatically default to `tls=true` when `mongodb+srv://` is used.

source/includes/connection-snippets/x509-snippets/python-connection.py
```diff
@@ -0,0 +1,13 @@
...
+from pymongo import MongoClient
+
+uri = "mongodb+srv://<cluster-url>/test?authSource=$external&retryWrites=true&w=majority&authMechanism=MONGODB-X509"
+client = MongoClient(uri,
+                     tls=True,
+                     tlsCertificateKeyFile='/etc/certs/mongodb/client.pem')
+
+db = client['testDB']
+collection = db['testCol']
+docCount = collection.count_documents({})
```
ShaneHarvey: `doc_count` would be more idiomatic in Python.

mongomoe: Fixed

source/includes/connection-snippets/x509-snippets/python-connection.py
mongomoe: For this PR we're sticking to `tls=true` because other drivers require it. I have another ticket (https://jira.mongodb.org/browse/DOCSP-7932) for updating these snippets once all the drivers match the spec, but currently, since some drivers require it, this bit has to be there.

PR #585, opened 2019-11-26
(DOCSP-7894): added ruby x509 snippet

mongomoe:
Ticket: https://jira.mongodb.org/browse/DOCSP-7894
Python reviewer(s): either @jmikola or @p-mongo
Docs reviewer: @zach-carr

### DIFF
--- a/source/includes/connection-snippets/x509-snippets/ruby-connection.rb
+++ b/source/includes/connection-snippets/x509-snippets/ruby-connection.rb
@@ -1,17 +1,11 @@
-# frozen_string_literal: true
-
 # begin x509 connection
 require 'mongo'
+certificate_key_file_path = '/etc/certs/mongodb/client.pem'
+uri = 'mongodb+srv://<cluster-url>/test?authSource=$external&tlsCertificateKeyFile=' + certificate_key_file_path + '&retryWrites=true&w=majority&authMechanism=MONGODB-X509'
 
-client = Mongo::Client.new(
-  ['mongodb+srv://<cluster-url>/test'],
-  auth_mech: :mongodb_x509,
-  ssl: true,
-  ssl_cert: '/etc/certs/mongodb/client.pem',
-  ssl_key: '/etc/certs/mongodb/client.pem'
-)
-db = client.database[:testDB]
+client = Mongo::Client.new([ uri ], :database => 'testDB')
+
 collection = client[:testCol]
-docCount = collection.count_documents({})
-print docCount
+doc_count = collection.count_documents({})
+print doc_count
 # end x509 connection
--- a/source/includes/connection-snippets/x509-snippets/ruby-connection.rb
+++ b/source/includes/connection-snippets/x509-snippets/ruby-connection.rb
@@ -1,0 +1,17 @@
+# frozen_string_literal: true
+
+# begin x509 connection
+require 'mongo'
+
+client = Mongo::Client.new(
+  ['<cluster-url>'],
+  auth_mech: :mongodb_x509,
+  ssl: true,
+  ssl_cert: '/etc/certs/mongodb/client.pem',
+  ssl_key: '/etc/certs/mongodb/client.pem'
+)
+db = client.database[:testDB]
+collection = client[:testCol]
+docCount = collection.count_documents({})
+print docCount
+# end x509 connection
--- a/source/includes/connection-snippets/x509-snippets/ruby-connection.rb
+++ b/source/includes/connection-snippets/x509-snippets/ruby-connection.rb
@@ -4,7 +4,7 @@
 require 'mongo'
 
 client = Mongo::Client.new(
-  ['<cluster-url>'],
+  ['mongodb+srv://<cluster-url>/test'],
   auth_mech: :mongodb_x509,
   ssl: true,
   ssl_cert: '/etc/certs/mongodb/client.pem',


source/includes/connection-snippets/x509-snippets/ruby-connection.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+# begin x509 connection
+require 'mongo'
+
+client = Mongo::Client.new(
+  ['mongodb+srv://<cluster-url>/test'],
```
p-mongo: When the client constructor is given an array, each array element must be a host address (not a uri). I filed https://jira.mongodb.org/browse/RUBY-2031 to improve our error reporting in this case but the code as provided here does not work.

Additionally, although it is permissible to provide both URI and Ruby options in the same call, I think the code will be easier to follow and reuse if it picked one of the two strategies (i.e. either use host address and Ruby options or provide all options in the URI).

p-mongo:
https://jira.mongodb.org/browse/DOCSP-7894 says this is for an atlas connection, I would expect that all drivers support all required URI options and since atlas uses URI syntax elsewhere, wouldn't it make sense to use URI options for x509 examples as well?

source/includes/connection-snippets/x509-snippets/ruby-connection.rb
jmikola: Additionally, I suggest that `/test` be removed here. 

In the original PR combining SCRAM and X509, it was never included for X509 examples. I also made the argument in https://github.com/mongodb/docs-ecosystem/pull/573#discussion_r350334451 that it was irrelevant for most drivers. Lastly, the corresponding X509-only Python PR (https://github.com/mongodb/docs-ecosystem/pull/584/files) does not include a default database name.

mongomoe: Updated the PR to use uri syntax (aside from the DB specification) with tls* uri options. The one thing is that `/test` needs to be in the uri to be equivalent to the other driver examples.

@p-mongo 
@jmikola

source/includes/connection-snippets/x509-snippets/ruby-connection.rb
```diff
@@ -1,17 +1,11 @@
-# frozen_string_literal: true
-
 # begin x509 connection
 require 'mongo'
+certificate_key_file_path = '/etc/certs/mongodb/client.pem'
+uri = 'mongodb+srv://<cluster-url>/test?authSource=$external&tlsCertificateKeyFile=' + certificate_key_file_path + '&retryWrites=true&w=majority&authMechanism=MONGODB-X509'
```
jmikola: @mongomoe: In https://github.com/mongodb/docs-ecosystem/pull/573#discussion_r349805444 you mentioned that `authSource=$external` was required for "other drivers". Do you know exactly what drivers those are? Our [auth spec's X509 section](https://github.com/mongodb/specifications/blob/master/source/auth/auth.rst#mongodb-x509) states that the default value is `$external` for the X509 mechanism, so I would expect all drivers to correctly enforce that default if they comply with the spec.

If this is not needed for any driver, then I would propose we remove this.

Separate note regarding `retryWrites=true`. Any driver that supports `tls*` URI options ([DRIVERS-557](https://jira.mongodb.org/browse/DRIVERS-557)) should also have implemented `retryWrites=true` by default ([DRIVERS-531](https://jira.mongodb.org/browse/DRIVERS-531)), so I think that is also redundant. Both of those issues were scheduled for 4.2-compat releases.

source/includes/connection-snippets/x509-snippets/ruby-connection.rb
jmikola: > The one thing is that /test needs to be in the uri to be equivalent to the other driver examples.

My previous argument in https://github.com/mongodb/docs-ecosystem/pull/573#discussion_r348462260 was that `/<db-name>` was not relevant for _any_ drivers. I would still advocate that it be removed from the SCRAM and X509 examples for all languages.

mongomoe: Unfortunately, some drivers do throw errors without it, for instance in go I get the following error if the url is `mongodb+srv://balboaclusterm10-5sfvt.mongodb.net/authSource=$external&tlsCertificateKeyFile=%s&authMechanism=MONGODB-X509` : 

# error:
``` error parsing uri: authsource without username is invalid```

source/includes/connection-snippets/x509-snippets/ruby-connection.rb
mongomoe: So for `authSource=$external` I don't currently have a complete list of the drivers that don't work without this (I can eventually put together a sheet of the inconsistencies amongst drivers - currently I just don't have the time for it). But I can definitely confirm that at the very least go doesn't work with it (just using go as an example because that's a language that I'm fairly comfortable with and currently have pulled up). For instance if I use the following url 
`mongodb+srv://balboaclusterm10-5sfvt.mongodb.net/test?tlsCertificateKeyFile=%s&authMechanism=MONGODB-X509` 

I get the error:

# error:
`auth source must be $external `

source/includes/connection-snippets/x509-snippets/ruby-connection.rb
p-mongo: This appears to be a bug in the Go driver then. https://github.com/mongodb/specifications/blob/master/source/auth/auth.rst#connection-string-options says:

> authSource

>    For GSSAPI and MONGODB-X509 authMechanisms the authSource defaults to $external

source/includes/connection-snippets/x509-snippets/ruby-connection.rb
p-mongo: This appears to be a bug in the Go driver then. Since auth source ought not to be required, I would say all drivers where it is in fact not required shouldn't have it in the example code.

p-mongo: If `certificate_key_file_path` is given as a variable, it should be escaped with e.g. `URI.encode`.

jmikola: > This appears to be a bug in the Go driver then. Since auth source ought not to be required, I would say all drivers where it is in fact not required shouldn't have it in the example code.

@mongomoe alluded to this in https://github.com/mongodb/docs-ecosystem/pull/584#discussion_r352800377. The immediate goal is to keep the examples consistent for all languages and punt on this until [DOCSP-7932](https://jira.mongodb.org/browse/DOCSP-7932). The fact that this is an outstanding bug in Go will unfortunately affect all language examples for the time being.

> If certificate_key_file_path is given as a variable, it should be escaped with e.g. URI.encode.

@p-mongo: I originally raised this point in https://github.com/mongodb/docs-ecosystem/pull/583#discussion_r352678773 but backed down after realizing that it was only moved to a variable for readability. I agree that it's technically correct to encode any credentials or query string values being inserted into the URI, but it may be out of scope for these simple examples. If the goal is consistency across the various languages, it might be problematic to have to source an external library for languages that don't have encoding in their standard library (e.g. C).

PR #587, opened 2019-12-02
Docsp 7885 x509 c snippet

mongomoe:


source/includes/connection-snippets/x509-snippets/c-connection.c
```diff
@@ -0,0 +1,32 @@
...
+   ssl_opts.pem_file = "/etc/certs/mongodb/client.pem";
+
+   client = mongoc_client_new ("mongodb+srv://<cluster-url>/test?authSource=$external&retryWrites=true&w=majority&authMechanism=MONGODB-X509");
+   mongoc_client_set_ssl_opts (client, &ssl_opts);
+
+   collection = mongoc_client_get_collection (client, "testDB", "testCol");
+   count = mongoc_collection_count_documents(collection, bson_new (), NULL, NULL, NULL, NULL);
+   printf ("%" PRId64 " documents counted.\n", count);
+
+   mongoc_client_destroy (client);
```
jmikola: Missing a call to [`mongoc_collection_destroy()`](http://mongoc.org/libmongoc/current/mongoc_collection_destroy.html) before this line. As-is, I believe the `collection` object will leak.

source/includes/connection-snippets/x509-snippets/c-connection.c
```diff
@@ -0,0 +1,32 @@
...
+int
+main (int argc, char *argv[])
+{
+   mongoc_collection_t *collection;
+   mongoc_client_t *client;
+   int64_t count;
+
+   mongoc_init ();
+
+   const mongoc_ssl_opt_t *ssl_default = mongoc_ssl_opt_get_default ();
```
jmikola: In https://github.com/mongodb/docs-ecosystem/pull/573/files#r348473040 and https://github.com/mongodb/docs-ecosystem/pull/573/files#r348478333 we talked about removing `mongoc_ssl_opt_t` entirely and relying on the `tls*` options. This would make the C example more consistent with other languages, since we can rely on the connection string alone.

source/includes/connection-snippets/x509-snippets/c-connection.c
```diff
@@ -0,0 +1,32 @@
+// begin x509 connection
+#include <mongoc/mongoc.h>
```
jmikola: @kevinAlbs Would it be prudent to also add `#include <stdio.h>` here as well? I think several of the libmongoc examples that print also include that.

source/includes/connection-snippets/x509-snippets/c-connection.c
```diff
@@ -0,0 +1,32 @@
...
+   mongoc_client_set_ssl_opts (client, &ssl_opts);
+
+   collection = mongoc_client_get_collection (client, "testDB", "testCol");
+   count = mongoc_collection_count_documents(collection, bson_new (), NULL, NULL, NULL, NULL);
+   printf ("%" PRId64 " documents counted.\n", count);
+
+   mongoc_client_destroy (client);
+   mongoc_cleanup ();
+
+   return EXIT_SUCCESS;
```
jmikola: I realize some of the example files that ship with libbson and libmongoc refer to `EXIT_SUCCESS`, but I suggest we just use `0` here.

source/includes/connection-snippets/x509-snippets/c-connection.c
mongomoe: Fixed

source/includes/connection-snippets/x509-snippets/c-connection.c
kevinAlbs: Good point, I think so. It's probably best not to rely on stdio.h being included through the include chain mongoc.h -> bson.h -> bson-compat.h.

source/includes/connection-snippets/x509-snippets/c-connection.c
```diff
@@ -0,0 +1,33 @@
...
+
+   memcpy(&ssl_opts, ssl_default, sizeof ssl_opts);
+
+   ssl_opts.pem_file = "/etc/certs/mongodb/client.pem";
+
+   client = mongoc_client_new ("mongodb+srv://<cluster-url>/test?authSource=$external&retryWrites=true&w=majority&authMechanism=MONGODB-X509");
+   mongoc_client_set_ssl_opts (client, &ssl_opts);
+
+   collection = mongoc_client_get_collection (client, "testDB", "testCol");
+   count = mongoc_collection_count_documents(collection, bson_new (), NULL, NULL, NULL, NULL);
```
kevinAlbs: Add a space before `(` for consistency.

The call to bson_new() leaks memory. Instead, define above:
`bson_t filter = BSON_INITIALIZER;`

And use that:
`count = mongoc_collection_count_documents (collection, &filter, NULL, NULL, NULL, NULL);`

And destroy it after:
`bson_destroy (&filter);`

source/includes/connection-snippets/x509-snippets/c-connection.c
```diff
@@ -0,0 +1,33 @@
...
+   printf ("%" PRId64 " documents counted.\n", count);
```
kevinAlbs: It might be worth adding an error check, and exiting on failure. `count_documents` returns -1 on error.

source/includes/connection-snippets/x509-snippets/c-connection.c
mongomoe: Fixed

source/includes/connection-snippets/x509-snippets/c-connection.c
mongomoe: Fixed

source/includes/connection-snippets/x509-snippets/c-connection.c
mongomoe: Fixed

source/includes/connection-snippets/x509-snippets/c-connection.c
mongomoe: Fixed

source/includes/connection-snippets/x509-snippets/c-connection.c
mongomoe: Fixed

mongomoe:
@jmikola @kevinAlbs all comments addressed, back to you for approval/additional comments

source/includes/connection-snippets/x509-snippets/c-connection.c
```diff
@@ -0,0 +1,32 @@
...
+int
+main (int argc, char *argv[])
+{
+   mongoc_collection_t *collection;
+   mongoc_client_t *client;
+   int64_t count;
+
+   mongoc_init ();
+
+   client = mongoc_client_new ("mongodb+srv://<cluster-url>/test?authSource=$external&tlsCertificateKeyFile=/etc/certs/mongodb/client.pem&retryWrites=true&w=majority&authMechanism=MONGODB-X509");
```
kevinAlbs: Per discussion, authSouce=$external isn't necessary, as that's the default for X509, but it's also harmless. I'm fine either way.

Also FYI for Jeremy, though w=majority and retryWrites=true are not strictly necessary for the example, they're intentionally left in to be consistent with all driver examples.

source/includes/connection-snippets/x509-snippets/c-connection.c
```diff
@@ -0,0 +1,32 @@
...
+
+   collection = mongoc_client_get_collection (client, "testDB", "testCol");
+   bson_t filter = BSON_INITIALIZER;
+   count = mongoc_collection_count_documents (collection, &filter, NULL, NULL, NULL, NULL);
+   if(count == -1){
```
kevinAlbs: You may want to make the parenthesis spacing consistent. clang-format can do this for you if you have it installed. And run
`clang-format -style=file -i c-connection.c`
Alternatively, since there are only a few inconsistencies, you can update them manually here.

source/includes/connection-snippets/x509-snippets/c-connection.c
```diff
@@ -0,0 +1,32 @@
...
+   mongoc_client_t *client;
+   int64_t count;
+
+   mongoc_init ();
+
+   client = mongoc_client_new ("mongodb+srv://<cluster-url>/test?authSource=$external&tlsCertificateKeyFile=/etc/certs/mongodb/client.pem&retryWrites=true&w=majority&authMechanism=MONGODB-X509");
+
+   collection = mongoc_client_get_collection (client, "testDB", "testCol");
+   bson_t filter = BSON_INITIALIZER;
+   count = mongoc_collection_count_documents (collection, &filter, NULL, NULL, NULL, NULL);
```
jmikola: http://mongoc.org/libmongoc/1.14.0/mongoc_collection_count_documents.html#example already has a good example for passing a `bson_errot_t` address as the last argument and using that for printing an error message. I'd suggest you just borrow that syntax, which would also change the conditional to `count < 0`.

source/includes/connection-snippets/x509-snippets/c-connection.c
```diff
@@ -0,0 +1,32 @@
...
+   mongoc_collection_t *collection;
+   mongoc_client_t *client;
+   int64_t count;
+
+   mongoc_init ();
+
+   client = mongoc_client_new ("mongodb+srv://<cluster-url>/test?authSource=$external&tlsCertificateKeyFile=/etc/certs/mongodb/client.pem&retryWrites=true&w=majority&authMechanism=MONGODB-X509");
+
+   collection = mongoc_client_get_collection (client, "testDB", "testCol");
+   bson_t filter = BSON_INITIALIZER;
```
jmikola: @kevinAlbs Should all declarations appear at the top of the scope (`main` function body in this case)? I know C99 doesn't require it, but most of the example code for libmongoc follows that practice. This would apply to the `bson_t filter` and any `bson_error_t error` that may be added.

source/includes/connection-snippets/x509-snippets/c-connection.c
mongomoe: Fixed

source/includes/connection-snippets/x509-snippets/c-connection.c
kevinAlbs: I did not mention it before because it did not seem necessary to change, but if we're going with C89 standard for all docs examples, then yes the declaration should be moved up.

source/includes/connection-snippets/x509-snippets/c-connection.c
mongomoe: Fixed

source/includes/connection-snippets/x509-snippets/c-connection.c
mongomoe: Fixed - moved all declarations to the top

PR #589, opened 2019-12-02
(DOCSP-7895): scala x509 conn snippet

mongomoe:


mongomoe:
@rozza Thanks, I've updated the PR to address your earlier comment

source/includes/connection-snippets/x509-snippets/scala-connection.scala
```diff
@@ -0,0 +1,31 @@
+// begin x509 connection
+import scala.collection.JavaConverters._
+import org.mongodb.scala._
+import org.mongodb.scala.bson.collection.mutable.Document
+import org.mongodb.scala.connection.{ClusterSettings, SslSettings}
+
+object MongoDBx509 extends App {
+
+  System.setProperty("javax.net.ssl.keyStore", "client.keystore");
```
zach-carr: Question: Are the semicolons on line 9/ 10 necessary? They're removed elsewhere.

mongomoe: Good catch, scala semicolons are optional, but it should be consistent in a file so I've removed the unneeded semicolons.

PR #590, opened 2019-12-02
(DOCSP-7891): perl connection snippet for x509

mongomoe:
Ticket: https://jira.mongodb.org/browse/DOCSP-7891


mongomoe:
Would love to see the non-uri based option @xdg

xdg:
You can adapt this: https://metacpan.org/pod/MongoDB::MongoClient#MONGODB-X509-(for-SSL-client-certificate)

The authMechanism can stay in the URI, though.

PR #591, opened 2019-12-02
(DOCSP-7603): add scram snippets

mongomoe:


source/includes/connection-snippets/scram-snippets/c-connection.c
```diff
@@ -0,0 +1,24 @@
...
+int
+main (int argc, char *argv[])
+{
+   mongoc_database_t *database;
+   mongoc_client_t *client;
+
+   mongoc_init ();
+
+   client = mongoc_client_new(
+      "mongodb+srv://<username>:<password>@<cluster-url>/test?retryWrites=true&w=majority"
```
jmikola: As previously mentioned in https://github.com/mongodb/docs-ecosystem/pull/573#discussion_r348462260, `/<db-name>` has no relevance for any drivers. It _may_ be used by some higher level libraries/ODMs, but that's outside the scope of any of these code examples.

I also assume we don't need to specify the authentication database via `authSource` since we're relying on that to be specified in the SRV/TXT result.

Furthermore, it's a bit odd that all of these connection strings were using `/test` but most of the examples select the `testDB` database. That's a bit inconsistent even if we ignore the issue of `/<db-name>` having no function.

source/includes/connection-snippets/scram-snippets/cpp-connection.cpp
```diff
@@ -0,0 +1,28 @@
...
+auto client = mongocxx::client conn{
+   mongocxx::uri{
+      "mongodb+srv://<username>:<password>@<cluster-url>/test?retryWrites=true&w=majority"
+   }
+};
+
+mongocxx::database db = client["testDB"];
+mongocxx::collection collection = db["testCol"];
+bsoncxx::builder::basic::document basic_builder{};
+basic_builder.append(kvp("", ""));
```
jmikola: See https://github.com/mongodb/docs-ecosystem/pull/583/files#r352347017. This example should be updated to be consistent with the X509 example once you settle on the correct syntax.

I suppose a more general question is whether there's any way for the docs team to utilize the same template for each language example. The only bit that really needs to change between SCRAM and X509 is the construction of the MongoClient object, and most of that is actually just the connection string. If there's some way to extract that bit out and utilize a common program template, that may make it easier to ensure that the programs themselves are correct and also facilitate the addition of future auth examples.

Ideally, I think drivers would own these language examples and run them as part of their test suite as we do for CRUD and transactions (@kay-kim is familiar with that), but I assume there's a reason that method was not utilized here.

source/includes/connection-snippets/scram-snippets/c-connection.c
```diff
@@ -0,0 +1,24 @@
...
+   );
+   database = mongoc_client_get_database (client, "test");
+
+   mongoc_database_destroy (database);
+   mongoc_client_destroy (client);
```
jmikola: Note that this program is not performing any IO and thus isn't testing that auth works. Other language examples issue a `count` command (see: [example in libmongoc docs](http://mongoc.org/libmongoc/current/tutorial.html#counting-documents)).

You may want to ensure this program is consistent with https://github.com/mongodb/docs-ecosystem/pull/587.

source/includes/connection-snippets/scram-snippets/php-connection.php
```diff
@@ -0,0 +1,10 @@
+<?php
+/* Begin scram connection */
+$client = new MongoDB\Client(
+    'mongodb+srv://<username>:<password>@<cluster-url>/test?retryWrites=true&w=majority'
+);
+
+$collection = $client->testDB->testCol;
+$docCount = $collection->count([]);
```
jmikola: [`count()`](https://docs.mongodb.com/php-library/current/reference/method/MongoDBCollection-count/) is deprecated. Please use [`countDocuments()`](https://docs.mongodb.com/php-library/current/reference/method/MongoDBCollection-countDocuments/). Also, note that the filter argument is optional defaults to `[]`, so you can simply call `$collection->countDocuments()`.

source/includes/connection-snippets/scram-snippets/php-connection.php
```diff
@@ -0,0 +1,10 @@
...
+print($docCount);
```
jmikola: The [`echo`](https://www.php.net/echo) language construct is more idiomatic for PHP. I'd suggest:

```php
echo $docCount, "\n";
```

ccho-mongodb:
@mongomoe If this PR is still relevant, can you link to the correct JIRA ticket? And are you missing a Swift connection snippet?

PR #601, opened 2019-12-16
DOCSP-7793: CSFLE security strategies chart

ccho-mongodb:
JIRA: https://jira.mongodb.org/browse/DOCSP-7793

Staging: https://docs-mongodbcom-staging.corp.mongodb.com/ecosystem/docsworker/DOCSP-7793-CSFLE-security-strategies-chart/use-cases/client-side-field-level-encryption-guide.html

docs-assets PR (must be merged into ecosystem branch before building on staging:  https://github.com/mongodb/docs-assets/pull/40

### DIFF
--- a/source/drivers/php.txt
+++ b/source/drivers/php.txt
@@ -60,8 +60,18 @@ You can install the extension using
 
 .. code-block:: sh
 
-   $ pecl install mongodb
-   $ echo "extension=mongodb.so" >> `php --ini | grep "Loaded Configuration" | sed -e "s|.*:\s*||"`
+   $ sudo pecl install mongodb
+
+Finally, add the following line to your ``php.ini`` file:
+
+.. code-block:: text
+
+   extension=mongodb.so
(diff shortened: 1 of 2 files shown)


ccho-mongodb:
> LGTM, but I have a suggestion:
> 
> it would be nice if there was some way to see the full-sized image of the diagram. Perhaps either adding a link to the image file, or if clicking on it would open the image file on a new tab. As it is currently, the non-bold text is a little hard to read.

I agree @mongomoe; however, the only alternative we have is to use the `..figure:` directive to have the image inherit a lightbox-type behavior. The way it renders can be inconsistent depending on the image's width. It either opens a lightbox or replaces your browser's location with just the url of the image, pulling you away from the page. Unfortunately, these are the tools we need to work with.

You can view the full-sized image by using your browser's function to load the image into a new tab or window. Hopefully, this is an acceptable solution for our readers.

PR #621, opened 2020-03-10
DOCS-12433: .NET driver link fixes

nathan-contino-mongo:
JIRA: https://jira.mongodb.org/browse/DOCS-12433

Stage: https://docs-mongodbcom-staging.corp.mongodb.com/ecosystem/pegasus/DOCS-12433/drivers/csharp.html

### DIFF
--- a/source/drivers/csharp.txt
+++ b/source/drivers/csharp.txt
@@ -20,11 +20,9 @@ Introduction
 The official MongoDB C#/.NET Driver provides asynchronous interaction
 with MongoDB.
 
-- :csharp-docs-latest:`Getting Started </getting_started/>`
+- :csharp-docs-latest:`Usage Guide </getting_started/>`
 
-- `Usage Guide <http://mongodb.github.io/mongo-csharp-driver/?jmp=docs>`__
-
-- :api:`API Reference <csharp/current?jmp=docs>`
+- :csharp-docs-latest:`API Reference </apidocs/html/R_Project_CSharpDriverDocs.htm>`
 
 - :csharp-docs-latest:`Changelog </what_is_new/>`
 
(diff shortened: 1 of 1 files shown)


nathan-contino-mongo:
NOTE: When The C# course debuts, we'll probably add that to this page as well (a la the Java driver page right now) but that's not quite here yet. This until then.

source/drivers/csharp.txt
```diff
@@ -20,11 +20,9 @@ Introduction
 The official MongoDB C#/.NET Driver provides asynchronous interaction
 with MongoDB.
 
-- :csharp-docs-latest:`Getting Started </getting_started/>`
+- :csharp-docs-latest:`Usage Guide </getting_started/>`
```
ccho-mongodb: Should the link name be changed? The section it links to is called Getting Started, so it could be confusing to jump context without an explanation.

nathan-contino-mongo: It's meant to follow the structure of the C++ page: https://docs-mongodbcom-staging.corp.mongodb.com/ecosystem/pegasus/DOCS-12433/drivers/cxx.html
which also contains a "Usage Guide" that just points to a general landing page. Take a look at the comments on the JIRA ticket because I'm mostly following those.

source/drivers/csharp.txt
```diff
@@ -34,7 +32,9 @@ with MongoDB.
 Installation
 ------------
 
-``NuGet`` is the simplest way to get the driver. `MongoDB.Driver <http://www.nuget.org/packages/mongodb.driver>`__ should be used for all new projects.
+``NuGet`` is the simplest way to get the driver. Use
+`MongoDB.Driver <http://www.nuget.org/packages/mongodb.driver>`__ for
```
ccho-mongodb: Thanks for changing this to active voice.
You should only need one underscore after an external link. I think the double underscore usually refers to an anonymous hyperlink https://docutils.sourceforge.io/docs/user/rst/quickref.html#external-hyperlink-targets

source/drivers/csharp.txt
nathan-contino-mongo: changing

source/drivers/csharp.txt
nathan-contino-mongo: fixing

PR #629, opened 2020-04-06
(DOCSP-9809) - adds swift and rust compat

terakilobyte:
### Issue JIRA link:
https://jira.mongodb.org/browse/DOCSP-9809

### Docs staging link:
https://docs-mongodbcom-staging.corp.mongodb.com/861ea72/drivers/docsworker/DOCSP-9809/driver-compatibility-reference

### A/C:
- [x] Swift compatibility added in lexicographic order
- [x] Rust compatibility added in lexicographic order

### DIFF
--- a/source/driver-compatibility-reference.txt
+++ b/source/driver-compatibility-reference.txt
@@ -632,6 +632,19 @@ Ruby Compatibility
      -
      - |checkmark|
 
+Rust Driver Compatibility
+-------------------------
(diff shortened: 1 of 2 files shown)


terakilobyte:
Thanks. I took a quick look at how to render the csharp table without it being so "squished" but then reverted the commit as it's going to take more digging into.

PR #634, opened 2020-04-09
Corrects links on C# page

terakilobyte:
### Issue JIRA link:
https://jira.mongodb.org/browse/DOCS-13576

### Docs staging link:
https://docs-mongodbcom-staging.corp.mongodb.com/d1766f9/drivers/docsworker-xlarge/DOCS-13576/csharp

### DIFF
--- a/source/csharp.txt
+++ b/source/csharp.txt
@@ -20,11 +20,11 @@ Introduction
 The official MongoDB C#/.NET Driver provides asynchronous interaction
 with MongoDB.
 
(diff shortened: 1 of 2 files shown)


ccho-mongodb:
> LGTM.
> However, could we either get rid of `:csharp-docs:` or just update it in `conf.py`? I think if we use the latter, it will be slightly easier to transition to fastly CDN to do the redirects later.

Never mind, we're on next gen now. I would remove it from conf.py after doing all the substitutions.

PR #637, opened 2020-05-05
delist all use-cases except csfle

terakilobyte:
## Pull Request Info

Removes all use-cases except for CSFLE as part of planned use-case deprecation. It also removes content that should have been removed previously. Lastly, this cleans up *all* build errors and warnings.

https://docs-mongodbcom-staging.corp.mongodb.com/e31de1f/drivers/docsworker-xlarge/remove-use-cases/use-cases


source/cxx.txt
```diff
@@ -68,7 +68,7 @@ MongoDB Compatibility
 .. admonition:: Development
    :class: important
 
-   `*` denotes a version in development.
+   ``*`` denotes a version in development.
```
ccho-mongodb, line 71: Should we keep this text if none of the displayed driver versions are in development?

PR #642, opened 2020-05-19
Update PHP compatibility tables

jmikola:
## Pull Request Info

@akung0324 asked about this in Slack.

PHPC 1.7 and PHPLIB 1.6 were no different from the previous versions, respectively, aside from adding FLE support and I don't believe we track that information in this table.

cc: @alcaeus


jmikola:
@terakilobyte: I don't have perms to request reviewers for this. Can you assign someone?

PR #643, opened 2020-05-27
DOCSP-10523: add link to golang tutorials

ccho-mongodb:
### Issue JIRA link:
https://jira.mongodb.org/browse/DOCSP-10523

https://docs-mongodbcom-staging.corp.mongodb.com/0fc41ed/drivers/docsworker/DOCSP-10523-add-golang-tutorials/go/

### DIFF
--- a/source/go.txt
+++ b/source/go.txt
@@ -19,8 +19,6 @@ Introduction
 
 This is the official MongoDB Go Driver.
 
-- `Get Started <https://www.mongodb.com/blog/post/mongodb-go-driver-tutorial>`__
-
 - `Quick Start Tutorials <https://www.mongodb.com/blog/search/golang%20quickstart>`__
 
 - `Migration from Community Drivers <https://www.mongodb.com/blog/post/go-migration-guide>`__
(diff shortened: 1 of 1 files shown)


source/go.txt
```diff
@@ -19,34 +19,33 @@ Introduction
 
 This is the official MongoDB Go Driver.
 
-- `General documentation for the MongoDB Go Driver <https://godoc.org/go.mongodb.org/mongo-driver/mongo>`__
+- `Get Started <https://www.mongodb.com/blog/post/mongodb-go-driver-tutorial>`__
```
terakilobyte, line 22: Overall I think this is fine, but a reason to not like it is it recommends to use the now out of fashion godep instead of gomod.

ccho-mongodb: I only moved this list item and renamed it (from "Get Started with Go"), but can remove it if you think it's not necessary since I added links to "Quick Start Tutorials".

terakilobyte: Sounds good, I think that’s better if there isn’t something more recent.

PR #653, opened 2020-07-13
Update PHP compat table for MongoDB 4.4

jmikola:
@ccho-mongodb: There's no JIRA ticket for this. I just figured I'd update this before you got around to it.

PR #652, opened 2020-07-13
DOCSP-11198: Update driver compatibility matrices for 4.4

ccho-mongodb:
## Pull Request Info

**Note: this PR will be merged and deployed only when 4.4 GA is released**

### Issue JIRA link:
https://jira.mongodb.org/browse/DOCSP-11198

### Docs staging link (requires sign-in on MongoDB Corp SSO):
https://docs-mongodbcom-staging.corp.mongodb.com/1a149b0/drivers/docsworker-xlarge/DOCSP-11198-update-driver-compat-4.4/

[C](https://docs-mongodbcom-staging.corp.mongodb.com/1a149b0/drivers/docsworker-xlarge/DOCSP-11198-update-driver-compat-4.4/c/#mongodb-compatibility)

[C++](https://docs-mongodbcom-staging.corp.mongodb.com/1a149b0/drivers/docsworker-xlarge/DOCSP-11198-update-driver-compat-4.4/cxx#mongodb-compatibility)

[C#](https://docs-mongodbcom-staging.corp.mongodb.com/1a149b0/drivers/docsworker-xlarge/DOCSP-11198-update-driver-compat-4.4/csharp#mongodb-compatibility)

[Go](https://docs-mongodbcom-staging.corp.mongodb.com/1a149b0/drivers/docsworker-xlarge/DOCSP-11198-update-driver-compat-4.4/go#mongodb-compatibility)

[Java](https://docs-mongodbcom-staging.corp.mongodb.com/1a149b0/drivers/docsworker-xlarge/DOCSP-11198-update-driver-compat-4.4/java#mongodb-compatibility)

[Java (Reactive Streams)](https://docs-mongodbcom-staging.corp.mongodb.com/1a149b0/drivers/docsworker-xlarge/DOCSP-11198-update-driver-compat-4.4/reactive-streams#mongodb-compatibility)

[Node](https://docs-mongodbcom-staging.corp.mongodb.com/1a149b0/drivers/docsworker-xlarge/DOCSP-11198-update-driver-compat-4.4/node#mongodb-compatibility)

[PHP](https://docs-mongodbcom-staging.corp.mongodb.com/1a149b0/drivers/docsworker-xlarge/DOCSP-11198-update-driver-compat-4.4/php#mongodb-compatibility)

[...]

PR #654, opened 2020-07-13
Update PHP language compat table

jmikola:
@ccho-mongodb: This adds the 4.4-compat release and complements https://github.com/mongodb/docs-ecosystem/pull/653. Notable change is that we dropped support for PHP 5.x.


jmikola:
Addressed in #652.

PR #655, opened 2020-07-15
DOCSP-10566 landing pages

ccho-mongodb:
### Issue JIRA link:
https://jira.mongodb.org/browse/DOCSP-10566

https://docs-mongodborg-staging.corp.mongodb.com/7ac8732/drivers/docsworker-xlarge/DOCSP-10566-landing-pages/java

### DIFF
--- a/snooty.toml
+++ b/snooty.toml
@@ -5,4 +5,4 @@ title = "MongoDB Drivers"
 version = 4.0
 package-name-org = "docs-ecosystem"
 pgp-version = "{+version+}"
-toc_landing_pages = ["/drivers/python"]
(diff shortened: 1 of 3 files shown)


snooty.toml
```diff
@@ -5,3 +5,4 @@ title = "MongoDB Drivers"
 version = 4.0
 package-name-org = "docs-ecosystem"
 pgp-version = "{+version+}"
+toc_landing_pages = ["/drivers/python"]
```
schmalliso, line 8: You need to use the path to the file, not the path to the URL, so this should just be `"/python"` and then everything will work.

PR #463, opened 2018-11-14
Include Mongolid Library

ravanscafi:
Easy, powerful and ultrafast ODM for PHP7 build on top of the [new mongodb driver](https://docs.mongodb.org/ecosystem/drivers/php/).

Mongolid

PR #669, opened 2020-08-11
Add Mongolid library

jmikola:
This is based on mongodb/docs-ecosystem#463 with suggested changes applied.


jmikola:
> but there's precedent on that page so it doesn't look too out of place as-is.

The prior art is something I was probably going to remove down the line while cleaning up stale projects for the old driver (`ext-mongo`), so I'm happy to revise this.

jmikola:
Please close https://github.com/mongodb/docs-ecosystem/pull/463 after merging/deploying this.

PR #670, opened 2020-08-17
DOCSP-11498: explain compatibility matrix

ccho-mongodb:
### Issue JIRA link:
https://jira.mongodb.org/browse/DOCSP-11498

[Content](https://docs-mongodbcom-staging.corp.mongodb.com/01387f4/drivers/docsworker-xlarge/DOCSP-11498-explain-compat-matrix/about-compatibility)

[Sample placement of link to content](https://docs-mongodbcom-staging.corp.mongodb.com/01387f4/drivers/docsworker-xlarge/DOCSP-11498-explain-compat-matrix/c#language-compatibility)

### DIFF
--- a/source/about-compatibility.txt
+++ b/source/about-compatibility.txt
@@ -39,7 +39,7 @@ to identify the following:
 .. note:: 
    
    Although the drivers do not advertise compatibility with the versions of
-   the server released after them, they are usually inter-operable.
+   the server released after them, they are usually interoperable.
 
 Avoid using older versions of the drivers that do not appear on the chart
 because they are not supported.
--- a/source/about-compatibility.txt
+++ b/source/about-compatibility.txt
@@ -1,0 +1,57 @@
+.. _about-driver-compatibility:
+
+============================
+MongoDB Compatibility Tables
+============================
+
+We include compatibility tables for each version of the driver to guide
+your decisions on what versions you need to ensure your environment
+remains fully operational.
+
+We maintain the following tables for each driver:
+
+- :ref:`MongoDB compatibility <mongodb-compatibility-tables>`
+- :ref:`Language compatibility <language-compatibility-tables>`
+
+Read the sections below for detailed explanations of each table.
+
+.. _mongodb-compatibility-tables
+
+MongoDB Compatibility Tables
+----------------------------
+
+In the **MongoDB Compatibility** tables, the columns are labeled with
+versions of MongoDB server and the rows are labeled with major release
+versions of the driver.
+
+The checkmarks (✓) indicate that the driver can access **all the
+features** of that specific version of MongoDB server unless those features 
+have been deprecated or removed. 
+
+If you are upgrading your server or driver version, you can use this table
+to identify the following:
+
+- version of the driver you need to access all the newest features
+  included in that version of the server
+
+- versions of the drivers tested against versions of the server
+
+.. note:: 
+   
+   Although the drivers do not advertise compatibility with the versions of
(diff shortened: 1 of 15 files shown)


source/about-compatibility.txt
```diff
@@ -0,0 +1,57 @@
...
+
+- versions of the drivers tested against versions of the server
+
+.. note:: 
+   
+   Although the drivers do not advertise compatibility with the versions of
+   the server released after them, they are usually inter-operable.
+
+Avoid using older versions of the drivers that do not appear on the chart
+because they are not supported.
```
nlarew, line 45: nit: A single sentence that follows a note/figure at the end of a section tends to feel a bit lost or out of place IMO. Maybe we could move this into an `.. important::` or bump it up above the note somewhere?

source/about-compatibility.txt
```diff
@@ -0,0 +1,57 @@
...
+
+- version of the driver you need to access all the newest features
+  included in that version of the server
+
+- versions of the drivers tested against versions of the server
+
+.. note:: 
+   
+   Although the drivers do not advertise compatibility with the versions of
+   the server released after them, they are usually inter-operable.
```
nlarew, line 42:
```suggestion
   the server released after them, they are usually interoperable.
```

source/about-compatibility.txt
```diff
@@ -0,0 +1,57 @@
...
+features** of that specific version of MongoDB server unless those features 
+have been deprecated or removed. 
+
+If you are upgrading your server or driver version, you can use this table
+to identify the following:
+
+- version of the driver you need to access all the newest features
+  included in that version of the server
+
+- versions of the drivers tested against versions of the server
```
nlarew, line 37: What do we mean by "tested" here. Is it "MongoDB currently tests" or somesuch? Or is it more like "this driver version would pass all integration tests with this server version"?

source/about-compatibility.txt
ccho-mongodb: Originally had this, but my spellchecker liked inter-operable better. Double-checked and looks like I can use the single-word version

source/about-compatibility.txt
ccho-mongodb: Last time I spoke with people on the drivers teams (though not all were present), they mentioned they were tested with those versions, but not necessarily using evergreen nor are they still tested. I'm guessing all the point releases get tested, but maybe not with the same versions of server. Maybe I'll leave it out for now and convert this into a sentence rather than a list of one.

ccho-mongodb:
> On a broader note (out of scope for now) it feels like this page could use some examples of a compatibility table (maybe all of them single-sourced and tabbed?). Alternatively, it feels like there may be enough space on the individual pages to include these as single-sourced explanations directly above their respective compatibility table. Interested in your thoughts on that type of approach!

I know this is a point of contention, but we used to have them all on one page but that wasn't working that well (despite substantial number of unique page views) because:

1) There is an increasing number and size of the tables. We'd be presenting a lot of irrelevant information for most users that only want to see one.
2) Driver teams such as Ruby have been maintaining their own table in a separate repo. There's no great way to sync with them.
3) Getting consensus on a single PR around 4.4 release was unnecessarily complex due to the large number of reviewers.

source/about-compatibility.txt
ccho-mongodb: Ok. I haven't heard this feedback before, but can reorder sections. I also want to avoid two admonitions in a row if possible.

PR #675, opened 2020-09-25
DOCSP-12465 rust connection examples

ccho-mongodb:
### Issue JIRA link:
https://jira.mongodb.org/browse/DOCSP-12465

https://docs-mongodborg-staging.corp.mongodb.com/506eafa/drivers/docsworker-xlarge/DOCSP-12465-rust-connection-examples/rust/#connect-to-mongodb-atlas

### DIFF
--- a/source/rust.txt
+++ b/source/rust.txt
@@ -47,13 +47,20 @@ corresponding connection code samples.
    .. tab:: Sync
       :tabid: rust-sync
 
+
+      Make sure you set ``default-features=false`` and enabled the
+      ``sync`` feature in your ``Cargo.toml`` file.
+      
+      Make sure you enabled the sync API. See
+      `Enabling the sync API <https://github.com/mongodb/mongo-rust-driver#enabling-the-sync-api>`__
+      for more details.
+
       .. code-block:: rust
 
-         // Be sure to set default-features=false and enable the "sync" feature in your Cargo.toml
          use mongodb::{bson::doc, sync::Client};
 
          fn main() -> mongodb::error::Result<()> {
-             // Get a handle to the deployment.
+             // Get a handle to the cluster
              let client = Client::with_uri_str(
                  "mongodb+srv://<username>:<password>@<cluster-url>/test?w=majority",
              )?;
@@ -63,7 +70,7 @@ corresponding connection code samples.
                  .run_command(doc! {"ping": 1}, None)?;
              println!("Connected successfully.");
 
-             // List the names of the databases in that deployment.
+             // List the names of the databases in that cluster
              for db_name in client.list_database_names(None, None)? {
                  println!("{}", db_name);
              }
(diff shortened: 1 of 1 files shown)


source/rust.txt
```diff
@@ -37,35 +37,71 @@ See `Installation <https://github.com/mongodb/mongo-rust-driver#Installation>`__
...
 
-.. code-block:: rust
+.. tabs::
+
+   .. tab:: Sync
+      :tabid: rust-sync
+
+      .. code-block:: rust
+
+         // Be sure to set default-features=false and enable the "sync" feature in your Cargo.toml
```
nlarew, line 52: This comment feels important - is there a page we can link to that covers `Cargo.toml` for the Rust driver? (Maybe not since it's new docs for a new driver)

source/rust.txt
```diff
@@ -37,35 +37,71 @@ See `Installation <https://github.com/mongodb/mongo-rust-driver#Installation>`__
...
+         use mongodb::{bson::doc, sync::Client};
+
+         fn main() -> mongodb::error::Result<()> {
+             // Get a handle to the deployment.
+             let client = Client::with_uri_str(
+                 "mongodb+srv://<username>:<password>@<cluster-url>/test?w=majority",
+             )?;
+
+             client
```
nlarew, line 61: Maybe add a comment explaining why we ping the admin db? Is this required to connect or just something we're doing for demo purposes?

source/rust.txt
```diff
@@ -37,35 +37,71 @@ See `Installation <https://github.com/mongodb/mongo-rust-driver#Installation>`__
...
+   .. tab:: Sync
+      :tabid: rust-sync
+
+      .. code-block:: rust
+
+         // Be sure to set default-features=false and enable the "sync" feature in your Cargo.toml
+         use mongodb::{bson::doc, sync::Client};
+
+         fn main() -> mongodb::error::Result<()> {
+             // Get a handle to the deployment.
```
nlarew, line 56: nit: is "deployment" a term we're using commonly across the driver docs? I would expect this to be "cluster" instead.

nlarew (approved):
LGTM w/ some small comments

source/rust.txt
ccho-mongodb: Not sure it's necessary, will ask Patrick.

source/rust.txt
ccho-mongodb: I know choosing terms is on Tony Sansone's task list, but I agree that cluster makes more sense especially since it's the term used in the connection string. Will update all instances.

source/rust.txt
ccho-mongodb: I think this is covered [here](https://github.com/mongodb/mongo-rust-driver#Installation), but maybe duplicating that or linking to that section would be useful in the tab. Will try something along those lines.

source/rust.txt
patrickfreed: The `Client` will be connecting in the background, so the ping isn't necessary. Running a ping here is basically a test that the background connections are succeeding, but the same thing could be achieved via just the subsequent `list_database_names` call. Perhaps it makes sense to just remove the ping and associated `println` then if it adds confusion.

source/rust.txt
```diff
@@ -37,35 +37,78 @@ See `Installation <https://github.com/mongodb/mongo-rust-driver#Installation>`__
...
+             println!("Connected successfully.");
+
+             // List the names of the databases in that cluster
+             for db_name in client.list_database_names(None, None)? {
+                 println!("{}", db_name);
+             }
+             Ok(())
+         }
+
+   .. tab:: Async
```
patrickfreed, line 76: Is there a possibility we could make the async tab the default one? We expect a large majority of users will be using the async API.

source/rust.txt
```diff
@@ -37,35 +37,78 @@ See `Installation <https://github.com/mongodb/mongo-rust-driver#Installation>`__
...
+      :tabid: rust-async
+
+      Make sure you specified your async runtime. See
```
patrickfreed, line 79: We also think most users will be using the `tokio` runtime, which is the default and requires no configuration. Perhaps we could reword this to the something like the following:
> The default async runtime used by the driver is [`tokio`](tokio.rs). To use a different runtime, see [Configuring the async runtime](https://github.com/mongodb/mongo-rust-driver#configuring-the-async-runtime).

source/rust.txt
ccho-mongodb: Sure, will update.

ccho-mongodb:
Addressed comments, back to you @patrickfreed

ccho-mongodb:
Addressed comments, back to you @patrickfreed !

patrickfreed (approved):
LGTM, thanks for updating this!

PR #676, opened 2020-09-30
DOP-931: Name footnotes in included files

mango-db:
## Pull Request Info
Anonymous footnotes spread across multiple include files are problematic. To maintain sanity, it's helpful to name the footnotes and footnote references so that the backend can match them up.

### Issue JIRA link:
https://jira.mongodb.org/browse/DOP-931

### With this change:
https://docs-mongodborg-staging.corp.mongodb.com/DOP-931/drivers/cgoecknerwald/master/

### Without this change:
https://docs-mongodborg-staging.corp.mongodb.com/master/drivers/cgoecknerwald/master/

### DIFF
--- a/source/includes/language-compatibility-table-java-rs.rst
+++ b/source/includes/language-compatibility-table-java-rs.rst
@@ -7,7 +7,7 @@
      - Java 6
      - Java 7
      - Java 8
-     - Java 11 [#language-compatibility-rs]_
+     - Java 11 [#backwards-compatible-rs]_
 
    * - 1.13
      - |checkmark|
@@ -81,4 +81,4 @@
      - |checkmark|
      - |checkmark|
 
-.. [#language-compatibility-rs] Java versions 8 and above are all supported thanks to the JVM backwards compatibility promise. Only LTS versions will be explicitly listed in future.
+.. [#backwards-compatible-rs] Java versions 8 and above are all supported thanks to the JVM backwards compatibility promise. Only LTS versions will be explicitly listed in future.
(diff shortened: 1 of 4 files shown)


ccho-mongodb:
1) I like the highlighting behavior -- I think that's really important to make the mental connection between the anchor link and the footnote.

2) I'm fine with naming footnotes in the RST, but is it possible to render them using more traditional symbols like star or cross?

3) I noticed that in the example you provided, the corresponding text isn't carried over to the actual note label. Instead, it renders as open and close square brackets with no text. Is that intentional? It seems like something is missing there.

mango-db:
> * I like the highlighting behavior -- I think that's really important to make the mental connection between the anchor link and the footnote.
> * I'm fine with naming footnotes in the RST, but is it possible to render them using more traditional symbols like star or cross?
> * I noticed that in the example you provided, the corresponding text isn't carried over to the actual note label. Instead, it renders as open and close square brackets with no text. Is that intentional? It seems like something is missing there.

1. Yay, thanks! :)
2. Yes, this could be possible -- file a ticket and assign it to Elle.
3. If you are opening the "with change" link, try refreshing; if you are opening the "without change" link, this is why the PR is necessary

ccho-mongodb:
> 3. If you are opening the "with change" link, try refreshing; if you are opening the "without change" link, this is why the PR is necessary

Here's what I see after clicking the backwards-compatible footnote link on drivers/java on the "with change" link:
Screen Shot 2020-09-30 at 4 16 47 PM
The label for the footnote description is "[]"

ccho-mongodb (approved):
After the latest render, looks good to me now. The item I mentioned about using more traditional symbols was only brought up since I was seeing the entire text of the footnote label so is no longer relevant.

PR #684, opened 2020-10-30
DOCSP-12575: Python CSFLE updates for Azure

ccho-mongodb:
### Issue JIRA link:
https://jira.mongodb.org/browse/DOCSP-12575
Note: this addresses Python driver changes for Azure support only. All the other combinations will be in separate tickets.

https://docs-mongodbcom-staging.corp.mongodb.com/77b2aad/drivers/docsworker-xlarge/DOCSP-12575-CSFLE-azure/security/client-side-field-level-encryption-local-key-to-kms

For relevant changes, select the Azure Key Vault tab (and AWS tab) and Python driver tab within that tabset.


terakilobyte (approved):
The changes lgtm

source/includes/steps-fle-convert-to-a-remote-master-key-azure.yaml
```diff
@@ -0,0 +1,235 @@
...
+        :tabid: python
+
+        In ``app.py``:
+
+        .. code-block:: python
+
+           master_key = {
+             "keyName": "<Azure key name>",
+             # "keyVersion": "<Azure key vault endpoint>",
+             "keyVaultEndpoint": "<Azure key version>"
```
prashantmital, line 222: These seem to be swapped. It should be:
```
"keyVersion": "<Azure key version>",
"keyVaultEndpoint": "<Azure key vault endpoint>"
```

source/includes/steps-fle-convert-to-a-remote-master-key-azure.yaml
```diff
@@ -0,0 +1,235 @@
...
+           }
+
+        .. note::
+
+           Make sure you are using `pymongocrypt <https://pypi.org/project/pymongocrypt/>`__
+           version 1.0.2 or later in your application's environment.
```
prashantmital, line 228: This is incorrect. `pymongocrypt-1.1.0` will be the first version of this library to support Azure. At the moment, we plan to release `1.1.0b0` this week and `1.1.0` when CSFLE is expected to GA.
I think to keep things simple, we should just mention that users need `pymongocrypt-1.1` or later, e.g.:
```
To use the Azure Key Vault, you must use `pymongocrypt <https://pypi.org/project/pymongocrypt/>`__
 version 1.1 or later in your application's environment.
```
The patch version should not be specified here because it can change without warning and is liable to become a maintenance burden.

prashantmital: Can we also add a note in the corresponding location for AWS that mentions that users must be using `pymongocrypt` 1.0 or later in their application environment? I checked and there doesn't seem to be one currently.

prashantmital (changes requested):

source/includes/steps-fle-convert-to-a-remote-master-key-azure.yaml
ccho-mongodb: Whoops, thanks for spotting this, will update.

source/includes/steps-fle-convert-to-a-remote-master-key-azure.yaml
ccho-mongodb: Sounds good, thanks for letting me know.

prashantmital (approved):

PR #695, opened 2020-11-07
GCP dotnet info

terakilobyte:
## Pull Request Info

This includes the config information for the dotnet driver to use gcp as a remote kms. It also cleans up the Node.js notes about not supporting GCP and Azure.

### Issue JIRA link:
https://jira.mongodb.org/browse/DOCSP-13041

https://docs-mongodbcom-staging.corp.mongodb.com/336656f/drivers/docsworker-xlarge/DOCSP-13041/security/client-side-field-level-encryption-local-key-to-kms#specify-your-google-cloud-kms-credentials

### DIFF
--- a/source/includes/steps-fle-convert-to-a-remote-master-key-gcp.yaml
+++ b/source/includes/steps-fle-convert-to-a-remote-master-key-gcp.yaml
@@ -109,16 +109,9 @@ content: |
      .. tab::
         :tabid: nodejs
 
-        .. code-block:: javascript
+        .. note::
 
-           // TODO: check correctness
-           kmsProviders = {
-             gcp: {
-               email: '<GCP service account email>',
-               privateKey: '<GCP service account private key>',
-               endpoint: '<GCP authentication endpoint>',
-             }
-           }
+           The Node.js driver does not currently support GCP KMS.
 
      .. tab::
         :tabid: python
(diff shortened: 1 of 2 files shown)


source/includes/steps-fle-convert-to-a-remote-master-key-gcp.yaml
```diff
@@ -132,6 +125,28 @@ content: |
...
+
+           var gcpPrivateKey = Environment.GetEnvironmentVariable("FLE_GCP_PRIVATE_KEY");
+           var gcpEmail = Environment.GetEnvironmentVariable("FLE_GCP_EMAIL");
+           var gcpEndpoint = Environment.GetEnvironmentVariable("FLE_GCP_IDENTITY_ENDPOINT"); // Optional, defaults to "oauth2.googleapis.com".
+           var gcpKmsOptions = new Dictionary<string, object>
+           {
+               { "privateKey", gcpPrivateKey },
+               { "email", gcpEmail },
+           };
+           if (gcpEndpoint != null)
```
ccho-mongodb, line 144: Does it still work if, for one of the optional parameters, you have a key, but no value? It could be easier to read without this conditional. Another idea: include it in a commented line as most users shouldn't need to update the default value.

DmitryLukyanov: > you have a key, but no value?

if I understood the question correctly, it's not what happens here. If the value is not presented, then the key won't be added

source/includes/steps-fle-convert-to-a-remote-master-key-gcp.yaml
```diff
@@ -245,6 +247,45 @@ content: |
...
+
+           var clientEncryption = new ClientEncryption(clientEncryptionOptions);
+
+
+           var gcpDataKeyProjectId = Environment.GetEnvironmentVariable("FLE_GCP_PROJ_ID");
+           var gcpDataKeyLocation = Environment.GetEnvironmentVariable("FLE_GCP_KEY_LOC"); // Optional. e.g. "global"
+           var gcpDataKeyKeyRing = Environment.GetEnvironmentVariable("FLE_GCP_KEY_RING");
+           var gcpDataKeyKeyName = Environment.GetEnvironmentVariable("FLE_GCP_KEY_NAME");
+           var gcpDataKeyKeyVersion = Environment.GetEnvironmentVariable("FLE_GCP_KEY_VERSION"); // Optional
+           var gcpDataKeyEndpoint = Environment.GetEnvironmentVariable("FLE_GCP_KMS_ENDPOINT"); // Optional, KMS URL, defaults to https://www.googleapis.com/auth/cloudkms
```
DmitryLukyanov, line 271: Make sure that all env variables match between different documents

DmitryLukyanov (changes requested):
Check the comment and then it will be LGTM

source/includes/steps-fle-convert-to-a-remote-master-key-gcp.yaml
terakilobyte: Thanks. These are all the same everywhere now.

ccho-mongodb (approved):

source/includes/steps-fle-convert-to-a-remote-master-key-gcp.yaml
DmitryLukyanov: up to you guys

DmitryLukyanov (approved):

PR #709, opened 2020-11-24
go gcp kms

terakilobyte:
### Issue JIRA link:
https://jira.mongodb.org/browse/DOCSP-13214

https://docs-mongodbcom-staging.corp.mongodb.com/e46f36a/drivers/docsworker-xlarge/DOCSP-13214/security/client-side-field-level-encryption-local-key-to-kms

### DIFF
--- a/source/includes/steps-fle-convert-to-a-remote-master-key-gcp.yaml
+++ b/source/includes/steps-fle-convert-to-a-remote-master-key-gcp.yaml
@@ -165,6 +165,31 @@ content: |
              To use the GCP KMS, you must use
              `libmongocrypt <https://github.com/mongodb/libmongocrypt#installing-libmongocrypt-on-windows>`__ version 1.1 or later in your application's environment.
 
+        .. tab::
+           :tabid: go
+
+           .. code-block:: go
+              type gcpKMSCredentials struct {
+                  Email      string `structs:"email"`
+                  PrivateKey string `structs:"privateKey"`
+              }
+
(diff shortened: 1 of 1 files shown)


source/includes/steps-fle-convert-to-a-remote-master-key-gcp.yaml
```diff
@@ -306,6 +332,43 @@ content: |
...
 
+        .. tab::
+           :tabid: go
+
+           .. code-block:: go
+
+              clientEncryptionOpts := options.ClientEncryption().SetKeyVaultNamespace(keyVaultNamespace).SetKmsProviders(kmsProviders)
+              keyVaultClient, err := mongo.Connect(context.TODO(), options.Client().ApplyURI(uri))
+              if err != nil {
+                  log.Panicf("Client encryption connect error %v", err)
```
ccho-mongodb, line 343: Would it be better to omit some of the error handling code from the explanation similar to the other language tab examples, or do you think it's better to include it? If the latter, maybe we could provide the users with more direction on how to resolve the error such as:
"Unable to connect to uri %v, check that your server is running on that host and port: %v"

ccho-mongodb: For the other languages, I tried to keep the code to a minimum by referring directly to the values that they should change in the CSFLE guide project. Could go the other way, but would need to change all of those examples.

ccho-mongodb (approved):
LGTM

PR #719, opened 2021-02-16
DOCS-14196 added .Net 5

kyuan-mongodb:
### Issue JIRA link:
https://jira.mongodb.org/browse/DOCS-14196

https://docs-mongodbcom-staging.corp.mongodb.com/ccfcbef/drivers/docsworker-xlarge/DOCS-14196-added-Net5/csharp/

### DIFF
--- a/source/includes/language-compatibility-table-csharp.rst
+++ b/source/includes/language-compatibility-table-csharp.rst
@@ -5,18 +5,18 @@
    :class: compatibility-large no-padding
 
    * - Driver Version
-     - .NET 4.5
-     - .NET 4.6
-     - .NET 4.7
-     - .NET 4.8
      - .NET 5
-     - .NET Core 1.0
-     - .NET Core 1.1
-     - .NET Core 2.0
-     - .NET Core 2.1
-     - .NET Core 2.2
-     - .NET Core 3.0
+     - .NET 4.8
+     - .NET 4.7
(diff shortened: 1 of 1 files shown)


MongoCaleb (changes requested):
I know this is an inherited problem, but I think we should take the time here to order the table so that the newest version is on the left.

MongoCaleb (approved):
LGTM!

source/includes/language-compatibility-table-csharp.rst
```diff
@@ -5,17 +5,18 @@
...
+     - .NET 4.7
+     - .NET 4.6
+     - .NET 4.5
      - .NET Core 3.1
+     - .NET Core 3.0
+     - .NET Core 2.2
+     - .NET Core 2.1
+     - .NET Core 2.0
+     - .NET Core 1.1
+     - .NET Core 1.0
```
terakilobyte, line 19: This ordering is incorrect. Strangely, the plain `.NET x` < 5 should follow after `.Net Core x`. See [here](https://www.tutorialsteacher.com/csharp/csharp-version-history). For reasons in the past we've omitted the word "Framework" from those early versions. I'm ok keeping this as nobody has complained, but we should keep them in order.

terakilobyte (changes requested):
Please correct the order

PR #738, opened 2021-05-21
DOCSP-16228 java compatibility table update

kyuan-mongodb:
## Pull Request Info
Updated the Java compatibility table

### Issue JIRA link:
https://jira.mongodb.org/browse/DOCSP-16228

https://docs-mongodbcom-staging.corp.mongodb.com/45c678d/drivers/docsworker-xlarge/DOCSP-16228-JavaCompatabilityTable/java/#compatibility

### DIFF
--- a/source/includes/mongodb-compatibility-table-java.rst
+++ b/source/includes/mongodb-compatibility-table-java.rst
@@ -15,6 +15,16 @@
      - MongoDB 3.0
      - MongoDB 2.6
 
+   * - 4.3
+     - |checkmark|
+     - |checkmark|
+     - |checkmark|
+     - |checkmark|
+     - |checkmark|
+     - |checkmark|
+     - |checkmark|
+     - |checkmark|
+     
    * - 4.2
      - |checkmark|
      - |checkmark|
@@ -46,7 +56,7 @@
      - |checkmark|
 
    * - 3.12
-     -
+     - |checkmark|  [#3.12-driver-support]_
      - |checkmark|
      - |checkmark|
      - |checkmark|
@@ -184,7 +194,7 @@
      -
      -
      -
-     - |checkmark|  [#driver-support]_
+     - |checkmark|  [#2.14-driver-support]_
      - |checkmark|
      - |checkmark|
 
@@ -208,7 +218,12 @@
      -
      - |checkmark|
 
-.. [#driver-support] The 2.14 driver does not support all MongoDB 3.2 features (e.g.,
+.. [#3.12-driver-support]  The 3.12 driver does not support all MongoDB 4.4 features
+
+.. [#2.14-driver-support] The 2.14 driver does not support all MongoDB 3.2 features (e.g.,
    read concern); however, if you are currently on a version 2.x driver
    and would like to run against MongoDB 3.2 but cannot upgrade to driver
    version 3.2, use the 2.14 driver.
+
+
+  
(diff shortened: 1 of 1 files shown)


source/includes/mongodb-compatibility-table-java.rst
```diff
@@ -15,6 +15,16 @@
      - MongoDB 3.0
      - MongoDB 2.6
 
+   * - 4.3
```
ccho-mongodb, line 18: This is still beta, so we shouldn't include 4.3 yet. For context, I think we made this agreement not to document beta versions in compatibility tables in a past Drivers Engineering Leads meeting.

source/includes/mongodb-compatibility-table-java.rst
```diff
@@ -46,7 +56,7 @@
      - |checkmark|
 
    * - 3.12
-     -
+     - |checkmark|  [#3.12-driver-support]_
```
ccho-mongodb, line 59: We should not indicate compatibility if it doesn't support all the new features. For context, I think we made this agreement in a past Drivers Engineering Leads meeting. This led to the creating of this page: https://docs.mongodb.com/drivers/about-compatibility/

This section describes when checkmarks are appropriate:
> The check marks (✓) indicate that the driver can access all the features of that specific version of MongoDB server unless those features have been deprecated or removed.

ccho-mongodb (changes requested):
I think the ticket reporter probably wasn't aware of agreements we had made with Driver Engineer Leads on what the compatibility tables should include. For next steps, I would recommend having the issue reporter speak with the Java team about updating their compatibility tables, specifically for the version 3.12 item.

source/includes/mongodb-compatibility-table-java.rst
```diff
@@ -208,7 +208,9 @@
      -
      - |checkmark|
 
-.. [#driver-support] The 2.14 driver does not support all MongoDB 3.2 features (e.g.,
+.. [#3.12-driver-support]  The 3.12 driver does not support all MongoDB 4.4 features
```
ccho-mongodb, line 211: Nitpick: there's an extra space before "The". This doesn't cause errors or change the output, but just want to keep it in the expected format case something changes in the future.

source/includes/mongodb-compatibility-table-java.rst
```diff
@@ -184,7 +184,7 @@
      -
      -
      -
-     - |checkmark|  [#driver-support]_
+     - |checkmark|  [#2.14-driver-support]_
```
ccho-mongodb, line 187: Nitpick: there's an extra space before "The". This doesn't cause errors or change the output, but just want to keep it in the expected format case something changes in the future.

source/includes/mongodb-compatibility-table-java.rst
```diff
@@ -208,7 +208,9 @@
      -
      - |checkmark|
 
-.. [#driver-support] The 2.14 driver does not support all MongoDB 3.2 features (e.g.,
+.. [#3.12-driver-support]  The 3.12 driver does not support all MongoDB 4.4 features
+
+.. [#2.14-driver-support] The 2.14 driver does not support all MongoDB 3.2 features (e.g.,
```
ccho-mongodb, line 213: Nitpick: I don't think we should add a comma directly after "e.g.". We should probably also avoid "e.g." and parentheses (I think the style guide mentions both of these) and instead:
"... all MongoDB 3.2 features such as read concern. However, ..."

ccho-mongodb (approved):
LGTM, a couple nitpicks.

PR #740, opened 2021-06-04
DOCSP-16255 python updates

ccho-mongodb:
### Issue JIRA link:
https://jira.mongodb.org/browse/DOCSP-16255

https://docs-mongodbcom-staging.corp.mongodb.com/615a6b7/drivers/docsworker-xlarge/DOCSP-16255-python-updates/python/

https://docs-mongodbcom-staging.corp.mongodb.com/615a6b7/drivers/docsworker-xlarge/DOCSP-16255-python-updates/pymongo/

https://docs-mongodbcom-staging.corp.mongodb.com/615a6b7/drivers/docsworker-xlarge/DOCSP-16255-python-updates/motor/

### DIFF
--- a/source/pymongo.txt
+++ b/source/pymongo.txt
@@ -78,6 +78,7 @@ Connect to MongoDB Atlas
    // replace this with your MongoDB connection string
    conn_str = "<your MongoDB Atlas connection string>"
 
+   // set a 5-second connection timeout
    client = pymongo.MongoClient(conn_str, serverSelectionTimeoutMS=5000)
 
    try:
@@ -85,11 +86,14 @@ Connect to MongoDB Atlas
    except Exception:
        print("Unable to connect to the server.")
 
-If the connection succeeds before a five-second timeout, you should see a
+If the connection succeeds before a five-second timeout, you will see a
 dictionary containing information about the server you connected to.
 
-If the connection fails, you should see the message "Unable to connect to
-the server."
+If the connection fails, you should see the following message:
+
+.. code-block:: console
+
+   Unable to connect to the server.
 
 For more information on the connection string, see the MongoDB Server
 Manual entry on :manual:`Connection String URI Format </reference/connection-string/>`.
--- a/source/pymongo.txt
+++ b/source/pymongo.txt
@@ -92,6 +92,7 @@ dictionary containing information about the server you connected to.
 If the connection fails, you should see the following message:
 
 .. code-block:: console
+   :copyable: false
 
    Unable to connect to the server.
 
--- a/source/pymongo.txt
+++ b/source/pymongo.txt
@@ -75,10 +75,10 @@ Connect to MongoDB Atlas
 
    import pymongo
 
-   // replace this with your MongoDB connection string
+   # replace this with your MongoDB connection string
    conn_str = "<your MongoDB Atlas connection string>"
 
-   // set a 5-second connection timeout
+   # set a 5-second connection timeout
    client = pymongo.MongoClient(conn_str, serverSelectionTimeoutMS=5000)
 
    try:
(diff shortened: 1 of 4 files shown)


source/pymongo.txt
```diff
@@ -76,10 +75,24 @@ Connect to MongoDB Atlas
...
+   conn_str = "<your MongoDB Atlas connection string>"
+
+   client = pymongo.MongoClient(conn_str, serverSelectionTimeoutMS=5000)
+
+   try:
+       print(client.server_info())
+   except Exception:
+       print("Unable to connect to the server.")
+
+If the connection succeeds before a five-second timeout, you should see a
```
kyuan-mongodb, line 88:
```suggestion
If the connection succeeds before a five-second timeout, you see a
```
Nit: removing "should" makes the outcome sound more concrete opposed to saying "you might see this"

Also applies to Motor page.

source/pymongo.txt
```diff
@@ -76,10 +75,24 @@ Connect to MongoDB Atlas
 
    import pymongo
 
-   client = pymongo.MongoClient(
-      "mongodb+srv://<username>:<password>@<cluster-url>/test?retryWrites=true&w=majority")
-   db = client.test
+   // replace this with your MongoDB connection string
+   conn_str = "<your MongoDB Atlas connection string>"
+
```
kyuan-mongodb, line 80: Suggestion: Since you mention the timeout after the snippet, it would be beneficial to mention that it's being created here with a comment

Also applies to Motor page.

source/pymongo.txt
```diff
@@ -76,10 +75,24 @@ Connect to MongoDB Atlas
...
+   try:
+       print(client.server_info())
+   except Exception:
+       print("Unable to connect to the server.")
+
+If the connection succeeds before a five-second timeout, you should see a
+dictionary containing information about the server you connected to.
+
+If the connection fails, you should see the message "Unable to connect to
```
kyuan-mongodb, line 91: Suggestion: Since this is output, can it be in a code-block?

Also applies to Motor page.

kyuan-mongodb (changes requested):
Looks great overall. Just some suggestions that I think can improve the page.

source/pymongo.txt
ccho-mongodb: Sounds good if you think it helps the reader.

I intentionally omitted that since I wanted these to be as brief as possible. To me, the goal of the connection snippet on landing pages is to be as brief as possible and not serve as a quick start. For example, we don't explain the connection string format, the client, the command to test the connection, nor output on the other drivers. I wanted to change Python drivers so that we gave slightly more guidance but kept it as short as possible, but this seems like a reasonable addition.

source/pymongo.txt
ccho-mongodb: I'll try it out. I intentionally did not format it in a code-block because:
1) a code block is more visually prominent than the description I'm using for the cse of successful output in the prior paragraph
2) it takes up more vertical space, and much more vertical space if I tried to describe the output iof server_info.

source/pymongo.txt
ccho-mongodb: I'm undecided on this one. I think "should" can make it sound less definitive as you mention which would be more appropriate for the case in which something else can possibly happen, but  "you see" sounds awkward and informal in general. Maybe it needs to be "you will see" to describe the future.

ccho-mongodb:
Updated the PR to address your comments @kyuan-mongodb . Let me know whether the output as a code block seems beneficial despite drawbacks I listed.

kyuan-mongodb (approved):
LGTM! I like the code-block because it makes it apparent it's an output from the console.

PR #743, opened 2021-07-07
DOCSP-17443: versioned api client instantiation snippets - Part 2

ccho-mongodb:
Python, Rust, Scala, and Swift drivers

### Issue JIRA link:
https://jira.mongodb.org/browse/DOCSP-17443

https://docs-mongodborg-staging.corp.mongodb.com/drivers/docsworker-xlarge/DOCSP-17443-versioned-api-py-ru-scala-sw/

Python - https://docs-mongodbcom-staging.corp.mongodb.com/drivers/docsworker-xlarge/DOCSP-17443-versioned-api-py-ru-scala-sw/pymongo/
Scala - https://docs-mongodbcom-staging.corp.mongodb.com/drivers/docsworker-xlarge/DOCSP-17443-versioned-api-py-ru-scala-sw/scala/
Reactive Streams - https://docs-mongodbcom-staging.corp.mongodb.com/drivers/docsworker-xlarge/DOCSP-17443-versioned-api-py-ru-scala-sw/reactive-streams/


source/pymongo.txt
```diff
@@ -99,6 +99,19 @@ If the connection fails, you should see the following message:
 For more information on the connection string, see the MongoDB Server
 Manual entry on :manual:`Connection String URI Format </reference/connection-string/>`.
 
+Versioned API
+-------------
+
+You can use the `Versioned API <https://docs.mongodb.com/v5.0/reference/versioned-api/>`__ 
+feature starting with MongoDB Server version 5.0 and Pymongo driver
```
rustagir, line 106: I think we agreed to capitalize Driver in this sentence, to keep consistency with the capitalization of Server.

source/includes/versioned-api-snippets/pymongo-client.py
```diff
@@ -0,0 +1,7 @@
+from pymongo.mongo_client import MongoClient
+from pymongo.server_api import ServerApi
+
+// Replace <connection string> with your MongoDB deployment's connection string.
```
rustagir, line 4: Comments in Python are preceded with #

source/includes/versioned-api-snippets/scala-client.scala
```diff
@@ -0,0 +1,13 @@
+import com.mongodb.{ServerApi, ServerApiVersion}
+import org.mongodb.scala._
+
+// ...
```
rustagir, line 4: The indentation is off here

source/includes/versioned-api-snippets/pymongo-client.py
```diff
@@ -0,0 +1,7 @@
...
+conn_str = "<connection string>"
+
+client = pymongo.MongoClient(conn_str, server_api=ServerApi('1'), serverSelectionTimeoutMS=5000)
```
rustagir, line 7: As in the other Versioned API snippets, would it be helpful to have another comment indicating where the API Version is specified? The comment reads:
"Set the version of the Versioned API on the client."

rustagir (changes requested):
Left some comments about small changes to make the sections consistent with the ones I did!

source/reactive-streams.txt
```diff
@@ -68,6 +70,19 @@ Connect to ``localhost``
 
 .. include:: /includes/localhost-connection.rst
 
+Versioned API
```
rustagir, line 73: This section should be after "connect to MongoDB Atlas"

source/pymongo.txt
ccho-mongodb: From reading how it's used throughout the site, it doesn't look like we have a consistent way to refer to it. I do think "MongoDB Server" is a product, but haven't decided if "PyMongo Driver" is the product name or if it's just PyMongo, the Python driver.
Let's keep the capitalization of driver here (in this case, a change), but not make any changes across the rest of the docs.

source/includes/versioned-api-snippets/pymongo-client.py
ccho-mongodb: Bad copy-pasta, thanks for catching this!

source/includes/versioned-api-snippets/scala-client.scala
ccho-mongodb: I will modify the rest of the code to be aligned with the first column of the line.

I can see how the indentation is confusing.
 
I've used "// ..." to denote "the rest of your code required to make this work" which means you'd probably want to have the class definition without indentation and then all the code within it in a function inside, so actually I'd want double indentation.
Instead, I think it's easier to align everything to the left so we're left with just the imports and section of the relevant code. Hopefully, we can get a collapsible code block element so that we can show both the full code and relevant code.

source/includes/versioned-api-snippets/pymongo-client.py
ccho-mongodb: Good catch, will update.

rustagir (approved):
LGTM!

PR #745, opened 2021-07-08
DOCSP-17096: versioned api ex

rustagir:
### Issue JIRA link:
https://jira.mongodb.org/browse/DOCSP-17096

Changes only made for Go at the moment
https://docs-mongodbcom-staging.corp.mongodb.com/drivers/docsworker-xlarge/DOCSP-17096-versioned-api-ex/go/

### DIFF
--- a/source/c.txt
+++ b/source/c.txt
@@ -79,18 +79,20 @@ for more ways to connect.
 Versioned API
 -------------
 
-Starting with MongoDB 5.0 and C Driver Version 1.18, you can configure your MongoDB client
-instance to use the Versioned API. The Versioned API maintains backwards
-compatibility, which means that you can deploy your app and upgrade your MongoDB
-server without changing any code or updating your driver. To use the
-Versioned API, use the
-:atlas:`Atlas connection string </driver-connection>` for your cluster
-and specify the API Version:
+You can use the `Versioned API
+<http://docs.mongodb.com/v5.0/reference/versioned-api/>`__ feature
+starting with MongoDB server version 5.0 and C Driver version 1.18. When
+you use the Versioned API feature with an official MongoDB driver, you
+can update your driver or server without worrying about backward
+compatibility issues for any commands covered by the Versioned API. To
+use the Versioned API, use the :atlas:`Atlas connection string
+</driver-connection>` for your cluster or a ``localhost`` connection
+string. You also need to specify the version of the Versioned API:
 
 .. code-block:: c
 
    // Replace the uri string with your MongoDB deployment's connection string.
-   const char *uri_string = "mongodb+srv://<username>:<password>@<cluster-url>/test?w=majority";
+   const char *uri_string = ""<connection string>"";
    mongoc_uri_t *uri;
    mongoc_client_t *client;
    mongoc_server_api_t *api;
@@ -101,7 +103,7 @@ and specify the API Version:
    uri = mongoc_uri_new_with_error (uri_string, &error);
    client = mongoc_client_new_from_uri (uri);
 
-   // Set the versioned API version on the client.
+   // Set the version of the Versioned API on the client.
    api = mongoc_server_api_new (MONGOC_SERVER_API_V1);
    mongoc_client_set_server_api (client, api, &error);
 
--- a/source/c.txt
+++ b/source/c.txt
@@ -80,14 +80,12 @@ Versioned API
 -------------
 
 You can use the `Versioned API
-<http://docs.mongodb.com/v5.0/reference/versioned-api/>`__ feature
-starting with MongoDB server version 5.0 and C Driver version 1.18. When
-you use the Versioned API feature with an official MongoDB driver, you
-can update your driver or server without worrying about backward
-compatibility issues for any commands covered by the Versioned API. To
-use the Versioned API, use the :atlas:`Atlas connection string
-</driver-connection>` for your cluster or a ``localhost`` connection
-string. You also need to specify the version of the Versioned API:
+<https://docs.mongodb.com/v5.0/reference/versioned-api/>`__ feature
+starting with MongoDB Server version 5.0 and C Driver version 1.18. When
+you use the Versioned API feature, you can update your driver or server
+without worrying about backward compatibility issues for any commands
+covered by the Versioned API. To use this feature, specify the version
+of the Versioned API:
 
 .. code-block:: c
 
--- a/source/c.txt
+++ b/source/c.txt
@@ -92,7 +92,7 @@ string. You also need to specify the version of the Versioned API:
 .. code-block:: c
 
    // Replace the uri string with your MongoDB deployment's connection string.
-   const char *uri_string = ""<connection string>"";
+   const char *uri_string = "<connection string>";
    mongoc_uri_t *uri;
    mongoc_client_t *client;
    mongoc_server_api_t *api;
--- a/source/c.txt
+++ b/source/c.txt
@@ -83,7 +83,7 @@ You can use the `Versioned API
 <https://docs.mongodb.com/v5.0/reference/versioned-api/>`__ feature
 starting with MongoDB Server version 5.0 and C Driver version 1.18. When
 you use the Versioned API feature, you can update your driver or server
-without worrying about backward compatibility issues for any commands
+without worrying about backward compatibility issues with any commands
 covered by the Versioned API. To use this feature, construct a MongoDB
 client instance, specifying a version of the Versioned API:
 
--- a/source/c.txt
+++ b/source/c.txt
@@ -84,12 +84,12 @@ You can use the `Versioned API
 starting with MongoDB Server version 5.0 and C Driver version 1.18. When
 you use the Versioned API feature, you can update your driver or server
 without worrying about backward compatibility issues for any commands
-covered by the Versioned API. To use this feature, specify the version
-of the Versioned API:
+covered by the Versioned API. To use this feature, construct a MongoDB
+client instance, specifying a version of the Versioned API:
 
 .. code-block:: c
 
-   // Replace the uri string with your MongoDB deployment's connection string.
+   // Replace <connection string> with your MongoDB deployment's connection string.
    const char *uri_string = "<connection string>";
    mongoc_uri_t *uri;
    mongoc_client_t *client;
--- a/source/c.txt
+++ b/source/c.txt
@@ -59,6 +59,7 @@ Connect to MongoDB Atlas
 
       mongoc_init ();
 
+      // Replace the uri string with your MongoDB deployment's connection string.
       client = mongoc_client_new(
          "mongodb+srv://<username>:<password>@<cluster-url>/test?retryWrites=true&w=majority"
       );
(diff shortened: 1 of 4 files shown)


rustagir:
Staging Links:

C Driver: https://docs-mongodbcom-staging.corp.mongodb.com/drivers/docsworker-xlarge/DOCSP-17096-versioned-api-ex/c/

C#/.NET Driver: https://docs-mongodbcom-staging.corp.mongodb.com/drivers/docsworker-xlarge/DOCSP-17096-versioned-api-ex/csharp/

Go Driver: https://docs-mongodbcom-staging.corp.mongodb.com/drivers/docsworker-xlarge/DOCSP-17096-versioned-api-ex/go/

source/c.txt
```diff
@@ -75,6 +76,35 @@ Connect to MongoDB Atlas
 See `Advanced Connections <http://mongoc.org/libmongoc/current/advanced-connections.html>`__
 for more ways to connect.
 
+Versioned API
+-------------
+
+Starting with MongoDB 5.0 and C Driver Version 1.18, you can configure your MongoDB client
+instance to use the Versioned API. The Versioned API maintains backwards
+compatibility, which means that you can deploy your app and upgrade your MongoDB
+server without changing any code or updating your driver. To use the
```
ccho-mongodb, line 85: Could we omit deployment here because it doesn't imply any change to driver nor server version?

I think if we want to provide a conceptual explanation of Versioned API, we should provide all the scenarios in which it helps you maintain compatibility such as changing driver version or server version (not just upgrading).

Happy to discuss over Zoom if you think it would be helpful.

source/includes/versioned-api-blurb.rst
```diff
@@ -0,0 +1,7 @@
+Starting with MongoDB 5.0, you can configure your MongoDB client
```
ccho-mongodb, line 1: Do you still intend to use this file? Otherwise, can you remove it?

ccho-mongodb (changes requested):
I think you did a really good job with the writing aspect. I know this topic is one of the more tricky ones to understand and communicate since I was one of the reviewers for the Server manual page on Versioned API. There are minor but important details that I think we want to communicate as precisely as possible, so I've left comments on those parts.

rustagir:
Staging Links:

C Driver: https://docs-mongodbcom-staging.corp.mongodb.com/drivers/docsworker-xlarge/DOCSP-17096-versioned-api-ex/c/

C#/.NET Driver: https://docs-mongodbcom-staging.corp.mongodb.com/drivers/docsworker-xlarge/DOCSP-17096-versioned-api-ex/csharp/

Go Driver: https://docs-mongodbcom-staging.corp.mongodb.com/drivers/docsworker-xlarge/DOCSP-17096-versioned-api-ex/go/

source/c.txt
```diff
@@ -75,6 +76,37 @@ Connect to MongoDB Atlas
 See `Advanced Connections <http://mongoc.org/libmongoc/current/advanced-connections.html>`__
 for more ways to connect.
 
+Versioned API
+-------------
+
+You can use the `Versioned API
+<http://docs.mongodb.com/v5.0/reference/versioned-api/>`__ feature
+starting with MongoDB server version 5.0 and C Driver version 1.18. When
```
ccho-mongodb, line 84: I think the proper noun for server is suitable here since you're using the proper noun version of driver. There's a minor distinction between the two forms, but let's stick to the proper noun versions of both here.

source/c.txt
```diff
@@ -75,6 +76,37 @@ Connect to MongoDB Atlas
...
+can update your driver or server without worrying about backward
+compatibility issues for any commands covered by the Versioned API. To
+use the Versioned API, use the :atlas:`Atlas connection string
+</driver-connection>` for your cluster or a ``localhost`` connection
+string. You also need to specify the version of the Versioned API:
+
+.. code-block:: c
+
+   // Replace the uri string with your MongoDB deployment's connection string.
+   const char *uri_string = "<connection string>";
```
ccho-mongodb, line 95: I think this is ok, but let's make every example consistent with how we name the placeholder and the comment describing it.

source/c.txt
```diff
@@ -75,6 +76,37 @@ Connect to MongoDB Atlas
 See `Advanced Connections <http://mongoc.org/libmongoc/current/advanced-connections.html>`__
 for more ways to connect.
 
+Versioned API
+-------------
+
+You can use the `Versioned API
+<http://docs.mongodb.com/v5.0/reference/versioned-api/>`__ feature
+starting with MongoDB server version 5.0 and C Driver version 1.18. When
+you use the Versioned API feature with an official MongoDB driver, you
```
ccho-mongodb, line 85: I think we need to mention that they must be using both Versioned API on the server and driver. You will likely encounter an error if you try to use Versioned API with the driver on a server version that doesn't support it.

source/c.txt
```diff
@@ -75,6 +76,37 @@ Connect to MongoDB Atlas
...
+can update your driver or server without worrying about backward
+compatibility issues for any commands covered by the Versioned API. To
+use the Versioned API, use the :atlas:`Atlas connection string
+</driver-connection>` for your cluster or a ``localhost`` connection
```
ccho-mongodb, line 89: (discussed in Slack)

source/c.txt
```diff
@@ -75,6 +76,37 @@ Connect to MongoDB Atlas
...
+Versioned API
+-------------
+
+You can use the `Versioned API
+<http://docs.mongodb.com/v5.0/reference/versioned-api/>`__ feature
+starting with MongoDB server version 5.0 and C Driver version 1.18. When
+you use the Versioned API feature with an official MongoDB driver, you
+can update your driver or server without worrying about backward
+compatibility issues for any commands covered by the Versioned API. To
+use the Versioned API, use the :atlas:`Atlas connection string
```
ccho-mongodb, line 88: I think the primary requirement is specifying the version of the Versioned API, so we want to mention that first rather than mention that it depends on having a connection string. All your MongoDB commands also depend on having a connection string that establishes a connection to a cluster/instance, but we probably don't need to mention that.

ccho-mongodb (changes requested):

source/c.txt
```diff
@@ -75,6 +76,37 @@ Connect to MongoDB Atlas
 See `Advanced Connections <http://mongoc.org/libmongoc/current/advanced-connections.html>`__
 for more ways to connect.
 
+Versioned API
+-------------
+
+You can use the `Versioned API
+<http://docs.mongodb.com/v5.0/reference/versioned-api/>`__ feature
```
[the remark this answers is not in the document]
rustagir: change accepted

source/c.txt
rustagir: suggestion accepted

source/c.txt
rustagir: it is already consistent across the three drivers here!

source/c.txt
rustagir: I think since the server and driver versions are specified in the first sentence, this sentence would impart enough information if it just read, "When you use the Versioned API feature, you..."

source/c.txt
rustagir: suggestion accepted-- dropping reference to the connection string altogether as it is clear within the code snippet and more info available elsewhere on the page.

rustagir:
Staging Links:

C Driver: https://docs-mongodbcom-staging.corp.mongodb.com/drivers/docsworker-xlarge/DOCSP-17096-versioned-api-ex/c/#versioned-api

C#/.NET Driver: https://docs-mongodbcom-staging.corp.mongodb.com/drivers/docsworker-xlarge/DOCSP-17096-versioned-api-ex/csharp/#versioned-api

Go Driver: https://docs-mongodbcom-staging.corp.mongodb.com/drivers/docsworker-xlarge/DOCSP-17096-versioned-api-ex/go/#versioned-api

source/c.txt
ccho-mongodb: Ok.
I'm not sure if this is within the scope of this ticket, but we're presenting a term two different ways in the same sentence. I think the intent may have been to refer to the variable name or placeholder name in directing what should be replaced. The C# example does not use a variable so that was a little confusing. I'll let you decide whether you think this could be clearer and whether to include that as part of this ticket.

source/c.txt
```diff
@@ -75,6 +76,35 @@ Connect to MongoDB Atlas
...
+Versioned API
+-------------
+
+You can use the `Versioned API
+<https://docs.mongodb.com/v5.0/reference/versioned-api/>`__ feature
+starting with MongoDB Server version 5.0 and C Driver version 1.18. When
+you use the Versioned API feature, you can update your driver or server
+without worrying about backward compatibility issues for any commands
+covered by the Versioned API. To use this feature, specify the version
+of the Versioned API:
```
ccho-mongodb, line 88: How would you feel about using the text "construct a client, specifying a version of the Versioned API" as that more accurately describes what the code shows and what's required?

rustagir: yeah I think the wording in this sentence needs some work. How is this:
"To use this feature, construct a MongoDB client instance with the version of the Versioned API."

source/c.txt
rustagir: Would it be more clear if the comment said: 

// Replace <connection string> with your MongoDB deployment's connection string.

ccho-mongodb: Sure, and then modified the string to be ``"<connection string>"`` or whatever syntax is necessary for that language.

source/c.txt
ccho-mongodb: Personal opinion: "Z with Y" often omits information on how the two relate. In this case, here is a reasonable way in which someone -- especially new to programming or these concepts -- could read the proposed sentence:

Instruct the version of the Versioned API to build a client instance so that you can use the feature.

This reading isn't correct since the version is a number, not some factory class that constructs new instances of a client class.

rustagir: Okay, I will use the "..., specifying ..." structure. thanks!

source/c.txt
```diff
@@ -75,6 +76,35 @@ Connect to MongoDB Atlas
...
 for more ways to connect.
 
+Versioned API
+-------------
+
+You can use the `Versioned API
+<https://docs.mongodb.com/v5.0/reference/versioned-api/>`__ feature
+starting with MongoDB Server version 5.0 and C Driver version 1.18. When
+you use the Versioned API feature, you can update your driver or server
+without worrying about backward compatibility issues for any commands
```
ccho-mongodb, line 86: Nitpick: I think you're describing "issues with commands" rather than "issues for commands" unless you're offering the commands issues for something.

source/c.txt
ccho-mongodb: (of course feel free to come up with your own solutions)

source/go.txt
```diff
@@ -56,6 +56,29 @@ Connect to MongoDB Atlas
...
+
+.. code-block:: go
+
+   // Replace <connection string> with your MongoDB deployment's connection string.
+   uri := "<connection string>"
+   
+   // Set the version of the Versioned API on the client.
+   serverAPIOptions := options.ServerAPI(options.ServerAPIVersion1)
+   clientOptions := options.Client().ApplyURI(uri).SetServerAPIOptions(serverAPIOptions)
+   ctx := context.TODO()
```
ccho-mongodb, line 78: Nitpick: let's move this line before the comment "// Set the version..." since context.TODO has nothing to do with that comment.

ccho-mongodb (approved):

source/c.txt
rustagir: accepted

source/go.txt
rustagir: accepted

rustagir:
Staging Links:

C Driver: https://docs-mongodbcom-staging.corp.mongodb.com/drivers/docsworker-xlarge/DOCSP-17096-versioned-api-ex/c/#versioned-api

C#/.NET Driver: https://docs-mongodbcom-staging.corp.mongodb.com/drivers/docsworker-xlarge/DOCSP-17096-versioned-api-ex/csharp/#versioned-api

Go Driver: https://docs-mongodbcom-staging.corp.mongodb.com/drivers/docsworker-xlarge/DOCSP-17096-versioned-api-ex/go/#versioned-api

PR #749, opened 2021-07-19
Docsp 15578

terakilobyte:
## Pull Request Info

This simplifies navigation to the new java driver.

### Issue JIRA link:
https://jira.mongodb.org/browse/DOCSP-15578

### Docs staging link (requires sign-in on MongoDB Corp SSO):
https://docs-mongodbcom-staging.corp.mongodb.com/drivers/docsworker-xlarge/DOCSP-15578/java-drivers/

PR #761, opened 2021-09-17
redirect to new docset

terakilobyte:
## Pull Request Info

This redirects drivers/go to drivers/go/current/ and into the new docset.

### Issue JIRA link:
N/A

### Snooty build log:
https://workerpool-boxgs.mongodbstitch.com/pages/job.html?jobId=6144eca1c608bec57f3c4fa7

https://docs-mongodborg-staging.corp.mongodb.com/ecosystem/docsworker/NNNNN/index.html

### Self-Review Checklist

- [ ] Is this free of any warnings or errors in the RST?
- [ ] Does it render on staging correctly?
- [ ] Are the staging and workerpool job links in the PR description updated?

### DIFF
--- a/config/redirects
+++ b/config/redirects
@@ -4,6 +4,9 @@ define: manual https://docs.mongodb.com/manual
 define: atlas https://docs.atlas.mongodb.com
 define: devhub https://developer.mongodb.com
 
+# Go
+raw: drivers/go -> ${base}/go/current/
+
 # Java
 raw: drivers/java -> ${base}/java-drivers/
 
--- a/config/redirects
+++ b/config/redirects
@@ -54,7 +54,7 @@ raw: drivers/tutorial/use-java-dbobject-to-perform-saves -> ${base}/java/
 raw: drivers/tutorial/authenticate-with-java-driver/ -> http://mongodb.github.io/mongo-java-driver/3.0/driver/reference/connecting/authenticating
 
 # nodejs
-raw: drivers/javascript -> ${base}/node/
+raw: drivers/javascript -> ${base}/node/current/
 
 # ruby and mongoid
 raw: drivers/tutorial/getting-started-with-ruby-on-rails-3 -> https://docs.mongodb.com/ruby-driver/current/tutorials/bson-v4/


ccho-mongodb (approved):
LGTM

config/redirects
```diff
@@ -4,6 +4,9 @@ define: manual https://docs.mongodb.com/manual
 define: atlas https://docs.atlas.mongodb.com
 define: devhub https://developer.mongodb.com
 
+# Go
+raw: drivers/go -> ${base}/go/current/
```
biniona-mongodb, line 8: Question: Any reason why this redirect goes to the current branch? Seems like Java and [Node](https://github.com/mongodb/docs-ecosystem/blob/b78f0518081ab9ffcf176a323c335ec488500974/config/redirects#L57) don't go to a branch?

ccho-mongodb: > Question: Any reason why this redirect goes to the current branch? Seems like Java and [Node](https://github.com/mongodb/docs-ecosystem/blob/b78f0518081ab9ffcf176a323c335ec488500974/config/redirects#L57) don't go to a branch?

There's no redirect for Java (java/sync) in this file. We added a different redirect to address how we renamed the drawer that contains all the JVM drivers.

The `drivers/javascript` redirect was likely added when we only had one version of node. We should update that one to current as it currently leads to two redirects (drivers/javascript -> drivers/node -> drivers/node/current).

The Java one is not a redirect to the java/sync driver.

kyuan-mongodb:
Question: Don't you also need to update the TOC here like you did for the [Java docs](https://github.com/mongodb/docs-ecosystem/pull/749/files)?

terakilobyte:
@ccho-mongodb Good point about the node redirect, I'll add that to this. 
@kyuan-mongodb There's no need since the redirect will capture it.

ccho-mongodb:
> Question: Don't you also need to update the TOC here like you did for the [Java docs](https://github.com/mongodb/docs-ecosystem/pull/749/files)?

Yes, I think we should update the index ToC to avoid an unnecessary redirect:
```
Go Driver </go/current/>
```

PR #769, opened 2021-11-09
Docsp 19598 compat 5.1

ccho-mongodb:
### Issue JIRA link:
https://jira.mongodb.org/browse/DOCSP-19598

- Omitted SWIFT (another followup ticket addresses this)

### Snooty build log:
https://workerpool-boxgs.mongodbstitch.com/pages/job.html?jobId=618ab5fbc13055282d708379

https://docs-mongodbcom-staging.corp.mongodb.com/drivers/docsworker-xlarge/ccho-mongodb-DOCSP-19598-compat-5.1/

### Self-Review Checklist

- [ ] Is this free of any warnings or errors in the RST?
- [ ] Does it render on staging correctly?
- [ ] Are the staging and workerpool job links in the PR description updated?

### DIFF
--- a/source/includes/mongodb-compatibility-table-c.rst
+++ b/source/includes/mongodb-compatibility-table-c.rst
@@ -4,6 +4,7 @@
    :class: compatibility-large
 
    * - C Driver Version
+     - MongoDB 5.1
      - MongoDB 5.0
      - MongoDB 4.4
      - MongoDB 4.2
@@ -15,6 +16,7 @@
      - MongoDB 2.6
 
    * - 1.19
+     -
      - |checkmark|
      - |checkmark|
      - |checkmark|
@@ -27,6 +29,7 @@
 
    * - 1.18
      - |checkmark| [#c-1.18-driver-support]_
+     - |checkmark| [#c-1.18-driver-support]_
      - |checkmark|
      - |checkmark|
      - |checkmark|
@@ -37,6 +40,7 @@
      -
 
    * - 1.17
+     -
      -
      - |checkmark|
      - |checkmark|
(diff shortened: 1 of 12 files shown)


terakilobyte (approved):
LGTM, though I'm wondering if we should add an asterisk to each of these under 5.1 with a note that says they will work with 5.1 but may not support the updated versioned api

source/includes/mongodb-compatibility-table-c.rst
```diff
@@ -15,6 +16,7 @@
      - MongoDB 2.6
 
    * - 1.19
+     -
```
biniona-mongodb, line 19: Is the C driver version 1.19 not compatible with MongoDB v5. while version 1.18 of the driver is compatible? No worries, if this is correct, but it seemed surprising at first glance.

ccho-mongodb: No, good catch.

ccho-mongodb:
> LGTM, though I'm wondering if we should add an asterisk to each of these under 5.1 with a note that says they will work with 5.1 but may not support the updated versioned api

No, and we should try not to feature asterisks. Over a year ago, Bernie/Rachelle wanted to get rid of all asterisks. According to them, compatibility tables should show access to full feature set of server, not that you can use that version with a particular version of server. Almost all driver versions still in the tables work with the newest server version, but can't access features like versioned API. I brought up that it is not intuitive to everyone, but I don't see much of an alternative. My idea to document features individually was determined to be too much of a burden for those that would need to be involved, and that versioned API should address that.  To address this ambiguity, I created the "about compatibility tables" page to explain how to read them.

PR #793, opened 2022-02-03
fix Reactive Streams RST

ccho-mongodb:
### Issue JIRA link:
None

### Snooty build log:
https://workerpool-boxgs.mongodbstitch.com/pages/job.html?collName=queue&jobId=61fc003f382d7fe4ee6ee86a

https://docs-mongodbcom-staging.corp.mongodb.com/drivers/docsworker-xlarge/020321-table-fix/reactive-streams/#mongodb-compatibility

### Self-Review Checklist


ccho-mongodb:
> Renders and looks good to me. I do wonder if we need the statement after the table "The driver does not support older versions of MongoDB", though not important to shipping this.

I haven't investigated the history of the statement, but it seems accurate and might provide value through explicit statement.

PR #801, opened 2022-03-11
(DOCSP-19842) CSFLE IAM Role

biniona-mongodb:
## Pull Request Info

Add coverage of IAM Role support for authentication with AWS in CSFLE.

### Issue JIRA link:
https://jira.mongodb.org/browse/DOCSP-19842

### Snooty build log:

https://workerpool-boxgs.mongodbstitch.com/pages/job.html?collName=queue&jobId=622bb02a0ab54725f1904c76

https://docs-mongodbcom-staging.corp.mongodb.com/drivers/docsworker-xlarge/DOCSP-19842-CSFLE-IAM-Role/security/client-side-field-level-encryption-local-key-to-kms/#create-an-aws-iam-user

### Self-Review Checklist

### If your page documents a concept, does it meet the following criteria?

- [ ] Target the [Jasmin persona](https://drive.google.com/file/d/14FbBOLCVxwSP6M9BK4Nz1Ir9tzxT8_02/view)
- [ ] Target the [Lucas persona](https://drive.google.com/file/d/1J2vqJxo7ldv7OP_obA9Q-avf0o_ju4Lk/view)

### DIFF
--- a/source/includes/steps-fle-convert-to-a-remote-master-key-aws.yaml
+++ b/source/includes/steps-fle-convert-to-a-remote-master-key-aws.yaml
@@ -17,8 +17,8 @@ content: |
 
   .. important:: Use IAM roles
 
-     When deploying your CSFLE application in a production environment,
-     authenticate your application through an IAM role instead of an IAM user.
+     When deploying your CSFLE-enabled application to a production environment,
+     authorize your application through an IAM role instead of an IAM user.
 
      To authenticate with an IAM role, specify a your temporary IAM role credentials
      in your KMS provider object as follows:
@@ -34,10 +34,10 @@ content: |
      You can get your temporary IAM role credentials through the following mechanisms:
 
      - Call ``AssumeRole``
-     - `Retrieve Credentials from EC2 Instance Metadata <https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/iam-roles-for-amazon-ec2.html#instance-metadata-security-credentials>`
+     - `Retrieve Credentials from EC2 Instance Metadata <https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/iam-roles-for-amazon-ec2.html#instance-metadata-security-credentials>`__
 
      To learn more about IAM roles, see
-     `IAM roles <https://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles.html>`
+     `IAM roles <https://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles.html>`__
      in the AWS documentation.
 
      To learn how to assume a role in each of the languages supported in this guide,
--- a/source/includes/steps-fle-convert-to-a-remote-master-key-aws.yaml
+++ b/source/includes/steps-fle-convert-to-a-remote-master-key-aws.yaml
@@ -99,7 +99,7 @@ content: |
      .. list-table::
         :header-rows: 1
         :stub-columns: 1
-        :widths: 25 10 10 45
+        :widths: 25 15 15 45
 
         * - Field
           - Required for IAM User
--- a/source/includes/steps-fle-convert-to-a-remote-master-key-aws.yaml
+++ b/source/includes/steps-fle-convert-to-a-remote-master-key-aws.yaml
@@ -10,7 +10,7 @@ content: |
      - **access key ID**
      - **secret access key**
 
-  2. Grant the IAM user full ``List`` and ``Read`` permissions for the KMS
+  2. Grant the IAM user narrowly scoped ``List`` and ``Read`` permissions only for the KMS
      service. See Amazon's official documentation on
      `Adding permissions to a user <https://docs.aws.amazon.com/IAM/latest/UserGuide/id_users_change-permissions.html>`__
      to set these permissions.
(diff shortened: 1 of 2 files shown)


source/includes/steps-fle-convert-to-a-remote-master-key-aws.yaml
```diff
@@ -72,6 +108,12 @@ content: |
           - Yes
           - Contains the authentication credentials of the account user
```
zach-carr, line 109: [suggestion] minor, but I suggest adding periods to the end of the first two descriptions to match the Session Token description you added (or removing the period from the Session Token description, whichever way you want to standardize)

source/includes/steps-fle-convert-to-a-remote-master-key-aws.yaml
```diff
@@ -15,6 +15,42 @@ content: |
...
+        {
+          "accessKeyId":"<temporary access key ID>",
+          "secretAccessKey":"<temporary secret access key>",
+          "sessionToken":"<session token>"
+        }
+     
+     You can get your temporary IAM role credentials through the following mechanisms:
+
+     - Call ``AssumeRole``
+     - `Retrieve Credentials from EC2 Instance Metadata <https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/iam-roles-for-amazon-ec2.html#instance-metadata-security-credentials>`__
```
zach-carr, line 37: [question] is it worth explicitly mentioning/ linking to how to update expired temporary credentials?

source/includes/steps-fle-convert-to-a-remote-master-key-aws.yaml
```diff
@@ -15,6 +15,42 @@ content: |
      `Adding permissions to a user <https://docs.aws.amazon.com/IAM/latest/UserGuide/id_users_change-permissions.html>`__
      to set these permissions.
 
+  .. important:: Use IAM roles
```
zach-carr, line 18: [suggestion] minor - "Use IAM roles instead" would be a slightly clearer note title to me on first read - It took some reading to be sure this meant completely replace IAM user usage with roles (at least as far as your app is concerned).

I was going to suggest more language to make it clear you use roles instead of users, but I see there's a mobile app case where it looks like you would still create a user, then map it to a role.

source/includes/steps-fle-convert-to-a-remote-master-key-aws.yaml
```diff
@@ -61,7 +97,7 @@ content: |
         :widths: 30 15 45
 
         * - Field
-          - Required
+          - Required for IAM User
```
zach-carr, line 100: [suggestion] I think this table is a little confusing because whether a field is required is noted differently for users and roles.

Suggest adding a column for "Required for IAM Role" instead of this one "Required" column and the sentence "If you authenticate with an IAM role, you must specify this field." 

Field | Required for IAM Users | Required for IAM Roles | Description
x.      | Yes                                   | Yes
y.      | Yes                                   | Yes
z.      | No                                     | Yes

zach-carr (commented):
Content looks good to me, just left a question that I think is worth talking about + a suggestion for the table. Other stuff is very minor.

source/includes/steps-fle-convert-to-a-remote-master-key-aws.yaml
zach-carr: To clarify - when I was reading, I was wondering "do I still create a user in step 1, then create a role for my app to use in addition? or do I read this note first, then essentially replace "user" with "role" every time I see it in these steps?"

biniona-mongodb: Updated to "Authenticate with IAM Roles in Production", to try and more clearly say in note title exactly what you are using IAM role for.

Context: As mentioned in your message, it's tricky because you don't necessarily replace users with roles (a valid deployment can still require both AFAICT). I think we are trying to outsource most of the actual, "whats a relationship between a user and a role" to the AWS docs, and focus on "how do I use temporary credentials to assume a role" in this guide.

source/includes/steps-fle-convert-to-a-remote-master-key-aws.yaml
biniona-mongodb: Adding content explaining that users must implement logic to handle updating expired temporary credentials.

Context: I think all readers need to be aware of is "MongoDB does not refresh your credentials for you." Actually providing an implementation of how to refresh credentials I think goes too far into AWS world (I think in this case that would be providing code snippets that [request a secure token ](https://docs.aws.amazon.com/code-samples/latest/catalog/javav2-sts-src-main-java-com-example-sts-AssumeRole.java.html)on a 403 error).

If Kenn White or Cynthia want this level of detail we can add it, but I'm not sure it makes sense to do at this time, espcecially that next-gen FLE will handle this automatically for users.

source/includes/steps-fle-convert-to-a-remote-master-key-aws.yaml
biniona-mongodb: Nice suggestion! I think this is cleaner.

source/includes/steps-fle-convert-to-a-remote-master-key-aws.yaml
zach-carr: That makes sense, sounds good. Thanks for the context

zach-carr (approved):
LGTM!

PR #802, opened 2022-03-14
(DOCSP-21351) IAM User Permissions

biniona-mongodb:
## Pull Request Info

Correct/Update guidance around IAM user permissions.

### Issue JIRA link:
https://jira.mongodb.org/browse/DOCSP-21351

### Snooty build log:

https://workerpool-boxgs.mongodbstitch.com/pages/job.html?collName=queue&jobId=6234db880ab54725f130c579

https://docs-mongodbcom-staging.corp.mongodb.com/drivers/docsworker-xlarge/DOCSP-21351-IAM-User-Permissions/security/client-side-field-level-encryption-local-key-to-kms/#create-an-aws-iam-user

### Self-Review Checklist

### If your page documents a concept, does it meet the following criteria?

- [ ] Target the [Jasmin persona](https://drive.google.com/file/d/14FbBOLCVxwSP6M9BK4Nz1Ir9tzxT8_02/view)
- [ ] Target the [Lucas persona](https://drive.google.com/file/d/1J2vqJxo7ldv7OP_obA9Q-avf0o_ju4Lk/view)


biniona-mongodb:
Context for Reviewers:

In the tutorial, I tell readers to do the easiest to perform step rather than the most secure step, and then mention what the most secure step would be in a note.

The most secure step requires a decent amount of familiarity with the AWS UI (knowing that `kms:[En/De]crypt` is a child of `Write`) and having the `ARN` of your Symmetric key (which is created in the next step of the tutorial). Similar to using an IAM user instead of an IAM role in the tutorial, I believe it makes sense to opt for a less-secure but easier to execute system to help readers get to a working state and then mention throughout the tutorial how the system could be made as secure as possible through more complicated (and in this case non-linear (must create user, create key, add user to key, and then go back to user to add key's ARN)) processes.

ccho-mongodb:
> # Context for Reviewers:
> ## IAM Roles
> The "Authenticate with IAM Roles in Production" admonition was moved, but did not have its content changed. Any comments on the content of the note may be best discussed in a separate PR.

Issue: to me, the new placement is no longer sequential and interrupts the sequential flow of actions. The reader must first set up the User to be able to assume the Role. 
Suggestion: Create a new sub-step that fits in sequentially on how to authenticate with an IAM Role instead of placing that info in an admonition where it might not fit sequentially.

> ## Decision to mention `Write` instead of `kms:Encrypt` and `kms:Decrypt` in main Tutorial
> In the tutorial, I tell readers to do the easiest to perform permission-granting procedure rather than the most secure procedure, and then mention what the most secure procedure would be in a note.

Issue: It's usually not reasonable to depend on anyone remembering to update something for production if nothing breaks. I think the "Least Privilege" practice is popular in security/administration, so we should recommend that as the primary path. The admonition already seems more prominent than the step description itself.

Suggestion: 1) Move "grant your IAM user Write permissions" to a tip admonition for readers seeking to take a shortcut rather than suggesting it in the main flow, or 2) remove it to avoid confusion since you already have instructions on how to grant kms:Encrypt/Decrypt. Defer any complex AWS processes to AWS documentation if necessary, similar to the other admonition in this PR.

source/includes/iam-user-policy-minimum.json
```diff
@@ -0,0 +1,10 @@
+{
+  "Version": "2012-10-17",
+  "Statement": [
+    {
+      "Effect": "Allow",
+      "Action": ["kms:Decrypt", "kms:Encrypt"],
+      "Resource": "<the Amazon resource name (ARN) of your symmetric key>"
```
ccho-mongodb, line 7: Issue: I think we refer to it as "remote master key" rather than the more general descriptive term "symmetric key". Since there's no connection established between the two terms, there's a better chance the reader will recognize the one that was introduced.
Also, I think ARN is a proper noun(see https://docs.aws.amazon.com/general/latest/gr/aws-arns-and-namespaces.html), especially when used before an all-caps abbreviation.

```suggestion
      "Resource": "<the Amazon Resource Name (ARN) of your remote master key>"
```

ccho-mongodb (changes requested):
Seems like you noticed some issues with readability when making changes to the permissions required, which is a much smaller change. I would suggest just making enough changes to fulfill the PR requirements for this, and also creating a ticket to address the clarity of the steps, including the comments I've left here.

source/includes/iam-user-policy-minimum.json
```diff
@@ -0,0 +1,10 @@
+{
+  "Version": "2012-10-17",
```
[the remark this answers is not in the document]
biniona-mongodb: I believe there [are only 2 valid versions](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_elements_version.html) for the version variable in an IAM JSON policy (`2012-10-17` or `2008-10-17`). As AWS supplies this version for (AFAICT) all of their [sample inline policies](https://docs.aws.amazon.com/IAM/latest/UserGuide/access_policies_examples.html) (even policies that do not make use of [policy variables](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_examples_iam_policy-sim.html)), I believe it would be idiomatic for us to supply this value as well.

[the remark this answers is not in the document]
ccho-mongodb: I see, I was confused by the policy version which is not the same as this (Version policy element).

biniona-mongodb:
I created the following tickets to update the tutorial:

- [Remove non-linearity introduced in tutorial in using the least privilege IAM policy](https://jira.mongodb.org/browse/DOCSP-21440)
- [Use IAM roles in tutorial](https://jira.mongodb.org/browse/DOCSP-21439)

I added these tasks to the CSFLE merge epic and market them as to be completed AFTER the [CSFLE + KMS  - AWS](https://jira.mongodb.org/browse/DOCSP-21206) tutorial is migrated.

source/includes/steps-fle-convert-to-a-remote-master-key-aws.yaml
```diff
@@ -10,10 +10,22 @@ content: |
...
+     policy allows an IAM user to perform CSFLE with the least privileges
+     possible:
+
+     .. literalinclude:: /includes/iam-user-policy-minimum.json
+        :language: json
+
+     .. note:: Remote Master Key Amazon Resource Name (ARN)
+
+        The preceding policy requires the ARN of the key you generate in the
+        :ref:`Create the Master Key <aws-create-master-key>` step of this guide.
```
ccho-mongodb, line 24: Issue:
I noticed the step to which this links mentions the following in sub-step 2:
"In the Usage Permissions step of the key generation process, add the full KMS List and Read permissions to the IAM user you created in the previous step. This authorizes the user to encrypt and decrypt the new master key."
This conflicts with the text you've added in this step.

Suggestion: There doesn't need to be any skipping around for the tutorial. You can apply a policy to an IAM user after you create it, so the example policy and info about user permissions can be moved here. At least remove the conflicting/duplicate information from the create master key sub-step and either address or ticket the rest, which does seem in-scope for this ticket.

ccho-mongodb (changes requested):

source/includes/steps-fle-convert-to-a-remote-master-key-aws.yaml
biniona-mongodb: The policy created in the step to which this links is a [Key Policy](https://docs.aws.amazon.com/kms/latest/developerguide/key-policies.html), where as the step created for the user is an [IAM policy.](https://docs.aws.amazon.com/IAM/latest/UserGuide/access_policies_create.html)

AFAICT, you can configure a Key Policy to defer to an IAM policy for granting access, but you MUST grant an IAM user access to KMS operations, regardless of what is contained in the Key policy.

What is currently on the page seems to incorrectly discuss the two policies as equivalent (and doesn't mention that the default AWS KMS policy defers to IAM policies). I've updated the text to more clearly distinguish between IAM and KMS policies.

ccho-mongodb: I see, that makes sense. It wasn't clear to me in the description, and I think the policy listed there was for the IAM user in the revision I commented on. Let me know if you want another copy review if you incorporate any changes based on the info you receive from the external review.

source/includes/steps-fle-convert-to-a-remote-master-key-aws.yaml
```diff
@@ -10,10 +10,22 @@ content: |
...
+
+     To learn how to set this permission, see
+     `Changing permissions for an IAM user <https://docs.aws.amazon.com/IAM/latest/UserGuide/id_users_change-permissions.html>`__
+     from AWS.
```
ccho-mongodb, line 28: Issue: it's not clear what "from AWS" means.
Suggestion: clarify what the link refers to (e.g. "see the AWS documentation on Changing permissions..." or omit it.

source/includes/steps-fle-convert-to-a-remote-master-key-aws.yaml
```diff
@@ -66,15 +78,29 @@ content: |
...
-     encrypt and decrypt the new master key.
+     process, use the following default key policy to enable IAM policies
+     to allow access to your remote master key:
+
+     .. literalinclude:: /includes/key-policy-default.json
+        :language: json
+
+     .. note:: Key Policies and IAM policies
+
+        If you do not wish for your IAM policies to grant key access permissions,
```
ccho-mongodb, line 97: Issue: I'm not entirely sure I understand this note, or whether what it conveys is accurate. My impression from reading it is that it is stating some way to more granularly grant permissions to encrypt/decrypt using the key to individual IAM users.
Suggestion: If I understand correctly, this might communicate the message better: "If you intend to grant only specific users access to the key, you must grant access to each IAM user in the key policy. Otherwise, all IAM users can be granted access to a key through an IAM policy". If I don't understand correctly, feel free to ignore this comment.

ccho-mongodb (approved):
A couple minor issues, but otherwise LGTM

source/includes/steps-fle-convert-to-a-remote-master-key-aws.yaml
biniona-mongodb: The note does not really pose a more granular way to grant access, but more a way to make it such that you cannot add additional users to a key (AFAICT this also makes it possible for an AWS administrator to be unable to view a key they created in a provable way).

After sitting on this for a bit, I think it may make sense to remove this and instead just link to AWS documentation. This seems like a bit of an edge case, and I'm not sure if we should discuss it. If an external reviewer thinks this explanation could benefit readers I'll add it back, but at present I think it is alright to omit it.

source/includes/steps-fle-convert-to-a-remote-master-key-aws.yaml
biniona-mongodb: Changed "from AWS" to "in the official AWS documentation". I believe the "To learn more, see <title of other page> in <destination>" is the preferred structure of external links at the end of sections in the [style guide](https://docs.mongodb.com/meta/style-guide/style/links-and-cross-references/#link-and-cross-reference-text), so I think it's best to keep this structure. 

Relevant items from [links and cross references](https://docs.mongodb.com/meta/style-guide/style/links-and-cross-references/#link-and-cross-reference-text) entry of the style guide:

> For links at the end of an article or topic that point to related information or to a next step, use the title of or a heading in the target content as the link text.

> If a link points to a location other than the current site (for example, out of the Support website or away from docs.mongodb.com), provide context that describes the location.

biniona-mongodb:
Note for technical reviewer: Could you confirm that it is correct to recommend [the default key policy](https://docs.aws.amazon.com/kms/latest/developerguide/key-policy-default.html#key-policy-default-allow-root-enable-iam) for this tutorial as well as a production deployment?

PR #846, opened 2022-11-21
DOCS-15747 Update PyMongo Installation Instructions 

ajhuh-mdb:
# Pull Request Info

JIRA - <https://jira.mongodb.org/browse/DOCS-15747>
Staging - <https://docs-mongodbcom-staging.corp.mongodb.com/drivers/docsworker-xlarge/DOCS-15747-update-pymongo-installation/pymongo/#installation>

## Self-Review Checklist

### DIFF
--- a/source/pymongo.txt
+++ b/source/pymongo.txt
@@ -44,7 +44,8 @@ Python application. We recommend using
 
 PyMongo requires `dnspython <https://pypi.org/project/dnspython>`__  to support 
 ``mongodb+srv://`` connection strings and `MongoDB Atlas 
-<https://www.mongodb.com/cloud>`__.
+<https://www.mongodb.com/cloud>`__. dnspython is automatically installed when 
+installing or upgrading to the latest PyMongo version. 
 
 The following command demonstrates how you can install the latest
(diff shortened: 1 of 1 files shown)


source/pymongo.txt
```diff
@@ -42,27 +42,31 @@ You must install the PyMongo driver module to make it available to your
 Python application. We recommend using
 `pip <http://pypi.python.org/pypi/pip>`__ to install PyMongo.
 
+PyMongo requires `dnspython <https://pypi.org/project/dnspython>`__  to support 
+``mongodb+srv://`` connection strings and `MongoDB Atlas 
+<https://www.mongodb.com/cloud>`__.
```
ccho-mongodb, line 47: Question: Does making it a "required dependency" as stated in the ticket mean they need to explicitly install it or does it get installed when you run pip?
If the former, would it be possible to show an example of how to install it for readers that might not know as well as those that want the copy-paste convenience?
If the latter, could that be stated here?

ccho-mongodb (approved):
LGTM, though would want clarification on the new requirement.

PR #860, opened 2023-03-15
Docsp 26722: Adding TypeScript landing page

DBirtolo-mdb:
# Pull Request Info

JIRA - <https://jira.mongodb.org/browse/DOCSP-26722>
Staging - <https://docs-mongodbcom-staging.corp.mongodb.com/drivers/docsworker-xlarge/DOCSP-26722/typescript/>

## Self-Review Checklist

### DIFF
--- a/snooty.toml
+++ b/snooty.toml
@@ -9,8 +9,8 @@ sharedinclude_root = "https://raw.githubusercontent.com/10gen/docs-shared/main/"
 [constants]
 package-name-org = "docs-ecosystem"
 stable-api = "Stable API"
-node-driver-long = "MongoDB Node.js driver"
-node-driver-short = "Node.js driver"
+node-driver-long = "MongoDB Node.js Driver"
+node-driver-short = "Node.js Driver"
 
 csharp-version = "2.18"
 csharp-docs-versioned = "https://mongodb.github.io/mongo-csharp-driver/{+csharp-version+}"
--- a/snooty.toml
+++ b/snooty.toml
@@ -9,6 +9,8 @@ sharedinclude_root = "https://raw.githubusercontent.com/10gen/docs-shared/main/"
 [constants]
 package-name-org = "docs-ecosystem"
 stable-api = "Stable API"
+node-driver-long = "MongoDB Node.js driver"
+node-driver-short = "Node.js driver"
 
 csharp-version = "2.18"
 csharp-docs-versioned = "https://mongodb.github.io/mongo-csharp-driver/{+csharp-version+}"
--- a/source/typescript.txt
+++ b/source/typescript.txt
@@ -26,4 +26,4 @@ TypeScript types with MongoDB.
 To learn more about the features and limitations of using TypeScript with the
 {+node-driver-short+}, see
 `TypeScript Fundamentals <https://www.mongodb.com/docs/drivers/node/current/fundamentals/typescript/>`__
-in the Node.js Driver documentation.
+in the {+node-driver-short+} documentation.
--- a/source/typescript.txt
+++ b/source/typescript.txt
@@ -1,0 +1,35 @@
+.. _typescript-center:
+
+==========
+TypeScript
+==========
+
+.. default-domain:: mongodb
+
+.. contents:: On this page
+   :local:
+   :backlinks: none
+   :depth: 1
+   :class: twocols
+
+Introduction
+------------
+
+Welcome to the documentation site for official MongoDB TypeScript support.
+Since Typescript builds on JavaScript, you can use the
+`Node.js Driver <https://www.mongodb.com/docs/drivers/node/current/>`__
+to create TypeScript applications that communicate with MongoDB.
+
+The Node.js Driver started supporting TypeScript with version 4.0. You must
+install that version of the driver or a more recent one to use TypeScript.
+
+To learn more about the features and limitations of using TypeScript with the
+Node.js Driver, see
+`Java Driver <https://www.mongodb.com/docs/drivers/node/current/fundamentals/typescript/>`__.
+
+Compatibility
+-------------
+
+For the compatibility charts for the Node.js Driver, see
+`Compatibility <https://www.mongodb.com/docs/drivers/node/current/compatibility/>`__
+in the driver docs.
--- a/source/typescript.txt
+++ b/source/typescript.txt
@@ -16,12 +16,12 @@ Introduction
 ------------
 
 Welcome to the documentation site for official MongoDB TypeScript support.
-Since TypeScript builds on JavaScript, you can use the
+Since TypeScript compiles to JavaScript, you can use the
 `{+node-driver-long+} <https://www.mongodb.com/docs/drivers/node/current/>`__
 to create TypeScript applications that communicate with MongoDB.
 
-Versions 4.0 and later of the {+node-driver-short+} support TypeScript. You must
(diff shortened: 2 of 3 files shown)


source/typescript.txt
```diff
@@ -0,0 +1,35 @@
...
+   :local:
+   :backlinks: none
+   :depth: 1
+   :class: twocols
+
+Introduction
+------------
+
+Welcome to the documentation site for official MongoDB TypeScript support.
+Since Typescript builds on JavaScript, you can use the
```
mongoKart, line 19:
```suggestion
Since TypeScript builds on JavaScript, you can use the
```

source/typescript.txt
```diff
@@ -0,0 +1,35 @@
...
+`Node.js Driver <https://www.mongodb.com/docs/drivers/node/current/>`__
```
mongoKart, line 20: S: This sounds better to me as "MongoDB Node.js Driver" in this sentence. There might already be a source constant in docs-node to this effect.

mongoKart, line 20: Q: On a similar note, is there a source constant for "Node.js Driver"? We use those quite frequently in other repos and might be worth doing the same here.

source/typescript.txt
```diff
@@ -0,0 +1,35 @@
...
+to create TypeScript applications that communicate with MongoDB.
+
+The Node.js Driver started supporting TypeScript with version 4.0. You must
```
mongoKart, line 23: S: "started supporting" makes it sound like it hasn't finished supporting TypeScript (which, from a technical perspective, is probably accurate, because when is support ever _totally_ complete, but it makes it sound like we have a half-baked driver).

```suggestion
The Node.js Driver began support for TypeScript with version 4.0. You must
```

Or:

```suggestion
Support for TypeScript began in version 4.0 of the Node.js Driver. You must
```

Or:

```suggestion
Versions 4.0 and above of the Node.js Driver support TypeScript. You must
```

source/typescript.txt
```diff
@@ -0,0 +1,35 @@
...
+install that version of the driver or a more recent one to use TypeScript.
+
+To learn more about the features and limitations of using TypeScript with the
+Node.js Driver, see
+`Java Driver <https://www.mongodb.com/docs/drivers/node/current/fundamentals/typescript/>`__.
```
mongoKart, line 28: I: Wrong link title

S: 

```suggestion
the `TypeScript Fundamentals page <https://www.mongodb.com/docs/drivers/node/current/fundamentals/typescript/>`__ in the
Node.js Driver documentation.
```

source/typescript.txt
```diff
@@ -0,0 +1,35 @@
...
+
+Compatibility
+-------------
+
+For the compatibility charts for the Node.js Driver, see
+`Compatibility <https://www.mongodb.com/docs/drivers/node/current/compatibility/>`__
+in the driver docs.
```
mongoKart, line 35: This is a tricky one. When I hear about compatibility w/r/t TypeScript, my mind jumps to a version of JS (or, in this case, the version of the Node.js driver that we've spent the whole page talking about). 

S: Remove the Compatibility section altogether. Open to talking this out, too!

source/typescript.txt
mongoKart, line 1: Q: The Jira ticket includes the following A/C:

_As a user, I can read about Typescript typing on the mongodb.com/drivers/node/ landing page (for relevant versions)._

I'm not sure what Chris was envisioning here, but might be worth clarifying with him if you haven't already.

mongoKart (changes requested):
Looks good! Some minor suggestions and questions.

source/typescript.txt
DBirtolo-mdb: I went ahead and created constants that match the constants in the node driver. Thanks for that idea!

source/typescript.txt
DBirtolo-mdb: I will double-check, but this is already covered to my understanding since we have a link on the landing node page that goes to the TypeScript page for [Specify Type Parameters with TypeScript](https://www.mongodb.com/docs/drivers/node/current/fundamentals/typescript/). The landing page is pretty light and mainly link-based.

snooty.toml
```diff
@@ -9,6 +9,8 @@ sharedinclude_root = "https://raw.githubusercontent.com/10gen/docs-shared/main/"
 [constants]
 package-name-org = "docs-ecosystem"
 stable-api = "Stable API"
+node-driver-long = "MongoDB Node.js driver"
+node-driver-short = "Node.js driver"
```
mongoKart, line 13: Really small thing: Because these are technically a product name (as opposed to a driver in the abstract), it should be capital-D.

```suggestion
node-driver-long = "MongoDB Node.js Driver"
node-driver-short = "Node.js Driver"
```

source/typescript.txt
mongoKart: As long as it's what Chris was envisioning when he created the ticket, it's good with me.

source/typescript.txt
```diff
@@ -16,20 +16,14 @@ Introduction
...
-
-Compatibility
--------------
-
-For the compatibility charts for the Node.js Driver, see
-`Compatibility <https://www.mongodb.com/docs/drivers/node/current/compatibility/>`__
-in the driver docs.
+{+node-driver-short+}, see
+`TypeScript Fundamentals <https://www.mongodb.com/docs/drivers/node/current/fundamentals/typescript/>`__
+in the Node.js Driver documentation.
```
mongoKart, line 29: S: Use source constant here too

```suggestion
in the {+node-driver-short+} documentation.
```

mongoKart (approved):
LGTM + 2 small things!

source/typescript.txt
```diff
@@ -0,0 +1,29 @@
...
+   :local:
+   :backlinks: none
+   :depth: 1
+   :class: twocols
+
+Introduction
+------------
+
+Welcome to the documentation site for official MongoDB TypeScript support.
+Since TypeScript builds on JavaScript, you can use the
```
ccho-mongodb, line 19: Question:
I personally haven't heard "builds on" as the relationship and it seems like an ambiguous one. Usually, I see "transpiles to" or "compiles to" since there's a conversion process that happens to turn TypeScript into JavaScript before you can run it on a JS engine.

source/typescript.txt
DBirtolo-mdb: I did double check with him and he is good with this solution.

source/typescript.txt
```diff
@@ -0,0 +1,29 @@
...
+`{+node-driver-long+} <https://www.mongodb.com/docs/drivers/node/current/>`__
+to create TypeScript applications that communicate with MongoDB.
+
+Versions 4.0 and later of the {+node-driver-short+} support TypeScript. You must
+install that version of the driver or a more recent one to use TypeScript.
```
ccho-mongodb, line 24: Suggestion:

I think "support Typescript" could be more precise -- it's possible for any Javascript to "support" it by using the "any" type. Instead, it could be better to mention that the driver supports using Typescript types with it in your app.

I think the two sentences could also be combined into one.


```suggestion
You must use v4.0 or later of the driver to use Typescript types in your app. 
```

source/typescript.txt
mongoKart: @ccho-mongodb FWIW, I thought through the same thing during my review but ended up being okay with it. That said, I'm also fine with changing it.

PR #895, opened 2023-07-07
Create CONTRUBUTING.md

rustagir:
Creating a contributions guide per team sync


GromNaN (approved):
Very clear. Thank you for documenting the documentation 😄

ccho-mongodb (changes requested):
Thanks for starting this PR!

I have a number of comments that will likely lead to replacing this file, so I'll list them here.

I think a good starting point would be to read the following guides on "Contributing" files for open source projects/github repositories:
- https://mozillascience.github.io/working-open-workshop/contributing/
- https://docs.github.com/en/communities/setting-up-your-project-for-healthy-contributions/setting-guidelines-for-repository-contributors

Another resource you might consider looking at if you haven't seen it already. Based on the guidance provided in the prior links, I think it doesn't resemble what we want, but I thought it could be helpful to build awareness of it:
https://github.com/mongodb/docs-realm/blob/master/CONTRIBUTING.md
Note: the contributor agreement link on that page is broken, so don't use it. Perhaps it could be good to explain that the docs, while open source, is managed by a team at MongoDB.

I think it would be helpful to and provide more context. Consider the perspective of someone that might not have any experience using any of the tools we use nor the workflow we follow.

Here's a list of notes I took on areas that might be useful to address:
- Help the reader confirm whether they are the intended audience
- Help reader understand that they may not be able to test their change
- Who are the team members that manage the documentation repo and who they should contact?
- Describe our bug tracking system, understanding that a reader may not have ever used JIRA
- Searching JIRA might be a complex task and put people off from creating issues. Perhaps assume this is our responsibility instead.
- Explain what content we expect them to include in the ticket.

- Bug/issue - what is the difference to an outside reader?
- Explain why they can’t contribute directly
- What kind of response time should they expect once they contact the team?
- Provide more detailed instructions on how they should open the pull request.
- Explain the workflow of reviewing and/or migrating their changes into a team member's PR
- List important links (and describe them) somewhere
- Favor imperative mood to provide direction and help keep sentences brief

PR #904, opened 2023-09-06
DOCSP-32483 programmatic tagging script for ecosystem repo

sarahemlin:
JIRA - <https://jira.mongodb.org/browse/DOCSP-32483>

Hey @mongoKart - here's the facet.toml file for the Drivers ecosystem repo.  Please merge if everything's ok!

Thanks!
Sarah

### DIFF
--- a/source/facets.toml
+++ b/source/facets.toml
@@ -1,0 +1,7 @@
+[[facets]]
+category = "target_product"
+value = "drivers"
+
+[[facets]]
+category = "programming_language"
+value = "kotlin"
--- a/source/facets.toml
+++ b/source/facets.toml
@@ -1,7 +1,3 @@
 [[facets]]
 category = "target_product"
 value = "drivers"
-
-[[facets]]
-category = "programming_language"
-value = "kotlin"


source/facets.toml
```diff
@@ -0,0 +1,7 @@
+[[facets]]
+category = "target_product"
+value = "drivers"
+
+[[facets]]
+category = "programming_language"
+value = "kotlin"
```
mongoKart, line 7: This repo contains the landing pages for all of our drivers. I think this should be removed (or, if possible and useful, add a facet for every programming language on the page).

sarahemlin: Thanks for the background info, @mongoKart.  I agree; let's close out this PR, and we'll need to take the landing pages individually by hand.  The script doesn't seem like a good fit.

sarahemlin:
@mongoKart, I moved to quickly and forgot that we would want to tag everything 'drivers' but NOT tag for language via script.  Language tags need to be done per landing page.  I'll update the file to remove the language and then you can merge!

mongoKart (approved):
lgtm

PR #911, opened 2023-09-29
DOCSP-32875: add PyMongoArrow

ccho-mongodb:
# Pull Request Info

JIRA - <https://jira.mongodb.org/browse/DOCSP-32875>
Staging - https://preview-mongodbcchomongodb.gatsbyjs.io/drivers/DOCSP-32875-pymongoarrow/python/
## Self-Review Checklist

### DIFF
--- a/source/python.txt
+++ b/source/python.txt
@@ -29,7 +29,8 @@ MongoDB in Python:
 
 - :doc:`PyMongo </pymongo>` for synchronous Python applications
 
-  - `PyMongoArrow <https://mongo-arrow.readthedocs.io/en/latest/>`__, a PyMongo extension for loading Apache Arrow tables by using MongoDB queries
+  - `PyMongoArrow <https://mongo-arrow.readthedocs.io/en/latest/>`__, a 
+    PyMongo extension for loading MongoDB query result sets into Apache Arrow tables
 
 - :doc:`Motor </motor>` for asynchronous Python applications
 
--- a/source/python.txt
+++ b/source/python.txt
@@ -29,8 +29,8 @@ MongoDB in Python:
 
 - :doc:`PyMongo </pymongo>` for synchronous Python applications
 
-  - `PyMongoArrow <https://mongo-arrow.readthedocs.io/en/latest/>`__, a 
-    PyMongo extension for loading MongoDB query result sets into Apache Arrow tables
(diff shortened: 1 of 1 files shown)


source/python.txt
```diff
@@ -23,12 +23,15 @@ MongoDB Python Drivers
...
+Welcome to the documentation site for the official MongoDB Python
+drivers. Add one of the following drivers to your application to work with
+MongoDB in Python:
 
-- Use :doc:`PyMongo </pymongo>` for synchronous Python applications.
+- :doc:`PyMongo </pymongo>` for synchronous Python applications
 
-- Use :doc:`Motor </motor>` for asynchronous Python applications.
+  - `PyMongoArrow <https://mongo-arrow.readthedocs.io/en/latest/>`__, a 
+    PyMongo extension for loading MongoDB query result sets into Apache Arrow tables
```
rustagir, line 33: S: use the wording on the PyMongoArrow page

```suggestion
    PyMongo extension for loading MongoDB query result sets as Apache Arrow tables
```

source/python.txt
```diff
@@ -23,12 +23,15 @@ MongoDB Python Drivers
...
 
-Welcome to the documentation site for the official MongoDB Python 
-drivers. You can add one of the following drivers to your application 
-to work with MongoDB in Python:
+Welcome to the documentation site for the official MongoDB Python
+drivers. Add one of the following drivers to your application to work with
+MongoDB in Python:
 
-- Use :doc:`PyMongo </pymongo>` for synchronous Python applications.
+- :doc:`PyMongo </pymongo>` for synchronous Python applications
```
rustagir, line 30: S: would this list be better presented as a list of the drivers/extensions with colons and descriptions? I think that would flow better than the different sentence structures in each list item right now. Might require editing the introductory sentence

rustagir (approved):
approving with some small comments

source/python.txt
ccho-mongodb: I considered that, but I think it would require adding more information into the introduction or creating a separate section for extensions, which would make the page less concise. Since PyMongoArrow is a nested list item, I don't think it needs to be parallel with the top-level ones, which are described as drivers. I think if we needed to add more Python drivers or extensions, it might be worthwhile creating a new section.

PR #922, opened 2023-10-17
DOCSP-33692: php improvements

rustagir:
# Pull Request Info

JIRA - <https://jira.mongodb.org/browse/DOCSP-33692>
Staging - https://preview-mongodbrustagir.gatsbyjs.io/drivers/DOCSP-33692-php-arq-updates/php/#driver-architecture

## Self-Review Checklist


source/php.txt
```diff
@@ -72,6 +72,44 @@ tutorials.
...
+The following diagram illustrates the architecture of the PHP driver
+components:
+
+.. figure:: /figures/PHP_driver_architecture.svg
+   :alt: PHP driver component architecture
+   :figwidth: 600px
+   
+The PHP library provides an API that is consistent with the other
+drivers, and it is continually updated to meet cross-driver
+specifications. You must add the library as a dependency in most
```
jordan-smith721, line 96: S: I think this would be more clear as separate sentences

```suggestion
The PHP library provides an API that is consistent with the other
MongoDB drivers. The library is continually updated to meet cross-driver
specifications. You must add the library as a dependency in most
```

source/php.txt
```diff
@@ -72,6 +72,44 @@ tutorials.
...
+applications that use PHP to work with MongoDB.
```
jordan-smith721, line 97: The phrasing on this is a bit confusing to me. One question and one suggestion:

Q: This states that it's needed in "most applications" - does that mean that there are some instances where you don't need to the library in order to work with MongoDB in a PHP application? Or can we get rid of "most"?

S: If I'm understanding what's being said here, I think this could be a little more clear
```suggestion
specifications. You must add the library as a dependency to work with MongoDB in most
PHP applications.
```

jordan-smith721 (approved):
LGTM with a question and a couple minor suggestions

source/php.txt
rustagir: I think that there might be cases where a user could use only the extension to perform operations in their application, but this is not recommended.

source/php.txt
```diff
@@ -72,6 +72,44 @@ tutorials.
...
+.. figure:: /figures/PHP_driver_architecture.svg
+   :alt: PHP driver component architecture
+   :figwidth: 600px
+   
+The PHP library provides an API that is consistent with the other
+MongoDB drivers. The library is continually updated to meet cross-driver
+specifications. You must add the library as a dependency to access
+MongoDB in most PHP applications.
+
+The extension is distributed by `PECL
```
alcaeus, line 99:
```suggestion
The extension is distributed using `PECL
```

alcaeus (approved):
LGTM, just a small suggestion.

bisht2050 (approved):

PR #933, opened 2023-11-15
DOCSP-34438: php v1.17 release

rustagir:
# Pull Request Info

JIRA - <https://jira.mongodb.org/browse/DOCSP-34438>
Staging - https://preview-mongodbrustagir.gatsbyjs.io/drivers/DOCSP-34438-php-v1.17-compat/php-drivers/#compatibility

## Self-Review Checklist

- [x] Are the [facets and meta keywords](https://wiki.corp.mongodb.com/display/DE/Docs+Taxonomy) accurate?


source/includes/mongodb-compatibility-table-php.rst
```diff
@@ -18,6 +18,19 @@
      - MongoDB 3.0
      - MongoDB 2.6
 
+   * - ext + lib 1.17
```
mongoKart, line 21: S: update line 34 to "ext + lib 1.16 to 1.17" instead

mongoKart (approved):
lgtm + 1 suggestion

source/includes/mongodb-compatibility-table-php.rst
```diff
@@ -18,7 +18,7 @@
      - MongoDB 3.0
      - MongoDB 2.6
 
-   * - ext + lib 1.16
+   * - ext + lib 1.16 to 1.17
```
jmikola, line 21: Confirmed that 1.16 and 1.17 have the same server compatibility.

This is the first use of a range of versions AFAIK, but I trust this formatting is consistent with what is done for other drivers. I'm also a bit surprised that this is the first case of multiple versions having the same server compatibility (unless that's not the case but it just would have been strange to present a range like this given the ext/lib versions were previously out of sync).

jmikola (approved):

source/includes/mongodb-compatibility-table-php.rst
rustagir: We've recently been displaying versions with the same compatibility like this to avoid tables from becoming large. Yes, in the future we can keep condensing the PHP compat when it is the same as prev versions, but the versions being out of sync makes that too visually confusing.

PR #962, opened 2024-04-16
fix java-rs links

mmeigs:
# Pull Request Info

This PR fixes all links to the docs-java-rs property because of a change coming from [this PR](https://github.com/mongodb/docs-java-rs/pull/27) to correct its url path.


rustagir (approved):
lgtm

PR #971, opened 2024-05-24
DOCSP-39180: Symfony Quick Start

rustagir:
# Pull Request Info

JIRA - <https://jira.mongodb.org/browse/DOCSP-39180>
Staging - <https://preview-mongodbrustagir.gatsbyjs.io/drivers/DOCS-39180-symfony-qs/php-frameworks/symfony/>

## Self-Review Checklist

- [x] Are the [facets and meta keywords](https://wiki.corp.mongodb.com/display/DE/Docs+Taxonomy) accurate?


source/php-frameworks/symfony.txt
```diff
@@ -0,0 +1,334 @@
...
+connects to your MongoDB deployment, runs a query on
+the sample data, and renders a retrieved result.
+
+Learn more about Symfony and MongoDB by viewing the following resources:
+
+- `Build PHP Symfony Apps with MongoDB Atlas Workshop
+  <https://mongodb-developer.github.io/symfony-mongodb-rental-workshop/>`__
+- `Symfony Documentation <https://symfony.com/doc/current/index.html>`__
+- `Doctrine ODM Documentation
+  <https://www.doctrine-project.org/projects/doctrine-mongodb-odm/en/latest/index.html>`__
```
bisht2050, line 334: Suggest to add https://github.com/doctrine/mongodb-odm as well

source/php-frameworks/symfony.txt
```diff
@@ -0,0 +1,334 @@
...
+
+.. code-block:: none
+   :copyable: false
+
+   [OK] Web server listening
+        The Web server is using PHP FPM 8.3.4
+        http://127.0.0.1:8000
+
+Open the URL http://127.0.0.1:8000/restaurant/browse in your web browser.
+The page shows a list of restaurants and details about each of them.
```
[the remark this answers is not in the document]
norareidy: Adding `use App\Document\Restaurant;` to the controller file resolved the issue!

source/php-frameworks/symfony.txt
```diff
@@ -0,0 +1,334 @@
...
+Run the following commands to install the MongoDB PHP driver and the
+MongoDB Doctrine ODM:
+
+.. code-block:: bash
+
+   composer require mongodb/mongodb
+   composer require doctrine/mongodb-odm-bundle
+
+Next, create a ``src/Document`` directory to store the entities
+that represent your MongoDB documents:
```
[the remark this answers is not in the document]
rustagir: removed

source/php-frameworks/symfony.txt
```diff
@@ -0,0 +1,334 @@
...
+Next, create the ``RestaurantController.php`` file in the
+``src/Controller`` directory to handle the endpoints in your
+application:
+
+.. literalinclude:: /includes/php-frameworks/RestaurantController.php
+   :language: php
+
+The controller file defines the ``index()`` method, which shows some text on
+the web app's front page. The file also defines the ``browse()`` method,
+which finds documents that match documents in which the ``borough``
```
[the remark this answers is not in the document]
rustagir: oops ty

source/php-frameworks/symfony.txt
```diff
@@ -0,0 +1,387 @@
...
+to streamline your web app.
+
+The :ref:`php-symfony-qs` section of this guide contains a tutorial
+which you can follow to build a single page app that accesses data from
+a MongoDB collection.
+
+The :ref:`php-symfony-resources` section contains links to further
+learning resources and documentation.
+
+Advantages of Using MongoDB with Symfony
```
bisht2050, line 40:
```suggestion
Why use MongoDB with Symfony?
```

Suggest to use this SEO optimized title

source/php-frameworks/symfony.txt
```diff
@@ -0,0 +1,387 @@
...
+.. code-block:: bash
+
+   cd restaurants
+   pecl install mongodb
+
+Install PHP Driver and Doctrine ODM
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Run the following commands to install the MongoDB PHP driver and the
+MongoDB Doctrine ODM:
```
bisht2050, line 143: Suggest to add a line that talks about the mongodb-odm-bundle and why we are installing it instead of directly installing mongodb-odm

Quoting the reason from github: "This bundle integrates the Doctrine MongoDB Object Document Mapper (ODM) library into Symfony so that you can persist and retrieve objects to and from MongoDB."

Also add a hyperlink to it - https://github.com/doctrine/DoctrineMongoDBBundle

source/php-frameworks/symfony.txt
bisht2050: One more which I missed earlier - https://github.com/doctrine/DoctrineMongoDBBundle

source/php-frameworks/symfony.txt
```diff
@@ -0,0 +1,387 @@
...
+===========================
+Symfony MongoDB Integration
+===========================
+
+.. facet::
+   :name: genre
+   :values: tutorial
+
+.. meta::
+   :keywords: php framework, odm, web app
```
bisht2050, line 12: Should we also add `symfony`?

source/php-frameworks/symfony.txt
```diff
@@ -0,0 +1,387 @@
...
+   If you prefer to connect to MongoDB by using the MongoDB PHP Library without
+   Symfony, see `Connecting to MongoDB
+   <https://www.mongodb.com/docs/php-library/current/tutorial/connecting/>`__
+   in the MongoDB PHP Library documentation.
+
+MongoDB Atlas is a fully managed cloud database service that hosts your
+MongoDB deployments. You can create your own free (no credit card
+required) MongoDB Atlas deployment by following the steps in this guide.
+
+This guide uses MongoDB Doctrine ODM to allow you to map PHP objects to
```
norareidy, line 80: S: add "the"
```suggestion
This guide uses the MongoDB Doctrine ODM to allow you to map PHP objects to
```

source/php-frameworks/symfony.txt
```diff
@@ -0,0 +1,387 @@
...
+   composer require doctrine/mongodb-odm-bundle
+
+.. tip::
+
+   After running the preceding commands, you might encounter the
+   following prompt:
+
+   .. code-block:: bash
+
+      Do you want to execute this recipe?
```
norareidy, line 157: S: [nit] I don't think this code needs to be copyable
```suggestion
   .. code-block:: bash
      :copyable: false

      Do you want to execute this recipe?
```

norareidy (approved):
LGTM with some minor suggestions!

rustagir:
@rishitb-mongodb This PR is ready for a technical review by the PHP team.

rishitb-mongodb:
> @rishitb-mongodb This PR is ready for a technical review by the PHP team.
I think you meant @bisht2050 :)

source/php-frameworks/symfony.txt
```diff
@@ -0,0 +1,397 @@
...
+
+.. code-block:: yaml
+   :caption: config/packages/doctrine_mongodb.yaml
+
+   doctrine_mongodb:
+       auto_generate_proxy_classes: true
+       auto_generate_hydrator_classes: true
+       connections:
+           default:
+               server: '%env(resolve:MONGODB_URL)%&appName=devrel.content.php'
```
jmikola, line 198: `&appName=devrel.content.php` doesn't seem relevant to readers at all.

The preceding `&` also assumes that the connection string already has options in it. This may be the case for Atlas, since it tends to add some options to its URLs, but a more robust approach would be to add an extra URI options to the `options` hash on the following line.

Looking at [src/DependencyInjection/DoctrineMongoDBExtension.php](https://github.com/doctrine/DoctrineMongoDBBundle/blob/5.0.1/src/DependencyInjection/DoctrineMongoDBExtension.php#L363) in the bundle, the `options` hash gets used as the second argument to PHPLIB's Client constructor.

source/php-frameworks/symfony.txt
```diff
@@ -0,0 +1,397 @@
...
+
+   This step might result in an error message about unset environment
+   variables, but this is resolved in the following section.
+
+Modify Project Files
+~~~~~~~~~~~~~~~~~~~~
+
+This section demonstrates how to modify the files in your
+``restaurants`` project to create a Symfony web application that views
+restaurants based on filter criteria.
```
jmikola, line 234: Is "filter criteria" referring to the query in RestaurantController selecting restaurants in Queens with "Moon" in the name? As a reader, I would have assumed this is actually going to expose some UI to let users dynamically filter the results.

source/php-frameworks/symfony.txt
```diff
@@ -0,0 +1,397 @@
...
+install the ``twig`` bundle:
+
+.. code-block::
+
+   composer require symfony/twig-bundle
+
+.. note::
+
+   This step might result in an error message about unset environment
+   variables, but this is resolved in the following section.
```
jmikola, line 227: If "Set Environment Variables" is just modifying the `.env` file in the project's root directory, why not do that earlier? Perhaps before the "Configure the ODM" step, which is the first time you actually refer to one of those vars.

source/php-frameworks/symfony.txt
```diff
@@ -0,0 +1,397 @@
...
+Next, create the ``RestaurantController.php`` file in the
+``src/Controller`` directory and paste the following code to handle the
+endpoints in your application:
+
+.. literalinclude:: /includes/php-frameworks/RestaurantController.php
+   :caption: src/Controller/RestaurantController.php
+   :language: php
+
+The controller file defines the ``index()`` method, which displays text on
+the web app's front page. The file also defines the ``browse()`` method,
```
jmikola, line 273: @GromNaN: Does it make more sense to refer to method names here instead of actions? I'll defer to you on this paragraph.

source/php-frameworks/symfony.txt
```diff
@@ -0,0 +1,397 @@
...
+which finds documents in which the ``borough``
+field is ``'Queens'`` and the ``name`` field contains the string ``'Moon'``.
+This method then displays the documents at the ``/restaurant/browse/`` route. The
+``browse()`` method uses the ``QueryBuilder`` class to construct the query.
+
+Specify Routes
+``````````````
+
+In the ``config/routes.yaml`` file, add the following routes:
```
jmikola, line 282: @GromNaN: Is this section even necessary since the controller methods have Route attributes? I would have assumed Symfony handles this on its own.

source/php-frameworks/symfony.txt
```diff
@@ -0,0 +1,397 @@
...
+
+Next, create templates to customize the web app's appearance.
+
+Create the ``templates/restaurant`` directory and populate it with the
+following files:
+
+- ``index.html.twig``
+- ``browse.html.twig``
+
+Paste the following code into the ``index.html.twig`` file:
```
jmikola, line 315: Why are the Twig templates hard-coded here instead of referencing files in `includes/php-frameworks`?

source/includes/php-frameworks/Restaurant.php
```diff
@@ -0,0 +1,32 @@
+<?php
```
jmikola, line 1: `includes/php-frameworks` seems like a generic directory that might end up with more content unrelated to this PR. I'd suggest grouping all of the symfony-related files under a "symfony" sub-directory.

source/php-frameworks/symfony.txt
```diff
@@ -0,0 +1,397 @@
...
+
+.. code-block:: html
+   :caption: templates/restaurant/index.html.twig
+
+   {# templates/restaurant/index.html.twig #}
+   
+   {% extends 'base.html.twig' %}
```
jmikola, line 322: @GromNaN: does `base.html.twig` need to be defined or is this inheriting from something in the Symfony app skeleton?

source/php-frameworks/symfony.txt
```diff
@@ -0,0 +1,397 @@
...
+After the server starts, it outputs the following message:
+
+.. code-block:: none
+   :copyable: false
+
+   [OK] Web server listening
+        The Web server is using PHP FPM 8.3.4
+        http://127.0.0.1:8000
+
+Open the URL http://127.0.0.1:8000/restaurant/browse in your web browser.
```
jmikola, line 371: Should this be an actual hyperlink? If not, consider code formatting for the URL.

source/php-frameworks/symfony.txt
jmikola: I would suggest reducing the ODM links to the following:

 * https://www.doctrine-project.org/projects/doctrine-mongodb-odm/en/latest/index.html
 * https://www.doctrine-project.org/projects/doctrine-mongodb-bundle/en/latest/index.html

I don't think it's worth linking the individual GitHub repositories (note you aren't doing so for Symfony), as that just distracts from the documentation.

source/php-libraries.txt
```diff
@@ -68,6 +68,10 @@ Framework Integrations
 
 - Symfony
 
+  - :ref:`php-symfony-integration` describes the advantages of using MongoDB
```
jmikola, line 71: FWIW, I didn't pick up on any "advantages" in the tutorial. It seems more informative than marketing spiel, which I'd consider a good thing.

source/php-frameworks/symfony.txt
GromNaN: This file is added by symfony/flex when Twig is installed on a project: https://github.com/symfony/recipes/blob/main/symfony/twig-bundle/6.4/templates/base.html.twig

source/includes/php-frameworks/Restaurant.php
```diff
@@ -0,0 +1,32 @@
...
+use Doctrine\ODM\MongoDB\Types\Type;
+
+#[
+    MongoDB\Document(collection: 'restaurants')
+]
+
+class Restaurant
+{
+    #[ODM\Id]
+    public  $id;
```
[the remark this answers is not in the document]
alcaeus: It should be a `string`, as ODM's `Id` type converts ObjectIds to strings for better handling in web requests. Off the top of my head I don't know how well Doctrine handles uninitialised properties, so if we want to be cautious it should be a nullable string with default value:

```suggestion
    public ?string $id = null;
```

source/php-frameworks/symfony.txt
```diff
@@ -0,0 +1,397 @@
...
+
+In your applications, you can implement the **MongoDB Doctrine ODM**,
+which is an Object-Document Mapper (ODM) for MongoDB and PHP. It
+provides a way to work with MongoDB in Symfony, using the same
+principles as Doctrine ORM for relational databases.
+
+MongoDB Doctrine ODM allows you to map PHP objects to MongoDB documents
+and query MongoDB by using a builder API. This mapping enables you to
+use other MongoDB features such as flexible schema design and horizontal
+scaling. To learn more about this
```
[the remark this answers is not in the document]
alcaeus: The sharding implementation in ODM was added to facilitate setting up sharded clusters, but more importantly to ensure that the shard key is always added to queries. I don't think we should use "horizontal scaling" as a selling point here. The main selling point for this mapping is that ODM converts BSON documents into PHP objects of specific classes, which then can contain the business logic associated with the document. The rest (e.g. the whole schema management aspect, search indexes, etc.) are just a bonus on top.

source/includes/php-frameworks/Restaurant.php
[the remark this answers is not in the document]
jmikola: > ODM's Id type converts ObjectIds to strings for better handling in web requests

Yes, of course. I want to come out on record here and say that it's been _that_ long since I worked on an app with ODM that I legitimately forgot about that for a hot minute.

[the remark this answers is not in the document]
rustagir: Changed the field attribute to `STRING` as well

source/php-frameworks/symfony.txt
```diff
@@ -0,0 +1,397 @@
...
+
+.. _php-symfony-atlas-cluster:
+
+Create a MongoDB Atlas Cluster
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+You must create a MongoDB cluster where you can store and manage your
+data. Complete the
+:atlas:`Get Started with Atlas </getting-started>` guide to set up a new
+Atlas account, create a free tier MongoDB cluster, load datasets, and
```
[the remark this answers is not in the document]
rustagir: Yes

source/php-frameworks/symfony.txt
```diff
@@ -0,0 +1,397 @@
...
+
+Install MongoDB Extension
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Run the following commands to enter your project directory and install
+MongoDB:
+
+.. code-block:: bash
+
+   cd restaurants
```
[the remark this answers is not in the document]
rustagir: Changed order

source/php-frameworks/symfony.txt
rustagir: Removing that option

source/php-frameworks/symfony.txt
rustagir: I will update the wording. Creating a dynamic filter seemed out of scope for this simple app

source/php-frameworks/symfony.txt
rustagir: I chose to put this step here as it is in the part of the tutorial when users are modifying files in the project. Unless you feel strongly about moving it, I will keep it as the first step in that section

source/php-frameworks/symfony.txt
rustagir: Will move

source/php-frameworks/symfony.txt
rustagir: In the rendered page it is a hyperlink!

source/includes/php-frameworks/symfony/RestaurantController.php
```diff
@@ -0,0 +1,47 @@
...
+use Psr\Log\LoggerInterface;
+use Symfony\Bundle\FrameworkBundle\Controller\AbstractController;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\Routing\Annotation\Route;
+
+class RestaurantController extends AbstractController
+{
+    private DocumentManager $dm;
+    private LoggerInterface $logger;
```
jmikola, line 19: These property names were updated, but various methods below still refer to the underscore-prefixed names.

source/php-frameworks/symfony.txt
```diff
@@ -0,0 +1,372 @@
...
+Why Use MongoDB in a Symfony Application?
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+By using MongoDB as a data store in a Symfony web application, you can
+leverage the document data model to build rich query expressions,
+allowing you to easily interact with data. You can also customize your
+connections to persist and access data to and from multiple databases
+and collections.
+
+In your applications, you can implement the **MongoDB Doctrine ODM**,
```
jmikola, line 49:
```suggestion
In your applications, you can implement the **Doctrine MongoDB ODM**,
```

Consistent with the project's own naming (see: [docs](https://www.doctrine-project.org/projects/doctrine-mongodb-odm/en/latest/)).

I corrected various other references that follow, but you might alternatively prefer to just shorten those references to "Doctrine ODM".

source/php-frameworks/symfony.txt
```diff
@@ -0,0 +1,372 @@
...
+which is an Object-Document Mapper (ODM) for MongoDB and PHP. It
+provides a way to work with MongoDB in Symfony, using the same
+principles as Doctrine ORM for relational databases.
+
+MongoDB Doctrine ODM allows you to map PHP objects to MongoDB documents
```
jmikola, line 54:
```suggestion
Doctrine MongoDB ODM allows you to map PHP objects to MongoDB documents
```

source/php-frameworks/symfony.txt
```diff
@@ -0,0 +1,372 @@
...
+   If you prefer to connect to MongoDB by using the MongoDB PHP Library without
+   Symfony, see `Connecting to MongoDB
+   <https://www.mongodb.com/docs/php-library/current/tutorial/connecting/>`__
+   in the MongoDB PHP Library documentation.
+
+MongoDB Atlas is a fully managed cloud database service that hosts your
+MongoDB deployments. You can create your own free (no credit card
+required) MongoDB Atlas deployment by following the steps in this guide.
+
+This guide uses the MongoDB Doctrine ODM to allow you to map PHP objects to
```
jmikola, line 84:
```suggestion
This guide uses the Doctrine MongoDB ODM to allow you to map PHP objects to
```

source/php-frameworks/symfony.txt
```diff
@@ -0,0 +1,372 @@
...
+
+.. code-block:: bash
+
+   composer create-project symfony/skeleton restaurants
+
+Install PHP Driver and Doctrine ODM
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Enter your project directory, then add the MongoDB PHP driver and the
+MongoDB Doctrine ODM bundle to your application. The bundle integrates
```
jmikola, line 143:
```suggestion
Doctrine MongoDB ODM bundle to your application. The bundle integrates
```

source/php-frameworks/symfony.txt
```diff
@@ -0,0 +1,372 @@
...
+to Your Cluster </tutorial/connect-to-your-cluster/>` tutorial in the
+Atlas documentation.
+
+Save your connection string in a secure location.
+
+Install MongoDB Extension
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+To learn more about installing the MongoDB extension, see `Installing the Extension
+<https://www.mongodb.com/docs/php-library/current/tutorial/install-php-library/>`__
```
jmikola, line 126: This probably makes no difference, but the section link is https://www.mongodb.com/docs/php-library/current/tutorial/install-php-library/#installing-the-extension

source/php-frameworks/symfony.txt
```diff
@@ -0,0 +1,372 @@
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Run the following command to create a skeleton Symfony project called
+``restaurants``:
+
+.. code-block:: bash
+
+   composer create-project symfony/skeleton restaurants
+
+Install PHP Driver and Doctrine ODM
```
jmikola, line 139: I think it's fine to omit "MongoDB" from the ODM title here.

source/php-frameworks/symfony.txt
```diff
@@ -0,0 +1,372 @@
...
+the ODM library into Symfony so that you can read from and write objects
+to MongoDB. To learn more, see the :ref:`php-symfony-resources` section
+of this guide.
+
+Run the following commands to install the driver and the ODM:
+
+.. code-block:: bash
+
+   composer require mongodb/mongodb
+   composer require doctrine/mongodb-odm-bundle
```
jmikola, line 153: @alcaeus: If we're relying on the bundle to pull in the [doctrine/mongodb-odm](https://packagist.org/packages/doctrine/mongodb-odm) package, does it make sense to explicit require `mongodb/mongodb`? That's also a dependency and would be pulled in automatically.

source/php-frameworks/symfony.txt
jmikola: Assuming this was also copied from the dev rel tutorial, I expect there's some redundancy there as well. See: https://symfony.com/doc/current/routing.html

@alcaeus: Should this section be removed entirely? Perhaps "This file sets the following routes in the application" should be relocated to the previous section where the controller class is created, since that ends up defining routes via method attributes.

source/php-frameworks/symfony.txt
```diff
@@ -0,0 +1,372 @@
...
+displayed in the following screenshot:
+
+.. figure:: /includes/php-frameworks/symfony/app_render.png
+   :alt: Restaurants web app screenshot
+
+Congratulations on completing the Quick Start tutorial!
+
+After you complete these steps, you have a Symfony web application that
+connects to your MongoDB deployment, runs a query on
+the sample data, and renders a retrieved result.
```
jmikola, line 353: If you care, the line above this wraps prematurely.

source/php-frameworks/symfony.txt
alcaeus: That's correct, there's no advantage to manually requiring the package. There are some tools that forbid using transitive dependencies (that is, using classes from mongodb/mongodb when it's only required through a different dependency), but strictly speaking it makes no difference. I'd suggest removing the explicit require statement to make the installation process look simpler.

source/php-frameworks/symfony.txt
alcaeus: Yep, there's no need to specify routes manually. In fact, I'd highly suggest we stick to Symfony Best Practices and use `Route` attributes on controller actions to configure the router and do away with the config files.

source/includes/php-frameworks/symfony/Restaurant.php
```diff
@@ -0,0 +1,26 @@
...
+
+use Doctrine\Common\Collections\ArrayCollection;
+use Doctrine\Common\Collections\Collection;
+use Doctrine\ODM\MongoDB\Mapping\Annotations as ODM;
+use Doctrine\ODM\MongoDB\Types\Type;
+
+#[ODM\Document(collection: 'restaurants')]
+class Restaurant
+{
+    #[ODM\Field(type: Type::STRING)]
```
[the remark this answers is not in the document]
alcaeus: I'd add one more thing: `string` is the default type for all fields, so we might as well remove this entirely:
```suggestion
    #[ODM\Field]
```

source/php-frameworks/symfony.txt
rustagir: removed "MongoDB" from most instances

source/php-frameworks/symfony.txt
rustagir: SHould not make a difference for staging but good eye!

source/includes/php-frameworks/Restaurant.php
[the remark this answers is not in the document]
rustagir: Changing the field definition was fine, but changing the annotation to `#[ODM\Field]` for the `id` field resulting in this error: 
`No identifier/primary key specified for Entity...`
Is this expected? Fixed by changing the annotation back to `#[ODM\Id]`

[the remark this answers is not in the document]
jmikola: `#[ODM\Id]` is correct. Even though ODM converts ObjectIds to string in the model class, it should be tracked as an identifier.

source/php-frameworks/symfony.txt
jmikola: @rustagir: This entire section can be removed.

jmikola (approved):
LGTM with YAML routing configuration removed.

PR #975, opened 2024-06-19
DOP-4086: Correct download-and-compile-cpp-driver redirect

i80and:
With our upcoming `mut` release, redirects rejected by AWS S3 now show the line number in the redirects file that is a problem. So, here we resolve the issue.

Pudding proof:

```
$ mut-redirects ../docs-ecosystem/config/redirects | grep docs/drivers/tutorial/download-and-compile-cpp-driver
Redirect 301 /docs/drivers/tutorial/download-and-compile-cpp-driver https://www.mongodb.com/docs/languages/cpp/cpp-driver/current/installation/
```

If this isn't the intended redirect target, we can update as needed.

### DIFF
--- a/config/redirects
+++ b/config/redirects
@@ -52,7 +52,7 @@ raw: docs/drivers/syntax-table -> ${base}/
 
 # cpp
 
-raw: docs/drivers/tutorial/download-and-compile-cpp-driver -> languages/installation/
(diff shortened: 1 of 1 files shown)


norareidy (approved):
LGTM!

